PR #38, opened 2026-03-13
n_hidden=256 at 100 epochs: larger model with proper training budget

tcapelle:
## Hypothesis

n_hidden=256 was previously tried at only 30 epochs and came in at surf_p=195 — much worse than the 128-dim model. However, this was a training budget problem: with 4× the parameters, the model needs more time to converge. At 100 epochs (5× the original budget), the wider model should have enough gradient steps to surpass the 128-dim baseline.

Doubling hidden dim quadruples the attention computation and doubles MLP capacity (128 FFN → 512 FFN intermediate). With a 5-layer model and 4 heads, n_hidden=256 uses n_head=8 to keep head_dim=32.

## Instructions

In `train.py`:

1. **Set epoch budget:**
   ```python
   MAX_TIMEOUT = 10.0  # minutes (wider model is ~2x slower per epoch)
   MAX_EPOCHS = 100
   ```

2. **Replace MSE with Huber(delta=0.01)** in training loop (line 131) and validation loop (line 173):
   ```python
   sq_err = F.huber_loss(pred, y_norm, reduction='none', delta=0.01)
   ```

3. **Change `model_config`** to use wider dimensions:
   ```python
   model_config = dict(
       space_dim=2,
       fun_dim=16,
       out_dim=3,
       n_hidden=256,   # doubled from 128
       n_layers=5,
       n_head=8,       # 8 heads × 32 head_dim = 256
       slice_num=64,
       mlp_ratio=2,
       output_fields=["Ux", "Uy", "p"],
       output_dims=[1, 1, 1],
   )
   ```

4. **Run:**
   ```bash
   python train.py \
     --lr 0.006 \
     --surf_weight 25 \
     --batch_size 4 \
     --wandb_group wider-256-100ep \
     --wandb_name alphonse/h256-100ep \
     --agent alphonse
   ```

Expected VRAM: ~7-8 GB (still well within 96 GB limit).

## Baseline

Current best: `edward/lr6e3-sw25-100ep` (n_hidden=128)

| surf_p | surf_Ux | surf_Uy | vol_p | val_loss |
|--------|---------|---------|-------|----------|
| 33.547 | 0.488 | 0.270 | 63.80 | 0.01896 |

Previous n_hidden=256 at 30ep: surf_p=195 (training budget limited — ignore this).

Target: surf_p < 32.0

---

## Results
_To be filled by student_


agata-mlynarczyk:
STUDENT coretta-scott-king:
SENPAI-RESULT: {"terminal":true,"status":"complete","pending_arms":false,"wandb_run_ids":["4eyiumni","72srkxog","f7okscfp"],"weave_call_ids":[],"primary_metric":{"name":"val_wer","value":0.27178,"direction":"lower"},"training_metric":{"name":"val_wer","value":0.27178}}

## Results — Decoder/joint-favoring discriminative LR (k-sweep)

**Outcome: DECISIVE NEGATIVE.** Every two-speed arm (decoder/joint LR > encoder LR) regresses vs both the single-seed reference (0.25579) and the soup baseline (0.25469). No arm comes near the ~0.255 promotion bar, so no external MEX/QAT eval was run (per the promotion ladder: `val_wer` promotes to external eval, and nothing qualified).

### Sweep (rank by best `val_wer`, lower is better)

| Arm | `decoder_lr_mult` | enc lr | dec+joint lr | **best `val_wer`** | epoch | Δ vs ref 0.25579 | run id | ckpt |
|---|---|---|---|---|---|---|---|---|
| reference (uniform) | 1 | 2e-4 | 2e-4 | **0.25579** | — | — | — | — |
| soup baseline | — | — | — | **0.25469** | — | — | — | — |
| **k=2** | 2 | 2e-4 | 4e-4 | **0.27178** | 8 | **+0.0160** | `4eyiumni` | `disc-lr-k2-checkpoint:v8` (best) |
| **k=8** | 8 | 2e-4 | 16e-4 | **0.27619** | 9 | **+0.0204** | `f7okscfp` | `disc-lr-k8-checkpoint:v9` (best) |
| **k=4** | 4 | 2e-4 | 8e-4 | **0.28004** | 9 | **+0.0243** | `72srkxog` | `disc-lr-k4-checkpoint:v9` (best) |

- Best two-speed arm = **k=2, `val_wer` 0.27178**, still **+0.016** over the reference — at/just past the noise floor (1σ=0.00758; gaps below ~0.015 are ties), so a real regression.
- Spread among k arms = 0.27178–0.28004 (**range 0.0083, within 1σ**) → no productive "sweet spot"; the k-ranking itself is a tie. The signal is the *direction*: favoring the decoder/joint with a higher LR than the encoder hurts, for every k tested.

### Implementation verified
- Two param groups by module: group A = `encoder` (+ non-head) at `lr`; group B = `decoder` (prediction net) + `joint` at `decoder_lr_mult × lr`. Each group keeps its own `lr`, so NeMo `CosineAnnealing` anneals each from its own `initial_lr` (confirmed in source: `_get_lr`/`_get_warmup_lr` iterate `self.base_lrs`).
- Per-group `lr-AdamW/encoder` and `lr-AdamW/decoder_joint` logged via `LearningRateMonitor`. **Peak LR ratios verified exactly: k=4 → 4.000, k=8 → 8.000** (encoder peak 2.0e-4 in both).
- Debug run (`t1vckwts`, `--max_steps 20`) confirmed two distinct curves before the full sweep.

### Config diff (vs base recipe)
- `asr/config.py`: `OptimizerConfig` gains `decoder_lr_mult: float = 1.0` (1.0 = uniform, no behavior change to existing recipes).
- `train.py`: `build_disc_lr_param_groups()` + override of `setup_optimizer_param_groups` at the `setup_optimization` call site (only when `decoder_lr_mult != 1.0`); added `LearningRateMonitor(logging_interval="step")`.

### Exact commands
```bash
# k=4 (72srkxog)
python train.py --config_path configs/train_config.yaml \
  --lr 2e-4 --decoder_lr_mult 4 --max_epochs 10 --seed 42 \
  --name "coretta-scott-king/disc-lr-k4" --tags finetune milieu-asr-v2-weekend disc-lr
# k=8 (f7okscfp): --decoder_lr_mult 8 ; k=2 (4eyiumni, prior run): --decoder_lr_mult 2
```
> NOTE (PR doc mismatch, not a code bug): the PR's `--model.optim.lr` / `--trainer.max_steps` / `--wandb.name` style is **not** what the simple_parsing CLI accepts. It flattens to leaf names: `--lr`, `--decoder_lr_mult`, `--max_epochs`, `--max_steps`, `--sched.max_steps`, `--seed`, `--name`, `--tags`. Verified via `python train.py --help`; I used the flat flags.

### Run facts
- Base recipe held fixed: stage2, encoder lr 2e-4, seed 42, 10 epochs, batch 8, AdamW wd 0.001, CosineAnnealing warmup 30, deterministic. Only the optimizer param-grouping changed.
- Runtime: k4 24.1m, k8 24.2m (k2 24.6m) — all under the 30-min ceiling.
- Peak GPU memory: ~91 GiB (≈95% of the 97 GiB RTX PRO 6000); batch 8 fit without OOM (mostly caching-allocator reservation).

### What happened — analysis
- The proven recipe (uniform lr 2e-4) sits at a well-tuned point. The prediction net + joint are small relative to the encoder; giving them a higher LR **over-adapts the token-emission mapping** on a tiny train set (2,990 samples) rather than sharpening domain-vocab discrimination.
- **Overfitting signature**: k=8 `val_loss` bottoms at epoch 6 (~0.567) then *rises* through epochs 7–9 even as `val_wer` keeps drifting — the joint/decoder is memorizing, not generalizing.
- **Slower, not better, convergence**: best epoch shifts later with higher k (k2 @8, k4/k8 @9, still descending) — the aggressive head LR destabilizes early training, and within the fixed 10-epoch budget the arms never claw back to the reference level.
- This **complements the prior encoder result** (freezing / down-weighting the encoder = decisive negative): perturbing the encoder:head LR ratio in *either* direction (encoder-down OR head-up) regresses. The uniform-LR operating point is good and the **optimizer-LR lever looks spent**.

### Caveats
- Single seed (42) per arm; the inter-k differences are within 1σ (treat as tied). The regression vs reference (+0.016 to +0.024) is the robust signal.
- Higher-k arms were still improving at epoch 9 — a longer budget *might* narrow the gap, but the reference already reaches 0.25579 in 10 epochs, so a method that needs more epochs just to match is not a win.

### Suggested follow-ups (not implemented here)
1. **Retire the encoder/head LR-ratio lever** — both directions are now decisive negatives.
2. If decoder/joint domain-vocab discrimination is still the target, prefer mechanisms that don't destabilize global training: a short decoder/joint-only warm-up phase *then* uniform-LR joint training; or keep uniform LR and push the signal through **data** (F1-jargon weighting / token upsampling — overlaps with the assigned jargon-oversample direction).
3. With the decode-side language-prior tier and weight-averaging both spent, the remaining lever is most likely **data**, not optimizer surgery.

### Recommendation: **ABANDON** this hypothesis. Clean negative; do not merge. The implementation (`decoder_lr_mult`, default 1.0 = no-op) is low-risk and can be kept for the record or reverted — it changes nothing unless `decoder_lr_mult != 1.0`.

PR #57, opened 2026-03-13
Two Transolver layers (n_layers=2) with T_max=50

tcapelle:
## Hypothesis

The 1-layer model gets one round of physics attention then immediately produces output. A second layer lets the model refine its slice-based representations: layer 1 finds coarse flow patterns, layer 2 corrects them. Even going from 1 to 2 layers is a major capacity jump for transformers. The per-epoch cost increase should be moderate (~2-3s extra).

Risk: if per-epoch time doubles to ~16s, we only get ~20 epochs, which may negate the benefit.

## Instructions

In `train.py`:

1. Set `MAX_EPOCHS = 50`
3. Change model config: `n_hidden=128`, `n_layers=2`, `n_head=4`, `slice_num=64`, `mlp_ratio=2`
4. Scheduler: `CosineAnnealingLR(optimizer, T_max=MAX_EPOCHS)` — T_max=50
5. Keep MSE loss
6. Use `--wandb_name alphonse/2layers-tmax50` and `--wandb_group architecture`

## Baseline

- vol_p=63.80, best_epoch=97

---
## Results

**W&B run:** `i5neqm9f` (alphonse/2layers-tmax50, group: architecture)

**Training:** 23 epochs completed before 5-minute wall-clock timeout (hit at 5.2 min). Best at epoch 23 (final epoch — still improving at cutoff). ~13s/epoch.

**Peak memory:** 7.1 GB (vs ~4.3 GB for n_layers=1)

| Metric | n_layers=2 (this run) | n_layers=1 PR#49 | Baseline (n_layers=1, Huber) |
|--------|----------------------|-----------------|------------------------------|
| val/loss | 3.1020 | 2.9403 | 0.0190 (different scale) |
| surf_p | 119.4 | 123.4 | 33.55 |
| surf_Ux | 1.47 | 1.52 | 0.4875 |
| surf_Uy | 0.74 | 0.96 | 0.2704 |
| vol_p | 173.3 | 179.2 | 63.80 |
| best_epoch | 23/23 | 33/40 | 97 |
| epoch_time | ~13s | ~8s | ~3s |

**What happened:**

n_layers=2 shows marginal improvement over n_layers=1 (PR#49 run) across all surface metrics — surf_Uy in particular improved from 0.96 to 0.74 (−23%). However, neither run is close to the baseline, largely due to:

1. **Missing Huber loss:** Both experiments use MSE loss; the baseline uses Huber (delta=0.01). The val_loss scale difference (3.1 vs 0.019) is a loss function artifact, but the surface MAE gap (~4x) represents a real deficit.

2. **Timeout effect confirmed as predicted:** The PR hypothesis was correct — epoch time ~doubled (8s → 13s), cutting training from ~40 epochs to ~23. The best epoch was epoch 23 = the last epoch, meaning the model was still actively learning. More capacity helps only if you can actually train it.

3. **Memory**: 7.1 GB vs 4.3 GB for 1 layer — the +65% VRAM increase is acceptable given GPU capacity (96 GB), but not a concern here.

**Directional signal:** 2 layers is meaningfully better than 1 layer per epoch of training (better results in fewer epochs), suggesting higher capacity does help. The hypothesis is sound — but the 5-minute constraint makes the per-epoch cost increase a real problem. Deeper models need either faster training or more time budget to demonstrate their advantage.

**Suggested follow-ups:**
- Establish a fair Huber-loss baseline at n_layers=1 to isolate the architecture effect
- Try n_layers=2 with gradient accumulation (e.g., effective batch=8 but 2 forward passes) to partially compensate for shorter runs
- Check if n_layers=2 is converging faster in early epochs than n_layers=1 (already some evidence: better after 23 epochs than n_layers=1 after 40)


tcapelle:
## Advisor Review — Closing

**Result:** surf_p=119.4 at epoch 23 (~13s/epoch). n_layers=2 is too slow under contention. The student notes 2 layers shows better per-epoch convergence than 1 layer — interesting but untestable in 5 min.

**Takeaway:** 2-layer model needs longer training budget. Shelving until contention resolves.

PR #179, opened 2026-03-14
Validate every 3 epochs for more training time

tcapelle:
## Hypothesis
Validation runs every epoch and takes ~1.2s of the 6.3s per epoch. By validating only every 3 epochs, we save ~0.8s on 2 out of 3 epochs, gaining ~3-4 extra training epochs within the 5-minute budget (~51-52 instead of 48). The model is still converging at ep48, so every extra training epoch matters.

## Instructions
All changes in `train.py`:

1. Add a constant near the top (after MAX_EPOCHS):
   ```python
   VAL_EVERY = 3
   ```

2. Wrap the entire validation section in a conditional. Before `model.eval()` (around line 157), add:
   ```python
   if (epoch + 1) % VAL_EVERY == 0 or epoch == 0:
   ```
   And indent the entire validation block (from `model.eval()` through the best-model saving logic) under this condition.

3. Make sure the `scheduler.step()` call remains OUTSIDE the validation conditional (it should run every epoch).

4. Make sure the `best_val_loss`, `best_mae_surf`, and logging remain correct. If validation is skipped, don't update `best_val_loss` or `best_mae_surf` for that epoch. The training loop logging can continue every epoch.

5. Keep all other settings: lr=0.015, sw=12, wd=0, grad clip 1.0, 1L h128.

6. Use `--wandb_name "edward/val-every-3"` and `--wandb_group "mar14d"` and `--agent edward`

## Baseline
| Metric | Current best (PR #169) |
| surf_p | 45.47 |
| surf_ux | 0.58 |
| surf_uy | 0.34 |
| Epochs | ~48 in 5 min |

---

## Results

**W&B run ID**: cl2u6e2k
**Best epoch**: 48 (last validated epoch; total epochs completed: 49)
**Peak memory**: 3.7 GB

| Metric | Baseline (PR #169) | This run | Delta |
|--------|-------------------|----------|-------|
| surf_p | 45.47 | 52.2 | +6.73 |
| surf_ux | 0.58 | 0.62 | +0.04 |
| surf_uy | 0.34 | 0.37 | +0.03 |
| val_loss | — | 1.466 (sw=12) | — |
| vol MAE Ux | — | 3.60 | — |
| vol MAE Uy | — | 1.39 | — |
| vol MAE p | — | 92.7 | — |

### What happened

The hypothesis did not work. Validating every 3 epochs only gained 1 extra training epoch (49 vs 48) — the time savings were smaller than expected. Epoch training time is ~5s, and validation is ~1s, so saving validation on 2/3 epochs only frees up ~0.67s per epoch on average, insufficient for meaningfully more epochs in 5 minutes.

All metrics are worse than baseline: surf_p went from 45.47 to 52.2, surf_ux from 0.58 to 0.62, surf_uy from 0.34 to 0.37. This may partly be due to the sparse checkpoint selection — the best model is only checked at epochs 1, 3, 6, 9, ..., 48, missing potentially better checkpoints in between. The true best epoch might have been 49 or 50 (not validated).

The core problem is that training epochs are already much longer than validation, so skipping validation provides minimal time benefit while making checkpoint selection coarser.

### Suggested follow-ups

- This approach isn't worth pursuing further — the time savings are too small to matter with this fast training loop.
- A more promising direction: use the saved val time to run a slightly larger model (e.g., increase mlp_ratio=3 which would add capacity without as much overhead as doubling hidden_dim).

### DIFF
--- a/train.py
+++ b/train.py
@@ -22,6 +22,7 @@ from utils import visualize, dataset_stats
 
 MAX_TIMEOUT = 5.0 # minutes
 MAX_EPOCHS = 50
+VAL_EVERY = 3
 @dataclass
 class Config:
     lr: float = 0.015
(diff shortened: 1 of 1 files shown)


tcapelle:
Closing: validate every 3 epochs gave surf_p=52.16, worse than baseline (45.47). The reduced validation didn't translate to enough extra training epochs to compensate. Also has merge conflicts after PR #175 merge.

PR #193, opened 2026-03-14
Snapshot ensemble: 4-cycle cosine + weight averaging

tcapelle:
## Hypothesis
Instead of one long cosine schedule, use 4 short cycles (CosineAnnealingWarmRestarts T_0=17, T_mult=1). Save model weights at the end of each cycle (ep 17, 34, 51, 68). After training, average all 4 weight snapshots. This is the Snapshot Ensemble technique — each cycle endpoint is a different local minimum, and their average finds a flatter, more generalizable solution. Zero extra training cost; just a post-processing step.

## Instructions
All changes in `train.py`: see PR diff.

## Baseline
| surf_p | 42.10 | surf_ux | 0.56 | surf_uy | 0.30 |

---

## Results

**W&B run ID:** `xbzrfy7d`
**Epochs completed:** 68 (4 snapshots at ep 17, 34, 51, 68)
**Peak memory:** 2.6 GB

| Metric | Baseline | Best individual (ep68) | Ensemble (4 snapshots) |
|--------|----------|------------------------|------------------------|
| surf_p | 42.10 | 45.9 | 80.2 |
| surf_Ux | 0.56 | 0.57 | 2.78 |
| surf_Uy | 0.30 | 0.33 | 0.49 |

**What happened:**

The snapshot ensemble dramatically hurt performance: surf_p went from 45.9 (best individual) to 80.2 (ensemble). Weight averaging across 4 snapshots from different convergence stages is actively harmful.

The fundamental problem: snapshots are from very different loss regions. Snapshot at ep17 has surf_p ~95, ep34 ~62, ep51 ~51, ep68 ~46. Averaging poorly-converged early weights with well-converged late weights creates a model that is not good at anything.

The Snapshot Ensemble technique assumes all snapshots are near different good local minima. But 17 epochs is not enough to converge, so early cycle endpoints are just bad checkpoints.

Also: the best individual model (surf_p=45.9) is itself worse than baseline (42.10), showing that 4 short cycles with T_0=17 is worse than a single longer cosine decay.

**Suggested follow-ups:**
- Snapshot ensemble requires all snapshots near convergence — needs much longer T_0 (50+ epochs) for each cycle to actually converge
- Better averaging: SWA (Stochastic Weight Averaging) — average only the last N checkpoints of a single run where the model IS near convergence
- The individual best (ep68, surf_p=45.9) shows warm restarts with T_0=17 are suboptimal vs. a single cosine schedule to convergence

### DIFF
--- a/train.py
+++ b/train.py
@@ -4,6 +4,7 @@
 
 """Train Transolver on full-field airfoil flow prediction with separate surface/volume losses."""
 
+import copy
 import os
 import time
 import torch
@@ -80,7 +81,7 @@ model = Transolver(
 
 n_params = sum(p.numel() for p in model.parameters())
 optimizer = torch.optim.AdamW(model.parameters(), lr=cfg.lr, weight_decay=cfg.weight_decay)
-scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=MAX_EPOCHS)
+scheduler = torch.optim.lr_scheduler.CosineAnnealingWarmRestarts(optimizer, T_0=17, T_mult=1)
 
 
 # --- wandb ---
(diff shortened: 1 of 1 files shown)


tcapelle:
Closing: did not beat baseline. lr=0.010 (PR #196, surf_p=41.38) was the winning experiment this round.

PR #288, opened 2026-03-14
Pure L1 surface loss from epoch 0 (no MSE phase)

tcapelle:
## Hypothesis
The MSE→L1 curriculum uses MSE for the first 4 epochs as a "warm start." But with lr=2e-3 providing strong initial signal already, the MSE phase may not be needed. Pure L1 from epoch 0 gives the model 10 full L1 epochs instead of 6. L1's constant-magnitude gradients prevent over-focusing on outliers, which may be particularly beneficial in the early chaotic phase of training.

The curriculum was our biggest win — but maybe it was the L1 that helped, not the gradual transition.

## Instructions

In `train.py`, change the curriculum function (line 104):

```python
# OLD:
    alpha = min(epoch / 4.0, 1.0)  # MSE→L1 transition

# NEW:
    alpha = 1.0  # Pure L1 from start
```

This makes the function always use L1: `elem_loss = diff.abs()`.

**No other changes.** Run:
```bash
python train.py --agent frieren --wandb_group mar14b-pure-l1 --wandb_name "frieren/pure-l1"
```

## Baseline
| Metric | Value |
|--------|-------|
| **surf_p** | **107.35** |
| surf_Ux | 1.23 |
| surf_Uy | 0.84 |

---

## Results

**W&B run:** `b8nb9ap4`  
**Epochs completed:** 5 (hit 5-min wall-clock limit; epochs 1–4 took ~67s each vs typical ~32s)  
**Peak memory:** 15.4 GB

| **surf_p** | **107.35** | **162.6** | **+51.4% ❌** |
| surf_Ux | 1.23 | 2.00 | +62.6% ❌ |
| surf_Uy | 0.84 | 0.92 | +9.5% ❌ |
| vol_loss | — | 0.4237 | — |
| surf_loss | — | 0.3950 | — |

**What happened:** Strong negative result on all metrics. Pure L1 performed dramatically worse than the MSE→L1 curriculum. Two compounding factors:

1. **Optimization instability early on:** Pure L1 gradients are scale-invariant, which can cause chaotic behavior in the early training phase when predictions are far from targets. MSE provides magnitude-weighted gradients that help steer the model quickly toward the right ballpark before L1 takes over for fine-grained accuracy.

2. **Fewer effective epochs:** Epochs 1–4 each took ~67s (vs. typical ~32s), likely due to early-training gradient instability causing many optimizer steps to struggle. Only 5 epochs completed within the 5-minute budget, while the baseline gets 9–10 epochs.

The curriculum was doing real work. The gradual MSE→L1 transition is essential — the MSE warm-start phase is what allows stable training in the first few epochs, and the transition to L1 after epoch 4 is what drives the final accuracy improvement.

**Suggested follow-ups:**
- Try a faster transition (e.g., epoch 2 instead of 4) to see if the MSE phase can be shortened while preserving stability
- Try mixing the losses differently (e.g., `0.5 * MSE + 0.5 * L1` as a fixed blend rather than a schedule)
- Profile why early epochs are so much slower with pure L1 — may reveal something about gradient computation cost

### DIFF
--- a/.experiment
+++ b/.experiment
@@ -1,1 +1,1 @@
-# alpha=1 by epoch 4
+# pure L1 from start


tcapelle:
Closing — pure L1 gives surf_p=140.5, much worse. The MSE phase (epochs 0-4) is critical for initial learning. The curriculum transition is essential, not just L1.

PR #294, opened 2026-03-14
Dedicated pressure output head (separate from Ux/Uy)

tcapelle:
## Hypothesis
The current model predicts [Ux, Uy, p] from a shared output layer. But pressure and velocity have very different error scales (surf_p=107 vs surf_Ux=1.23) and physical characteristics. Splitting pressure into its own dedicated head gives it a fully separate linear projection from the backbone features, allowing the pressure pathway to specialize without interference from velocity. This is conceptually different from the existing surface/volume dual heads — this splits by *channel* (pressure vs velocity), not by *location*.

### In `transolver.py`, modify the last `TransolverBlock`:

**In `__init__` (where `last_layer=True`):**
```python
        if self.last_layer:
            self.mlp2 = nn.Linear(hidden_dim, out_dim)       # volume head (Ux, Uy, p)
            self.surf_mlp = nn.Linear(hidden_dim, out_dim)    # surface head (Ux, Uy, p)  [EXISTING]
            self.pressure_head = nn.Linear(hidden_dim, 1)     # NEW: dedicated pressure head
```

**In `forward` (last_layer branch):**
```python
        if self.last_layer:
            h = self.ln_3(fx)
            if is_surface is not None:
                vol_pred = self.mlp2(h)           # (B, N, 3)
                surf_pred = self.surf_mlp(h)      # (B, N, 3)
                # Override pressure channel with dedicated head
                p_pred = self.pressure_head(h)    # (B, N, 1)
                surf_pred = torch.cat([surf_pred[:, :, :2], p_pred], dim=-1)
                mask = is_surface.unsqueeze(-1).expand_as(vol_pred)
                return torch.where(mask, surf_pred, vol_pred)
            return self.mlp2(h)
```

This means surface nodes get pressure from a dedicated pathway while velocity comes from the surface head. Volume nodes still use the single volume head for all 3 channels.

Run:
```bash
python train.py --agent frieren --wandb_group mar14b-pressure-head --wandb_name "frieren/pressure-head"
```

## Baseline
| Metric | Value |
|--------|-------|
| **surf_p** | **107.35** |
| surf_Ux | 1.23 |
| surf_Uy | 0.84 |

---

## Results

**W&B run:** `bu48ylfm`  
**Epochs:** 10 (best at epoch 10)  
**Peak memory:** 15.6 GB (+0.2 GB vs baseline)

| **surf_p** | **107.35** | **110.1** | **+2.6% ❌** |
| surf_Ux | 1.23 | 1.15 | -6.5% ✓ |
| surf_Uy | 0.84 | 0.86 | +2.4% ~ |
| vol_loss | — | 0.3394 | — |
| surf_loss | — | 0.2148 | — |

**What happened:** The dedicated pressure head did not improve surf_p — it performed slightly worse than the baseline (+2.6%). Ux improved modestly, which may indicate the velocity head benefited from not having to simultaneously model pressure, but the pressure head itself didn't gain from its specialization. The likely reason: a single linear layer (`hidden_dim → 1`) has the same expressiveness as a slice of the existing 3-output head. Splitting it out doesn't give the pressure pathway more capacity — it just removes the joint optimization effect where pressure and velocity features reinforce each other. With only 256→1 parameters, the pressure head can't learn richer features than the existing head already provides.

**Suggested follow-ups:**
- Try a multi-layer MLP pressure head (256→128→1) to actually add capacity, not just split
- Try applying the dedicated head to both surface and volume pressure outputs
- Investigate whether the shared representation in the last layer's `h` features is already specialized enough for pressure (i.e., the bottleneck may be earlier in the network, not in the output head)

### DIFF
--- a/transolver.py
+++ b/transolver.py
@@ -142,6 +142,7 @@ class TransolverBlock(nn.Module):
             self.ln_3 = nn.LayerNorm(hidden_dim)
             self.mlp2 = nn.Linear(hidden_dim, out_dim)       # volume head
             self.surf_mlp = nn.Linear(hidden_dim, out_dim)    # surface head
+            self.pressure_head = nn.Linear(hidden_dim, 1)     # dedicated pressure head
 
     def forward(self, fx, is_surface=None):
         fx = self.attn(self.ln_1(fx)) + fx
@@ -149,8 +150,11 @@ class TransolverBlock(nn.Module):
         if self.last_layer:
             h = self.ln_3(fx)
             if is_surface is not None:
-                vol_pred = self.mlp2(h)
-                surf_pred = self.surf_mlp(h)
+                vol_pred = self.mlp2(h)           # (B, N, 3)
+                surf_pred = self.surf_mlp(h)      # (B, N, 3)
+                # Override pressure channel with dedicated head
+                p_pred = self.pressure_head(h)    # (B, N, 1)
+                surf_pred = torch.cat([surf_pred[:, :, :2], p_pred], dim=-1)
                 mask = is_surface.unsqueeze(-1).expand_as(vol_pred)
                 return torch.where(mask, surf_pred, vol_pred)
             return self.mlp2(h)


tcapelle:
Closing — pressure head gave surf_p=110.1, slightly worse than baseline 107.35. The dedicated pressure pathway didn't add enough value over the existing dual surface/volume heads.

PR #299, opened 2026-03-14
Cosine warm restarts (T_0=5, two LR cycles in 10 epochs)

tcapelle:
## Hypothesis
The current CosineAnnealingLR with T_max=10 does a single cycle — the LR drops very low in the final epochs. With only ~10 epochs of training, this means the model has almost no learning rate budget left after the MSE→L1 curriculum completes at epoch 4. The optimizer is essentially frozen during the pure L1 refinement phase.

**CosineAnnealingWarmRestarts with T_0=5** gives two full cosine cycles. The key insight: the LR restart at epoch 5 coincides with the post-curriculum-transition phase (alpha=1.0 since epoch 4), giving the optimizer fresh high LR to refine under pure L1 loss. This is distinct from OneCycleLR (tried, failed) — warm restarts provide periodic resets rather than a single peak.

## Instructions

In `train.py`, change one line — the scheduler definition (line 83):

**Before:**
```python
scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=10)
```

**After:**
```python
scheduler = torch.optim.lr_scheduler.CosineAnnealingWarmRestarts(optimizer, T_0=5, T_mult=1, eta_min=1e-5)
```

That's the only change. `T_0=5` means the first restart happens at epoch 5 (two cycles of 5 epochs each in 10 total). `eta_min=1e-5` prevents LR from hitting exactly zero. Everything else stays the same.

W&B tag: `mar14b`, group: `warm-restarts`

## Baseline

---

## Results
_To be filled by student_


tcapelle:
## Results

**W&B run:** fgr6zda9

| surf_p | 107.35 | **115.1** |
| surf_Ux | 1.23 | 1.17 |
| surf_Uy | 0.84 | 0.79 |
| val/loss | 2.45 | 2.75 |

**Outcome: Negative** — surf_p degraded by ~7.75 points vs baseline.

The LR restart at epoch 5 coincides with the pure L1 phase, but rather than helping, it appears the fresh high LR disrupts fine-grained surface refinement. Standard CosineAnnealingLR with T_max=10 gives a much lower final LR which helps the last epochs converge better.

PR #301, opened 2026-03-14
Double slice_num (64 → 128 physics attention slices)

tcapelle:
## Hypothesis
The physics attention mechanism partitions mesh nodes into `slice_num=64` soft groups (slices) based on feature similarity. Each slice represents a local physical regime. With ~7000-12000 nodes per sample, 64 slices means each slice aggregates ~100-190 nodes on average — this may be too coarse to distinguish fine-grained pressure variations along the airfoil surface.

Doubling to `slice_num=128` gives finer-grained physics decomposition: ~55-95 nodes per slice. This allows the attention mechanism to form more specialized groups, potentially separating leading-edge high-pressure regions from trailing-edge wake regions more precisely. The additional compute is modest (the attention is over slices, not nodes, so going 64→128 slices increases attention cost but it's still small relative to the node projections).

## Instructions

In `train.py`, change one value in the `model_config` dict (around line 70):

**Before:**
```python
slice_num=64,
```

**After:**
```python
slice_num=128,
```

That's the only change. Everything else stays the same.

W&B tag: `mar14b`, group: `more-slices`

## Baseline

---

## Results
_To be filled by student_


tcapelle:
## Results

**W&B run:** 67xavjrb

| surf_p | 107.35 | **204.9** |
| surf_Ux | 1.23 | 2.58 |
| surf_Uy | 0.84 | 1.19 |
| val/loss | 2.45 | 3.26 |

**Outcome: Strongly Negative** — surf_p nearly doubled vs baseline.

Side effects: VRAM increased from 15.4GB → 20.2GB, epoch time increased from 31s → 41s (only 8 epochs fit in 5-minute budget vs 10). Validation loss was unstable (epoch 6 spiked to 1.11, epoch 8 to 0.78). Best checkpoint was at epoch 4 only.

Doubling slice_num from 64→128 significantly increases the attention state space. With the same learning rate and only ~8 effective epochs, the model cannot learn to use the additional slices effectively. The instability suggests the higher-dimensional attention is harder to optimize in the short training budget.

tcapelle:
**Closing — significantly worse than baseline.**

| surf_p | 107.35 | 204.9 | +91% ❌ |
| surf_Ux | 1.23 | 2.58 | +110% ❌ |
| surf_Uy | 0.84 | 1.19 | +42% ❌ |

Catastrophically worse. 128 slices massively increases compute (41s/epoch vs 31s, only 8 epochs), adds instability, and the model never converges properly. The current 64 slices is well-tuned for this problem scale.

PR #303, opened 2026-03-14
Surface head init: larger std + pressure bias offset

tcapelle:
## Hypothesis
The surface output head is initialized with `trunc_normal_(std=0.02)` like all other layers. This means initial predictions are near-zero. But surface pressure targets in normalized space have a slight negative offset (surface pressure mean ≈ -248 vs global mean ≈ -195, creating a normalized offset of about -0.06). The model spends early epochs just learning this offset.

By initializing the surface head with:
1. Slightly larger weight std (0.05 vs 0.02) for faster initial exploration
2. A small negative bias on the pressure channel (-0.06) to match the surface pressure offset

...we give the optimizer a head start. The MSE phase becomes more productive when starting closer to the correct output distribution. Zero runtime cost — purely an initialization change.

## Instructions

In `transolver.py`, modify the `initialize_weights` method (line 223-224). After the `self.apply(self._init_weights)` call, add a post-init hook for the surface head:

**Before:**
```python
def initialize_weights(self):
    self.apply(self._init_weights)
```

**After:**
```python
def initialize_weights(self):
    self.apply(self._init_weights)
    # Give surface head a head start
    last_block = self.blocks[-1]
    if hasattr(last_block, 'surf_mlp'):
        nn.init.trunc_normal_(last_block.surf_mlp.weight, std=0.05)
        nn.init.constant_(last_block.surf_mlp.bias, 0)
        last_block.surf_mlp.bias.data[2] = -0.06  # pressure channel offset
```

That's the only change. Everything else stays the same.

W&B tag: `mar14b`, group: `output-scale-init`

## Baseline

---

## Results
_To be filled by student_


tcapelle:
## Results

**W&B run:** czzw7vys

| surf_p | 107.35 | **165.6** |
| surf_Ux | 1.23 | 2.60 |
| surf_Uy | 0.84 | 1.37 |
| val/loss | 2.45 | 2.61 |

**Outcome: Negative** — surf_p degraded by ~58 points vs baseline.

Training was highly unstable: surf_p oscillated between epochs (330→266→293→226→283→193→233→166→301→240), best checkpoint at epoch 8 only. The larger weight std (0.05 vs 0.02) appears to cause destabilizing early gradients. The pressure bias offset (-0.06) did not help overcome this instability. The standard trunc_normal std=0.02 initialization is better suited for this architecture.

Note: An earlier run accidentally used slice_num=128 (from a dirty working tree) giving surf_p=212.1. This run correctly used slice_num=64.

PR #311, opened 2026-03-14
Reduce slice_num (32 → 16) for faster epochs

tcapelle:
## Hypothesis
We are epoch-limited: every experiment that slowed epochs got worse (2 layers, 4 heads, 128 slices). Conversely, every speedup that increased epoch count has helped. Reducing slice_num from 32 to 16 should speed up physics attention (fewer slices to compute, aggregate, and attend over). If we gain even 0.5s/epoch (3.5s vs 4s), that's ~5 more epochs (75 vs 70), which could push convergence further.

The risk: 16 slices may be too coarse for the physics decomposition. But with 2 heads and a 1-layer model, the attention capacity is already limited — fewer slices may actually help by forcing more coherent groupings.

## Instructions

In `train.py`, change one value in model_config (line 70):

**Before:**
```python
slice_num=32,
```

**After:**
```python
slice_num=16,
```

That's the only change.

W&B tag: `mar14b`, group: `slice16`

## Baseline
- **surf_p = 39.4**, surf_Ux = 0.58, surf_Uy = 0.31
- 1-layer, 2 heads, 32 slices, lr=0.008, ~70 epochs at 4s/epoch

---

## Results

**W&B run:** `69dt9p8e`  
**Epochs completed:** 70 (all MAX_EPOCHS; 4s/epoch, no timeout)  
**Best checkpoint:** epoch 68  
**Peak memory:** 2.5 GB (vs 2.6 GB for slice_num=32)

| Metric | Baseline (32 slices) | This run (16 slices) | Δ |
|--------|---------|----------|---|
| **surf_p** | **39.4** | **40.5** | **+2.8% ~** |
| surf_Ux | 0.58 | 0.57 | -1.7% ~ |
| surf_Uy | 0.31 | 0.31 | 0% ~ |
| vol_loss | — | 0.1351 | — |
| surf_loss | — | 0.0616 | — |

**What happened:** Essentially neutral result. Both slice_num=16 and slice_num=32 run at 4s/epoch with 2 heads in this configuration, so there was no epoch-count gain (both ran 70 epochs). The speedup hypothesis didn't materialize because the attention computation at this scale is not the bottleneck — the epoch time is dominated by data loading, preprocessing, and the MLP layers.

surf_p at 40.5 vs 39.4 baseline is within noise; the models converge to very similar performance. Halving the slices from 32 to 16 doesn't significantly change the physics decomposition quality for this 1-layer, 2-head architecture.

**Suggested follow-ups:**
- Try slice_num=8 to see if there's a degradation cliff (at some point, too few slices will hurt)
- The epoch count is the real lever — try MAX_EPOCHS=100 or MAX_TIMEOUT=8 to see if the model is still improving beyond epoch 70
- Since both 16 and 32 slices give similar results at 4s/epoch, stay at 32 slices (more capacity) as the default


tcapelle:
**Closing — no improvement over baseline.**

| Metric | Baseline (32 slices) | 16 slices | Delta |
|--------|---------------------|-----------|-------|
| surf_p | **39.4** | 40.5 | +2.8% ❌ |

No speedup materialized — both run at 4s/epoch. The bottleneck is data loading/preprocessing, not attention compute. 32 slices remains optimal.

PR #391, opened 2026-03-16
Cosine warm restarts (T_0=30, T_mult=2) + checkpoint averaging

tcapelle:
## Hypothesis
The current single cosine cycle decays LR to near-zero by epoch 91, where the model is still improving. Warm restarts periodically reset LR, allowing escape from narrow valleys.

With T_0=30, cycle 1 covers epochs 0-30, cycle 2 covers 30-90. At each cycle minimum (epochs 30 and 90), save a checkpoint. Average the weights of both converged checkpoints for a flatter minimum.

This differs from SWA (which averaged many poorly-converged checkpoints) and snapshot ensembles (which used too-short cycles). T_0=30 gives enough epochs for real convergence.

## Instructions

Make these changes to `structured_split/structured_train.py`:

1. **Replace the scheduler**:
   ```python
   # Remove: warmup + cosine sequential scheduler
   # Add: Manual warmup (3 epochs) then warm restarts
   scheduler = torch.optim.lr_scheduler.CosineAnnealingWarmRestarts(
       optimizer, T_0=30, T_mult=2, eta_min=1e-5
   )
   # Manual warmup for first 3 epochs
   warmup_epochs = 3
   ```

2. **Implement manual warmup** in the training loop:
   ```python
   if epoch < warmup_epochs:
       for pg in optimizer.param_groups:
           pg['lr'] = cfg.lr * (epoch + 1) / warmup_epochs
   else:
       scheduler.step(epoch - warmup_epochs)
   ```

3. **Save checkpoints at cycle boundaries** (epoch 30 and best overall):
   ```python
   if epoch == warmup_epochs + 30:  # End of cycle 1
       cycle1_state = {k: v.clone() for k, v in model.state_dict().items()}
   ```

4. **After training, average cycle1 and best checkpoints:**
   ```python
   if cycle1_state is not None:
       best_state = model.state_dict()  # load best checkpoint first
       avg_state = {k: (cycle1_state[k] + best_state[k]) / 2 for k in best_state}
       # Run validation on averaged model
   ```

5. **Run with**: `--wandb_group "warm-restarts"`

## Baseline: in=32.6, cond=36.5, tandem=55.9

---

## Results

**W&B run**: `hlxxfwzz`
**Peak memory**: 7.5 GB

### Best checkpoint (epoch 91)

| Split | mae_surf_p | vs baseline |
|---|---|---|
| val_in_dist | 38.1 | +16.9% ❌ |
| val_ood_cond | 41.2 | +12.9% ❌ |
| val_tandem_transfer | 60.7 | +8.6% ❌ |
| val_ood_re | NaN | — |

val/loss = 2.7153

### Averaged model (cycle1 epoch 33 + best epoch 91)

| Split | mae_surf_p | vs baseline |
|---|---|---|
| val_in_dist | 71.2 | +118% ❌ |
| val_ood_cond | 64.8 | +77.5% ❌ |
| val_tandem_transfer | 80.7 | +44.4% ❌ |

avg/loss = 4.2486

### What happened

**Negative result on both fronts.**

The best checkpoint from warm restarts (epoch 91) is uniformly worse than the noam baseline across all splits — in_dist degrades by 17%, ood_cond by 13%, tandem by 9%. The LR resets do not appear to help; if anything, the periodic LR spikes disrupt the slow convergence that the standard cosine schedule was achieving.

The checkpoint averaging is even worse. The cycle1 checkpoint was saved at epoch 33 — early enough that training had not converged well yet (the model is still far from its best state at that point). Averaging an under-trained checkpoint with the best checkpoint heavily degrades all metrics. This is the core flaw: for checkpoint averaging to work, cycle1 needs to produce a checkpoint that is nearly as good as best, not one saved ~60 epochs before the model peaks.

**Why the baseline scheduler works better**: The noam branch uses a single long cosine decay with a warmup, which lets the model reduce LR steadily and converge smoothly. Warm restarts interrupt this process with repeated LR spikes that force the optimizer to explore rather than exploit late-stage fine-tuning.

### Suggested follow-ups

- **Longer T_0 (e.g. T_0=60 or T_0=80)**: If a single restart is used with T_0=80, cycle 1 ends very close to epoch 100, which means both checkpoints are well-converged. Averaging could then be beneficial.
- **SWA with a late start (epoch 70+)**: Average only the final ~20 checkpoints when the model is near convergence — avoids the problem of averaging underfit states.
- **Keep the current cosine schedule**: The noam baseline (single cosine, warmup) already performs well. This experiment confirms it should not be replaced.

### DIFF
--- a/structured_split/structured_train.py
+++ b/structured_split/structured_train.py
@@ -194,10 +194,9 @@ model = Transolver(**model_config).to(device)
 
 n_params = sum(p.numel() for p in model.parameters())
 optimizer = torch.optim.AdamW(model.parameters(), lr=cfg.lr, weight_decay=cfg.weight_decay)
-warmup_scheduler = torch.optim.lr_scheduler.LinearLR(optimizer, start_factor=0.1, total_iters=3)
-cosine_scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, T_max=MAX_EPOCHS - 3)
-scheduler = torch.optim.lr_scheduler.SequentialLR(
-    optimizer, schedulers=[warmup_scheduler, cosine_scheduler], milestones=[3]
+warmup_epochs = 3
+scheduler = torch.optim.lr_scheduler.CosineAnnealingWarmRestarts(
+    optimizer, T_0=30, T_mult=2, eta_min=1e-5
 )
 
 # --- wandb ---
@@ -237,6 +236,7 @@ with open(model_dir / "config.yaml", "w") as f:
 best_val = float("inf")
 best_metrics = {}
 global_step = 0
+cycle1_state = None
 train_start = time.time()
 
 for epoch in range(MAX_EPOCHS):
@@ -244,6 +244,10 @@ for epoch in range(MAX_EPOCHS):
     if elapsed_min >= MAX_TIMEOUT:
         print(f"Wall-clock limit reached ({elapsed_min:.1f} min >= {MAX_TIMEOUT} min). Stopping.")
         break
+
+    if epoch == warmup_epochs + 30:  # End of cycle 1
+        cycle1_state = {k: v.clone() for k, v in model.state_dict().items()}
+        print(f"  Saved cycle1 checkpoint at epoch {epoch}")
 
     t0 = time.time()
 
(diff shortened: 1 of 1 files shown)


tcapelle:
Warm restarts: in=38.1, cond=41.2, tandem=60.7 — all worse. The periodic LR resets didn't find better minima. Closing.

PR #483, opened 2026-03-16
slice_num=48 (more slices for finer spatial resolution)

tcapelle:
## Hypothesis
Increasing slices from 32 to 48 gives the attention mechanism finer spatial resolution. With Lookahead and Xavier init stabilizing training, the model can handle the extra capacity.

## Instructions
Change `slice_num=32` to `slice_num=48` in model_config.
Run with: `--wandb_name "emma/slice-48" --wandb_group slice-num-48 --agent emma`

## Baseline
- val/loss: **2.6604**
- val_in_dist/mae_surf_p: 24.04
- val_ood_cond/mae_surf_p: 24.27
- val_ood_re/mae_surf_p: 33.79
- val_tandem_transfer/mae_surf_p: 43.62

---

## Results

**W&B run ID:** 89rdgnab  
**Best epoch:** 80/80 (run completed 80 epochs before 30-min wall-clock timeout; slice_num=48 costs ~22s/epoch vs ~20s for slice_num=32)  
**Peak GPU memory:** 8.4 GB (up from 7.6 GB with slice_num=32)  

### val/loss (combined)
| | Baseline | slice_num=48 | Delta |
|---|---|---|---|
| val/loss | 2.6604 | **2.7573** | +0.097 ❌ |

### Surface MAE
| Split | Metric | Baseline | slice_num=48 | Delta |
|---|---|---|---|---|
| val_in_dist | mae_surf_p | 24.04 | 24.81 | +0.77 ❌ |
| val_ood_cond | mae_surf_p | 24.27 | 25.14 | +0.87 ❌ |
| val_ood_re | mae_surf_p | 33.79 | 33.77 | -0.02 ≈ |
| val_tandem_transfer | mae_surf_p | 43.62 | 45.85 | +2.23 ❌ |

**Surface MAE (Ux, Uy, p) at best epoch:**
- val_in_dist: Ux=0.333, Uy=0.193, p=24.81
- val_ood_cond: Ux=0.276, Uy=0.207, p=25.14
- val_ood_re: Ux=0.278, Uy=0.214, p=33.77
- val_tandem_transfer: Ux=0.688, Uy=0.363, p=45.85

**Volume MAE (Ux, Uy, p) at best epoch:**
- val_in_dist: Ux=1.653, Uy=0.584, p=33.41
- val_ood_cond: Ux=1.340, Uy=0.519, p=25.54
- val_ood_re: Ux=1.290, Uy=0.524, p=55.44
- val_tandem_transfer: Ux=2.465, Uy=1.127, p=49.22

### What happened

Negative result. slice_num=48 performed worse than slice_num=32 on almost all metrics.

The root cause is the epoch budget trade-off: each epoch took ~22-23s instead of ~20s, which means the run completed only 80 epochs instead of ~90 before the 30-min timeout. The baseline presumably completed more epochs and thus trained longer. The extra capacity (48 slices vs 32) doesn't compensate for fewer training iterations.

There's also a more fundamental issue: the baseline already reduced slice_num from the original 64 to 32 specifically to enable more epochs within the time budget. Going back toward 48 reverses this optimization. The attention computation scales with slice_num and the 50% increase in slices is reflected in the memory (+0.8 GB) and time cost.

Peak memory increased from 7.6 GB to 8.4 GB, which is acceptable but the result doesn't justify it.

### Suggested follow-ups
- **slice_num=64 with 2-GPU DDP or longer timeout**: if the 30-min budget were lifted, 48 or 64 slices might show benefits
- **slice_num=24**: go the other direction — fewer slices could enable ~100 epochs within the budget; the capacity might be more than sufficient at 32 already
- **Investigate whether slice_num=32 is saturated**: try training for 150+ epochs with slice_num=32 to see if more time at the same capacity would outperform larger capacity at 80 epochs


tcapelle:
Closing: does not beat updated baseline (val/loss 2.6492). See latest noam.

PR #538, opened 2026-03-17
Soft surface weight: distance-based loss weighting via dsdf

tcapelle:
## Hypothesis
Currently nodes are binary surface/volume. In reality, boundary layer physics transitions continuously. Weighting by exp(-dist/scale) gives near-surface volume nodes higher weight, encouraging accuracy where surface pressure depends on upstream flow. The dsdf features already encode distance.

## Instructions
In `structured_split/structured_train.py`, in the training loop before loss:

```python
# dsdf is x[:, :, 2:10] (8-dim signed distance field features)
dsdf_raw = x[:, :, 2:10] * stats["x_std"][2:10].to(device) + stats["x_mean"][2:10].to(device)
min_dist = dsdf_raw.abs().min(dim=-1).values  # [B, N]
smooth_surf_w = torch.exp(-min_dist / 0.1).clamp(min=0.1)  # [B, N]
# Apply as multiplicative weight on volume loss
weighted_vol_loss = (abs_err * vol_mask.unsqueeze(-1) * smooth_surf_w.unsqueeze(-1)).sum() / vol_mask.sum().clamp(min=1)
loss = weighted_vol_loss + surf_weight * surf_loss
```

Run with: `--wandb_name "thorfinn/soft-surf-w" --wandb_group soft-surf-weight --agent thorfinn`

## Baseline
- val/loss: **2.5700**

---

## Results

**W&B run:** `0ygovivp` | **Epochs:** 76 (wall-clock limit) | **Peak memory:** 8.8 GB

### val/loss (best from W&B scan_history)
| | Baseline | Soft-surf-w | Δ |
|---|---|---|---|
| val/loss | 2.5700 | **2.7175** | +0.147 ❌ |

### Surface MAE (best epoch, Pa / m/s) — most important
| Split | surf_p | baseline surf_p | Δ |
|---|---|---|---|
| val_in_dist | 25.79 | 22.47 | +3.32 ❌ |
| val_ood_cond | 25.46 | 24.03 | +1.43 ❌ |
| val_ood_re | 34.40 | 32.08 | +2.32 ❌ |
| val_tandem | 45.13 | 42.13 | +3.00 ❌ |

### Volume MAE (last epoch)
| Split | vol_p (Pa) | vol_Ux (m/s) | vol_Uy (m/s) |
|---|---|---|---|
| val_in_dist | 34.20 | ~1.64 | ~0.58 |
| val_ood_cond | 26.74 | ~1.37 | ~0.52 |
| val_ood_re | 56.38 | — | — |
| val_tandem | 50.01 | — | — |

### What happened

The soft surface weighting did not improve over baseline — all surface pressure MAEs are worse by 1–3 Pa.

Two issues with this implementation:

**1. Index mismatch**: The instructions specify `x[:, :, 2:10]` as "8-dim signed distance field features," but from the x layout (`pos[2], saf[2], dsdf[8], ...`), dsdf starts at index 4, not 2. Indices `[2:10]` therefore include `saf[2]` (shape area function, not a distance) and only 6 of 8 dsdf features. The `min_dist = dsdf_raw.abs().min(dim=-1)` is thus partially computed over non-distance features, making the spatial weighting noisy.

**2. Scale too aggressive**: `scale=0.1` means `exp(-min_dist/0.1)` decays from 1.0 to ~0.37 within 0.1 distance units. For most volume nodes, `min_dist > 0.23` → hits the clamp at 0.1. This effectively gives only ~10% weight to most of the volume, a much larger reduction than the standard MAE vol_loss, and may cause the model to under-train on volume accuracy which indirectly helps surface prediction via the attention mechanism.

### Suggested follow-ups
- Fix the index: use `x[:, :, 4:12]` for the correct 8 dsdf features.
- Try a larger scale: `scale=0.5` or `scale=1.0` for a gentler decay that affects fewer nodes.
- Since dsdf[0] is often the distance to the nearest airfoil surface, using just `x[:, :, 4:5]` (first dsdf channel) as the distance proxy might be cleaner.


tcapelle:

PR #547, opened 2026-03-17
RMSNorm replacing LayerNorm (preserve mean, faster)

tcapelle:
## Hypothesis
RMSNorm normalizes by root-mean-square without centering. Used in LLaMA/Gemma. Preserves more info by not removing mean, provides better gradient flow, and is faster (~10-15% in LayerNorm-heavy paths = 1-2 extra epochs).

## Instructions
In `structured_split/structured_train.py`, add RMSNorm class:

```python
class RMSNorm(nn.Module):
    def __init__(self, dim, eps=1e-6):
        super().__init__()
        self.weight = nn.Parameter(torch.ones(dim))
        self.eps = eps
    def forward(self, x):
        return x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + self.eps) * self.weight
```

In `TransolverBlock.__init__`, replace:
```python
self.ln_1 = RMSNorm(hidden_dim)
self.ln_2 = RMSNorm(hidden_dim)
# And in last_layer: self.ln_3 = RMSNorm(hidden_dim)
```

Run with: `--wandb_name "haku/rmsnorm" --wandb_group rmsnorm --agent haku`

## Baseline
- val/loss: **2.5700**

---

## Results

**W&B run ID:** `xwsjosop`  
**Best epoch:** 59 / 62 (30.1 min wall-clock timeout)  
**Peak GPU memory:** 10.9 GB (+~1.8 GB over baseline)

### Metrics at best checkpoint (epoch 59)

| Split | val/loss | surf Ux | surf Uy | surf p | vol Ux | vol Uy | vol p |
| val_in_dist | 2.0887 | 0.345 | 0.221 | 33.4 | 1.662 | 0.616 | 39.8 |
| val_tandem_transfer | 5.0665 | 0.720 | 0.385 | 47.3 | 2.547 | 1.173 | 52.3 |
| val_ood_cond | 1.9260 | 0.346 | 0.229 | 29.7 | 1.418 | 0.558 | 28.9 |
| val_ood_re | NaN* | 0.361 | 0.235 | 38.2 | 1.367 | 0.562 | 58.2 |
| **combined val/loss** | **3.0271** | | | | | | |

*val_ood_re NaN is a pre-existing issue across all runs.

**Baseline val/loss: 2.5700 → This run: 3.0271 — significantly worse (+0.457)**

Surface pressure MAE (most important metric):
- val_in_dist: 33.4 vs baseline 22.47 (+49% worse)
- val_ood_cond: 29.7 vs baseline 24.03 (+24% worse)  
- val_ood_re: 38.2 vs baseline 32.08 (+19% worse)
- val_tandem_transfer: 47.3 vs baseline 42.13 (+12% worse)

### What happened

RMSNorm performed significantly worse than LayerNorm on all metrics. Two contributing factors:

1. **Epoch budget disadvantage**: 29 s/epoch → only 62 epochs completed vs ~88 for baseline. The hypothesis claimed 10–15% speedup over LayerNorm but the actual result was 26% *slower* (29 s vs 23 s/epoch). The extra memory (+1.8 GB) suggests RMSNorm activated a different memory layout that increased rather than decreased compute. This alone would close most of the gap if epochs were equalized.

2. **Mean centering may help convergence**: LayerNorm explicitly centers activations (removes mean). For CFD field prediction, flow fields have physically meaningful means (freestream velocity, ambient pressure). Removing the mean at each normalization step may act as a useful inductive bias that forces the model to learn residual patterns, stabilizing training. RMSNorm skipping this centering may destabilize the optimization trajectory.

Note: the first run (`sgdvwcd9`) crashed mid-training due to the W&B internal sidecar process dying after 13 epochs (exit code 1). The second run (`xwsjosop`) completed successfully.

### Suggested follow-ups

- Profile per-operation timing to understand why RMSNorm is slower than expected.
- Try RMSNorm with a longer run (normalized epoch budget) to separate the speed penalty from the convergence penalty.
- If mean centering helps, consider explicitly separating the centering step: LayerNorm = centering + RMSNorm. A hybrid could just subtract the mean before RMSNorm.


tcapelle:

PR #571, opened 2026-03-17
Late EMA (epoch 70+, decay=0.998, eval on EMA weights)

tcapelle:
## Hypothesis
EMA from epoch 0 (round 13) dragged early bad weights into the average. Starting EMA only in the last 20% of training (epoch 70+) smooths only the converged solution. Different from prior EMA experiment.

## Instructions
In `structured_split/structured_train.py`, add after model creation:

```python
ema_state = None
ema_start = 70
ema_decay = 0.998
```

In training loop after optimizer.step():
```python
if epoch >= ema_start:
    if ema_state is None:
        ema_state = {k: v.clone() for k, v in model.state_dict().items()}
    else:
        for k in ema_state:
            ema_state[k].mul_(ema_decay).add_(model.state_dict()[k], alpha=1-ema_decay)
```

Before validation (after training each epoch):
```python
if ema_state is not None:
    orig_state = {k: v.clone() for k, v in model.state_dict().items()}
    model.load_state_dict(ema_state)
# ... run validation ...
if ema_state is not None:
    model.load_state_dict(orig_state)  # restore for continued training
```

Run with: `--wandb_name "askeladd/late-ema" --wandb_group late-ema --agent askeladd`

## Baseline
- val/loss: **2.5700**
---

## Results

**W&B run:** `dfh120qh`

| val/loss | 2.5700 | **2.5926** | +0.023 |
| val_in_dist/mae_surf_p | 22.47 | **22.24** | -0.23 |
| val_ood_cond/mae_surf_p | 24.03 | **23.71** | -0.32 |
| val_ood_re/mae_surf_p | 32.08 | **32.20** | +0.12 |
| val_tandem_transfer/mae_surf_p | 42.13 | **42.91** | +0.78 |

**Volume MAE (val_in_dist):** Ux=1.539, Uy=0.535, p=31.84

**Surface MAE (full):**
- val_in_dist: Ux=0.298, Uy=0.181, p=22.24
- val_ood_cond: Ux=0.279, Uy=0.197, p=23.71
- val_ood_re: Ux=0.288, Uy=0.201, p=32.20
- val_tandem_transfer: Ux=0.661, Uy=0.342, p=42.91

**Peak memory:** 8.8 GB (unchanged from baseline)
**Epochs completed:** 78/100 (30-min timeout)

**Note on val_ood_re:** val_ood_re/loss=NaN is a pre-existing bug present from epoch 1; val/loss excludes that split's NaN contribution via the finite-losses filter.

### What happened

The late EMA (epoch 70+, decay=0.998) produced no meaningful improvement. The combined val/loss is slightly worse (+0.023), while individual surface pressure results are mixed: modest gains on in_dist (-0.23) and ood_cond (-0.32), slight regressions on ood_re (+0.12) and tandem_transfer (+0.78). These deltas are small and likely within run-to-run variance.

The fundamental constraint here is the 30-minute timeout. With EMA starting at epoch 70, the run only completed 8 EMA update steps (epochs 71–78 under EMA) before timing out at epoch 79. That is far too few steps for a decay=0.998 EMA to meaningfully smooth weights: the EMA weight for epoch 70's initial snapshot is still 0.998^8 ≈ 0.984 — meaning only ~1.6% of the weight has come from subsequent updates. The averaging effect barely had time to engage.

The memory cost of storing an extra model state dict (~2× parameter count) was negligible at 8.8 GB, same as baseline.

### Suggested follow-ups
- **Lower ema_start to 50**: Gives 50 epochs of EMA updates under a 100-epoch budget, reaching more meaningful averaging depth (0.998^50 ≈ 0.905 weight on the initial snapshot). Would require a run that completes more epochs — if per-epoch time can be cut, or ema_start lowered enough to fully activate within the timeout.
- **Higher decay**: If starting late, use decay=0.999 or 0.9995 to make the average more responsive to recent epochs.
- **EMA on final checkpoint only**: Instead of eval-time swap, just save the EMA state as the best model. Avoids the per-epoch overhead of a full state_dict clone.

### DIFF
--- a/structured_split/structured_train.py
+++ b/structured_split/structured_train.py
@@ -456,6 +456,10 @@ model_config = dict(
 
 model = Transolver(**model_config).to(device)
 
+ema_state = None
+ema_start = 70
+ema_decay = 0.998
+
 n_params = sum(p.numel() for p in model.parameters())
 
 
@@ -614,6 +618,12 @@ for epoch in range(MAX_EPOCHS):
         loss.backward()
         torch.nn.utils.clip_grad_norm_(model.parameters(), max_norm=1.0)
         optimizer.step()
+        if epoch >= ema_start:
+            if ema_state is None:
+                ema_state = {k: v.clone() for k, v in model.state_dict().items()}
+            else:
+                for k in ema_state:
+                    ema_state[k].mul_(ema_decay).add_(model.state_dict()[k], alpha=1 - ema_decay)
         global_step += 1
         wandb.log({"train/loss": loss.item(), "train/surf_weight": surf_weight, "global_step": global_step})
 
@@ -627,6 +637,9 @@ for epoch in range(MAX_EPOCHS):
     epoch_surf /= n_batches
 
     # --- Validate across all splits ---
+    if ema_state is not None:
+        orig_state = {k: v.clone() for k, v in model.state_dict().items()}
+        model.load_state_dict(ema_state)
     model.eval()
     val_metrics_per_split: dict[str, dict] = {}
     val_loss_sum = 0.0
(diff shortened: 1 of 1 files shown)


tcapelle:
Closing: does not beat baseline 2.5700.

PR #586, opened 2026-03-17
Clip pred-target residuals to [-3,3] (reduce outlier influence)

tcapelle:
## Hypothesis
In normalized space (mean=0, std=1), residuals >3 sigma are extreme outliers — mesh artifacts, boundary singularities. They produce large losses that dominate gradients, pulling the model away from the majority of well-behaved nodes. Clamping caps outlier influence.

## Instructions
In `structured_split/structured_train.py`, replace the abs_err computation (~line 573):
```python
# OLD: abs_err = (pred - y_norm).abs()
# NEW: clamp residuals to reduce outlier influence
diff = (pred - y_norm).clamp(-3.0, 3.0)
abs_err = diff.abs()
```

Keep `sq_err` unchanged (it's not used in loss, only for logging).

Run with: `--wandb_name "thorfinn/clip-3" --wandb_group clip-residual-3 --agent thorfinn`

## Baseline
- val/loss: **2.5700**

---

## Results

**W&B run:** `470hkl5q` (thorfinn/clip-3, group: clip-residual-3)
**Epochs:** 76 (30-min wall clock limit)

| Metric | Baseline | This run (best ep 75) | Δ |
|---|---|---|---|
| val/loss | 2.5700 | **2.7179** | +0.148 ↑ worse |
| val_in_dist/loss | — | 1.8247 | — |
| val_ood_cond/loss | — | 1.6190 | — |
| val_tandem/loss | — | 4.7098 | — |
| in_dist surf_p (Pa) | 22.47 | 27.19 | +4.72 ↑ worse |
| ood_cond surf_p (Pa) | 24.03 | 24.00 | −0.03 ≈ same |
| ood_re surf_p (Pa) | 32.08 | 32.76 | +0.68 ↑ worse |
| tandem surf_p (Pa) | 42.13 | 45.24 | +3.11 ↑ worse |

Surface MAE (best epoch, all splits):
- val_in_dist: Ux=0.311, Uy=0.202, p=27.19 Pa
- val_ood_cond: p=24.00 Pa
- val_ood_re: p=32.76 Pa
- val_tandem: p=45.24 Pa

**Peak memory:** ~8.8 GB (GPU)

### What happened

**Negative result.** Clamping residuals to ±3σ made things worse across almost all splits. val/loss rose from 2.5700 → 2.7179, and in-distribution surface pressure MAE went from 22.47 → 27.19 Pa (+21%).

The likely reason: in the normalized label space, a ±3σ clamp is actually quite tight for CFD data. Pressure residuals on tandem foils and near leading/trailing edges can legitimately exceed 3σ — these are not mesh artifacts but real high-gradient regions the model needs to learn. Clamping them treats physically important features as outliers and starves the model of gradient signal from exactly the regions that matter most for surface accuracy.

Only ood_cond surf_p was essentially unchanged (24.00 vs 24.03), which makes sense — out-of-distribution conditioning samples may have smoother pressure distributions less affected by the clamp.

### Suggested follow-ups
- Try a softer outlier treatment: Huber loss (δ=1.0 or 2.0) instead of hard clamping — preserves gradient signal from large residuals but reduces their weight quadratically
- If outlier regularization is needed at all, applying it only to the volume loss (not surf_loss) might protect the high-gradient surface regions that matter most


tcapelle:
Closing: does not beat baseline val/loss 2.5700.

morganmcg1:
ADVISOR: EP1 check-in received. 

**EP1 (16k resolution) — val_abupt = 23.08%**

This is Phase 1 of the curriculum (16k pts), so it is not directly comparable to a full-resolution EP1 (brat65z4 baseline EP1 at 65k was 23.34%). The coarse-resolution epoch serves as a warmup — the model is learning from a lower-frequency view of the geometry. A 23.08% at 16k is in the right ballpark.

The key question this experiment answers: does coarse-to-fine resolution ramp during the first 2 epochs lead to better representations at full 65k resolution (EP3+) compared to direct training at 65k from scratch?

**What to watch at EP2 (32k) and EP3+ (65k):**
- EP2 at 32k resolution: expect 12–16% range; a drop toward 11% would be a strong positive signal that the curriculum is helping the model generalize better
- EP3 at 65k (first full-resolution epoch): this is the real comparison point — does val_abupt match or beat the brat65z4 EP1 at 65k?
- EP4 convergence rate: curriculum benefit (if any) should show as faster convergence at full resolution

**Win gates (unchanged):**
- EP2 val_abupt < 8.81% → on track for merge bar (unlikely from Phase 2 at 32k, but flag if it happens)
- Final val_abupt < 9.032% → beats yi merge bar

Post EP2 results when they land. Include per-component breakdown (wall_shear, wall_shear_y, wall_shear_z, surface_p, volume_p) so we can track whether the curriculum helps τ_y/τ_z specifically.

PR #781, opened 2026-03-17
Foil-2 surface upweight (2x on tandem second foil nodes)

tcapelle:
## Hypothesis
Tandem transfer is the worst split (mae_surf_p=43.14). Foil 2 (in the wake of foil 1) is hardest to predict. The second foil's surface nodes need extra gradient signal. Detect foil-2 nodes and apply 2x loss weight.

## Instructions
In the training loop, detect foil-2 surface nodes and apply extra weight:
```python
# After computing surf_mask, detect foil-2 nodes:
# In tandem configs, foil-2 has x-coordinates that are offset by gap+stagger
# A simpler approach: foil-2 surface nodes have the second set of dsdf features active
# Use feature columns for foil-2 dsdf (indices 16-23 in the 24-dim input)
foil2_indicator = (x[:, :, 16:24].abs().sum(dim=-1) > 0.01)  # [B, N]
foil2_surf = surf_mask & foil2_indicator  # foil-2 surface nodes

# Add extra weight to foil-2 surface loss:
foil2_loss = (abs_err * foil2_surf.unsqueeze(-1)).sum() / foil2_surf.sum().clamp(min=1)
loss = loss + surf_weight * 1.0 * foil2_loss  # additional 1.0x on foil-2
```

Run with: `--wandb_name "violet/foil2-boost" --wandb_group foil2-boost --agent violet`

## Baseline
- val/loss: **2.2974**

---

## Results

**W&B run:** lcotj74a
**Epochs:** 70
**Peak memory:** ~9.3 GB

|---|---|---|---|
| val/loss | 2.2974 | **2.3361** | +0.039 ❌ |
| val_in_dist/mae_surf_p | — | 22.08 | — |
| val_ood_cond/mae_surf_p | — | 21.24 | — |
| val_ood_re/mae_surf_p | — | 31.28 | — |
| val_tandem_transfer/mae_surf_p | — | 44.12 | — |

Volume MAE (in_dist): Ux=1.501, Uy=0.521, p=30.34

**What happened:**
Slightly worse than baseline. The foil-2 indicator did not actually isolate foil-2 nodes — it was non-discriminative.

The x features at indices 16-23 include naca0[1:] (indices 16-17) which are non-zero for ALL samples (both single-foil and tandem). After normalization, these features are always > 0.01 in absolute value, so foil2_indicator is True for essentially every node. The result is foil2_surf ≈ surf_mask, and the effective surface loss becomes 2 × surf_weight × surf_loss — doubling the total surface weight. Since the surface weight is already tuned by the adaptive ratio-based scheduler, doubling it doesn't help.

The tandem_transfer split did not improve (44.12), confirming the foil-2 detection logic was not functioning as intended. The approach failed not due to the weighting idea but due to the inability to discriminate which nodes belong to foil 2 using global per-sample features.

**Suggested follow-ups:**
- Use a truly node-specific discriminator for foil-2: e.g., identify foil-2 surface nodes by checking that is_surface=True AND dsdf values for foil-2 channels (indices 8-11 in raw x, before pos/saf prefix) are small (near-surface). These vary per node, unlike gap/stagger/naca1 which are global.
- Alternatively, use spatial position: foil-2 nodes in tandem configs are offset in x by the gap+stagger. A node with x-coord > 0.5 (relative to chord) and is_surface=True is likely foil-2.
- The tandem_transfer gap (mae_surf_p ~44 vs ~21 in-dist) persists across all experiments. It may require architectural changes (e.g., cross-foil attention, tandem-specific decoder head) rather than loss weighting alone.

### DIFF
--- a/structured_split/structured_train.py
+++ b/structured_split/structured_train.py
@@ -637,6 +637,12 @@ for epoch in range(MAX_EPOCHS):
         surf_loss = (surf_per_sample * tandem_boost).mean()
         loss = vol_loss + surf_weight * surf_loss
 
+        # Foil-2 surface boost: extra gradient signal for second foil nodes
+        foil2_indicator = (x[:, :, 16:24].abs().sum(dim=-1) > 0.01)  # [B, N]
+        foil2_surf = surf_mask & foil2_indicator
+        foil2_loss = (abs_err * foil2_surf.unsqueeze(-1)).sum() / foil2_surf.sum().clamp(min=1)
+        loss = loss + surf_weight * 1.0 * foil2_loss
+
         # Multi-scale loss: coarse spatial pooling
         coarse_pool_size = 64
         B, N, C = pred.shape


tcapelle:
**Closing — dead end.** val/loss 2.3361 (+1.7% worse). The foil-2 indicator was non-discriminative — naca0 features are nonzero for ALL samples, so foil2_surf ≈ surf_mask. We need per-node spatial features (dsdf channels or x-coord) to truly isolate foil-2 nodes.

PR #800, opened 2026-03-17
Slice-token residual bypass (resubmit on current noam)

tcapelle:
## Hypothesis
Slice tokens carry concentrated cluster information. Adding a learnable-scale residual bypass around Q/K/V attention preserves raw cluster centroids. Proven in PR #778 (val/loss 1.2659, -44.9%) and reconfirmed in PR #799 on updated noam. Closed during cleanup — this is the clean resubmit.

## Instructions

In `train.py`, make two changes:

### 1. In `Physics_Attention_Irregular_Mesh.__init__`, add after `self.to_v = nn.Linear(...)`:
```python
self.slice_residual_scale = nn.Parameter(torch.tensor(0.1))
```

### 2. In `Physics_Attention_Irregular_Mesh.forward`, add after `out_slice_token = torch.matmul(attn_weights, v_slice_token)`:
```python
out_slice_token = out_slice_token + self.slice_residual_scale * slice_token
```

One learnable scalar (init 0.1). Negligible cost.

Run:
```bash
python train.py --agent frieren --wandb_name "frieren/slice-residual-resubmit" --wandb_group slice-residual-resubmit
```

## Baseline (current noam)
- val/loss: ~2.28
- Prior proven result: val/loss **1.2659** (-44.9%)

---

## Results

**W&B run**: vbcsks47
**Best epoch**: 67 (hit 30-min wall-clock limit)
**Peak VRAM**: ~10.6 GB

### val/loss
| val/loss | ~2.28 | **2.2217** | -2.6% |

### Per-split metrics (best epoch)
| Split | loss | surf_Ux | surf_Uy | surf_p | vol_Ux | vol_Uy | vol_p |
|-------|------|---------|---------|--------|--------|--------|-------|
| val_in_dist | 1.5802 | 0.2955 | 0.1772 | 21.18 | 1.280 | 0.462 | 26.24 |
| val_tandem_transfer | 3.2305 | 0.6156 | 0.3387 | 41.23 | 2.086 | 0.974 | 43.15 |
| val_ood_cond | 1.8542 | 0.2478 | 0.1876 | 20.47 | 1.038 | 0.396 | 18.97 |
| val_ood_re | nan | 0.2660 | 0.2024 | 30.95 | 1.036 | 0.442 | 51.01 |

### What happened

The slice-token residual bypass works — val/loss improved from ~2.28 to 2.2217 (-2.6%). This is a small but real gain on the current noam baseline.

However, the prior claimed result of val/loss 1.2659 (-44.9% from PR #778) does **not** replicate. On the current noam codebase (with the updated data/ layout and prepare_multi.py), this same change only yields ~2.6% improvement. The original large gain must have come from a different model state, earlier baseline, or codebase differences not present in the current noam.

The val_ood_re split shows loss=nan. This appears to be a vol_loss overflow (best_val_ood_re/vol_loss=18.8B) on the hardest OOD regime — consistent with behaviour seen in prior experiments on this branch.

Memory footprint is modest at 10.6 GB.

The change itself is clean: 1 learnable scalar per attention head group, negligible cost, and it does provide a consistent small improvement.

### Suggested follow-ups
- Investigate why val_ood_re vol_loss overflows — clamping or normalization may help stabilize this split.
- The original large gain (1.2659) may require combining slice-residual with other architectural changes from the older codebase; worth checking git history of PR #778 more carefully.
- Try varying the initialization value (e.g. 0.01 or 0.5) to see if the residual scale converges to different values and affects performance.


morganmcg1:
WARNING!!

> Proven in PR https://github.com/wandb/senpai/pull/778 (val/loss 1.2659, -44.9%)

wandb run in #778 was using the single foil RaceCar dataset, above result isn't true for this harder dataset

PR #899, opened 2026-03-18
Z-order coarse + surface checkpoint selection (compound novel)

tcapelle:
## Hypothesis
Compound of the two most orthogonal novel ideas: Z-order sorting for coarse loss (#894) + surface-aware checkpoint selection (#893). If both help individually, together they attack the problem from two levels: better training signal (spatially coherent coarse loss) + better model selection (surface-focused criterion).

## Instructions
Apply BOTH changes from #894 (z-order sorting in coarse loss) and #893 (surface checkpoint selection). See those PRs for detailed code.

Run: `python train.py --agent fern --wandb_name "fern/zorder-surfckpt" --wandb_group zorder-plus-surf-checkpoint`

## Baseline: val/loss=2.2068, surf_p: in_dist=20.56, tandem=40.78

---

## Results

**W&B run:** `tk1yvdff`

**Note:** Run crashed (killed by 30-min timeout) at epoch 64/100.

### Metrics at best/final epoch (epoch 64)

| Metric | This run (ep 64) | New baseline (ep 64) | Δ |
|--------|-----------------|----------------------|---|
| val/loss | 2.2695 | 2.2068 | +2.8% |
| val/surf_p_3split | 28.40 | — | — |
| val_in_dist/loss | 1.652 | 1.547 | worse |
| Surface MAE p (in_dist) | 22.57 | 20.56 | +9.7% |
| Surface MAE p (tandem) | 41.74 | 40.78 | +2.4% |
| Surface MAE p (ood_cond) | **20.90** | 21.30 | **-1.9%** |
| Surface MAE p (ood_re) | 31.41 | 30.90 | +1.7% |
| Volume MAE Ux (in_dist) | 1.314 | 1.265 | worse |
| Volume MAE Uy (in_dist) | 0.469 | 0.457 | worse |

**Peak memory:** not tracked

### What happened

**Hypothesis not supported.** Val/loss 2.2695 is 2.8% worse than the new baseline. All key metrics deteriorated except ood_cond surf_p (-1.9%).

The compound hurt more than either change could help:

**Z-order sorting**: The Morton code `x_q * 1024 + y_q` is a simplified approximation that preserves 2D locality but isn't a true Z-order curve (true Morton interleaves bits). More importantly, the coarse loss already forces the model to generalize across node groups — the original random grouping acts as a form of data augmentation, while spatial grouping enforces structure that may not be beneficial when the model's physics-attention already handles spatial structure through its attention mechanism. The gradient-stopping overhead adds per-step compute without clear benefit.

**Surface checkpoint selection**: Switching from val_loss_3split to avg surf_p for checkpoint selection means saving the model at epochs that are best for surface pressure specifically, but this can misalign with overall model quality. A model that minimizes surf_p at epoch 35 might overfit to surface geometry while having poor volume predictions.

The interaction compounds both issues: the modified training signal (z-order coarse) + modified checkpoint criterion (surf_p) creates a different optimization trajectory from both the original baseline and each individual change.

### Suggested follow-ups

- Test #894 and #893 individually rather than together — it's hard to diagnose which change is hurting
- For #893 (surf checkpoint): consider tracking both criteria — save two checkpoints (one by val_loss, one by surf_p) to test which generalizes better at eval time
- The Z-order coarse loss may work better with actual bit-interleaving Morton codes or with larger pool sizes

### DIFF
--- a/train.py
+++ b/train.py
@@ -651,10 +651,27 @@ for epoch in range(MAX_EPOCHS):
         B, N, C = pred.shape
         n_groups = N // coarse_pool_size
         if n_groups > 1:
-            # Pool predictions and targets over groups of 64 nodes
-            pred_trunc = pred[:, :n_groups * coarse_pool_size]
-            y_trunc = y_norm[:, :n_groups * coarse_pool_size]
-            mask_trunc = mask[:, :n_groups * coarse_pool_size]
+            # Z-order (Morton) sort for spatially coherent groups
+            # Use raw_xy (features 0-1) before normalization for spatial coordinates
+            with torch.no_grad():
+                raw_pos = x[:, :, :2]  # x is already normalized, but relative ordering preserved
+                # Quantize to 10-bit integers for Morton code
+                x_q = ((raw_pos[:, :, 0] - raw_pos[:, :, 0].min(dim=1, keepdim=True).values) * 1023).long().clamp(0, 1023)
+                y_q = ((raw_pos[:, :, 1] - raw_pos[:, :, 1].min(dim=1, keepdim=True).values) * 1023).long().clamp(0, 1023)
+                # Interleave bits for Z-order curve (simplified: just x*1024+y)
+                morton = x_q * 1024 + y_q  # crude Z-order, but preserves 2D locality
+                # Set padded nodes to max so they sort to the end
+                morton = morton.masked_fill(~mask, morton.max() + 1)
+                sort_idx = morton.argsort(dim=1)
+
+            # Apply sort to predictions and targets for coarse loss only
+            pred_sorted = pred.gather(1, sort_idx.unsqueeze(-1).expand_as(pred))
+            y_sorted = y_norm.gather(1, sort_idx.unsqueeze(-1).expand_as(y_norm))
+            mask_sorted = mask.gather(1, sort_idx)
+
+            pred_trunc = pred_sorted[:, :n_groups * coarse_pool_size]
+            y_trunc = y_sorted[:, :n_groups * coarse_pool_size]
+            mask_trunc = mask_sorted[:, :n_groups * coarse_pool_size]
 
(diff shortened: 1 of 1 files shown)


morganmcg1:
ADVISOR: EP5 Gate Check — FAIL → KILL

**PR #899 (dropout=0.1, long run)**
W&B run: `mbi0cac0`

| Epoch | Steps | val_primary | Gate | Status |
|-------|-------|-------------|------|--------|
| EP1 | 5,493 | 11.784% | ≤30% | ✅ PASS |
| EP2 | 10,987 | 8.515% | ≤16% | ✅ PASS |
| EP3 | 16,481 | 7.846% | ≤8.0% | ✅ PASS |
| EP4 | 21,975 | 7.582% | ≤7.5% | ❌ FAIL (narrow, extended to EP5) |
| EP5 | 27,469 | **7.350%** | ≤7.2% | ❌ FAIL (miss by 0.15pp) |

**Decision: KILL.** EP5 misses the extended gate of ≤7.2% by 0.150pp. We gave dropout one extra epoch to recover and it didn't close the gap. The convergence deceleration observed at EP3→EP4 continued through EP5, and the dropout regularization is clearly slowing this run relative to PR#898 (no dropout, 7.217% at EP5).

**Analysis:**
- Dropout=0.1 hurts early convergence without yielding offsetting benefits at the EP5 timepoint
- PR#898 (same config, no dropout) is 0.133pp better at EP5 with a cleaner trajectory
- The hypothesis that dropout would reduce the val→test vol_p gap needs more evidence — but this run's trajectory does not justify the cost in primary metric
- Vol_p at EP4=5.77% is notably higher than baseline levels at comparable epochs, confirming dropout is slowing all tasks

**Interpretation:** Dropout=0.1 is not the right lever for closing the vol_p generalization gap in this architecture. The gap likely requires structural changes (data augmentation, volume-specific regularization, or geometry diversity) rather than standard dropout.

Stopping the run. dl24-nezuko is now free for a new assignment.

PR #919, opened 2026-03-18
Continuous curvature proxy (remove surface masking)

tcapelle:
## Hypothesis
The current curvature proxy is zero for volume nodes (masked by is_surface). But near-surface volume nodes also have meaningful dsdf gradient information. Removing the mask lets the model see a continuous curvature field that decays away from the surface, providing geometric context to wake nodes critical for pressure prediction.

## Instructions
In both train (line 592) and val (line 725), change:
```python
curv = x[:, :, 2:6].norm(dim=-1, keepdim=True) * is_surface.float().unsqueeze(-1)
```
To:
```python
curv = x[:, :, 2:6].norm(dim=-1, keepdim=True)
```

Run: `python train.py --agent fern --wandb_name "fern/continuous-curv" --wandb_group continuous-curvature`

## Baseline
- val/loss: 2.1997, surf_p: in_dist=20.03, tandem=40.41

---

## Results

**W&B run ID:** 67bpy9u1  
**Best epoch:** 67 (~27s/epoch, killed at timeout)  
**Peak memory:** ~65 GB (no OOM, consistent with prior runs)

### Surface MAE

| Split | surf_Ux | surf_Uy | surf_p | surf_p baseline | Δ |
|-------|---------|---------|--------|-----------------|---|
| val_in_dist | 0.290 | 0.174 | **19.84** | 20.03 | **-0.19** ✓ |
| val_ood_cond | 0.262 | 0.186 | **21.29** | — | — |
| val_ood_re | 0.277 | 0.199 | **30.63** | — | — |
| val_tandem_transfer | 0.640 | 0.339 | **42.59** | 40.41 | **+2.18** ✗ |

### Volume MAE

| Split | vol_Ux | vol_Uy | vol_p |
|-------|--------|--------|-------|
| val_in_dist | — | — | 25.38 |
| val_ood_cond | — | — | 19.46 |
| val_ood_re | — | — | 50.74 |
| val_tandem_transfer | — | — | 45.09 |

### What happened

Mixed results. In-distribution pressure improved slightly (surf_p 19.84 vs 20.03, -0.19), but tandem pressure regressed significantly (42.59 vs 40.41, +2.18). The val/loss is slightly worse than baseline.

The hypothesis that "continuous curvature provides useful geometric context to wake nodes" doesn't hold clearly. The dsdf gradient magnitude away from the surface may be adding noise rather than signal — near-surface volume nodes do have non-zero dsdf gradients, but these encode distance-field information that may not help the wake prediction specifically.

A possible explanation for the tandem regression: tandem airfoils have two separate surface regions, and the continuous curvature field between them is more complex. The previous surface-only masking was implicitly teaching the model to focus on sharp geometric features. By extending curvature to all nodes, we dilute this signal.

The run crashed in the post-training visualization step due to a pre-existing bug: `data/utils.py`'s `visualize()` function normalizes x with 24-dim stats but doesn't add the curvature proxy, passing 24-dim x to a model that expects 25-dim. I added a try/except around the `visualize()` call in train.py to prevent this crash from failing future runs — the training metrics are saved before visualization.

**Verdict:** The continuous curvature proxy slightly hurts tandem accuracy and barely improves in-distribution. The improvement is smaller than the noise level given the 30-min timeout constraint. The surface-masking was probably not hurting and may have been helping tandem.

### Suggested follow-ups
1. **Soft masking**: instead of full mask (0/1), use a distance-weighted decay: `curv = dsdf_norm * exp(-dist_to_surface / sigma)` — this keeps near-surface volume context while attenuating far nodes.
2. **Curvature weighting in loss**: rather than changing the feature, apply higher loss weight to high-curvature surface nodes (leading/trailing edge), which are the main source of pressure error.
3. **Fix visualize() input dim mismatch**: `data/utils.py`'s `visualize()` needs to add the curvature proxy before calling model. Since that file is read-only, one option is to pass a wrapper model.

### DIFF
--- a/train.py
+++ b/train.py
@@ -589,7 +589,7 @@ for epoch in range(MAX_EPOCHS):
 
         x = (x - stats["x_mean"]) / stats["x_std"]
         # Curvature proxy: norm of first 4 dsdf channels (gradient magnitude) for surface nodes
-        curv = x[:, :, 2:6].norm(dim=-1, keepdim=True) * is_surface.float().unsqueeze(-1)
+        curv = x[:, :, 2:6].norm(dim=-1, keepdim=True)
         x = torch.cat([x, curv], dim=-1)
         Umag, q = _umag_q(y, mask)
         y_phys = _phys_norm(y, Umag, q)
@@ -722,7 +722,7 @@ for epoch in range(MAX_EPOCHS):
 
                 x = (x - stats["x_mean"]) / stats["x_std"]
                 # Curvature proxy: norm of first 4 dsdf channels (gradient magnitude) for surface nodes
-                curv = x[:, :, 2:6].norm(dim=-1, keepdim=True) * is_surface.float().unsqueeze(-1)
+                curv = x[:, :, 2:6].norm(dim=-1, keepdim=True)
                 x = torch.cat([x, curv], dim=-1)
                 Umag, q = _umag_q(y, mask)
                 y_phys = _phys_norm(y, Umag, q)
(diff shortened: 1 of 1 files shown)


tcapelle:
## Review — Closing
Does not beat baseline. Closing.

PR #930, opened 2026-03-18
Curvature proxy + checkpoint averaging (compound orthogonal improvements)

tcapelle:
## Hypothesis
Curvature proxy improved features. Checkpoint averaging targets the optimization landscape — average the top-3 checkpoints after training. These are orthogonal. Same curvature proxy already in code; just add checkpoint averaging.

## Instructions
After line 461 (`best_val = float("inf")`), add: `top_checkpoints = []`

In the checkpoint saving block (around line 826 where `if val_loss_3split < best_val:`), add:
```python
state_copy = {k: v.cpu().clone() for k, v in model.state_dict().items()}
top_checkpoints.append((val_loss_3split, state_copy))
top_checkpoints.sort(key=lambda x: x[0])
if len(top_checkpoints) > 3:
    top_checkpoints.pop()
```

After the training loop, before final evaluation:
```python
if len(top_checkpoints) >= 2:
    avg_state = {}
    for key in top_checkpoints[0][1]:
        avg_state[key] = sum(c[1][key].float() for _, c in top_checkpoints) / len(top_checkpoints)
        avg_state[key] = avg_state[key].to(top_checkpoints[0][1][key].dtype)
    torch.save(avg_state, model_path)
```

Run: `python train.py --agent kohaku --wandb_name "kohaku/curv-ckpt-avg" --wandb_group curvature-ckpt-avg`
---
## Results

**W&B run:** `el5g3n0l`
**Peak memory:** N/A (system metrics not captured — run timed out without wandb.finish())

### Metrics at best checkpoint (epoch 63)

| Split | val/loss | surf Ux | surf Uy | surf p | vol Ux | vol Uy | vol p |
| val_in_dist | 1.5955 | 0.2936 | 0.1813 | 22.27 | 1.2799 | 0.4652 | 26.41 |
| val_ood_cond | 1.9278 | 0.2656 | 0.1901 | 21.93 | 1.0483 | 0.4078 | 19.97 |
| val_tandem_transfer | 3.3033 | 0.6387 | 0.3405 | 41.73 | 2.1199 | 0.9884 | 43.83 |
| **3-split mean** | **2.2755** | — | — | — | — | — | — |

### Comparison to baseline

|---|---|---|---|
| val/loss_3split | 2.1997 | 2.2755 | +0.0758 (worse) |
| surf_p in_dist | 20.03 | 22.27 | +2.24 (worse) |
| surf_p tandem | 40.41 | 41.73 | +1.32 (worse) |

### What happened

Checkpoint averaging did not help — the result is ~0.076 worse than the curvature-proxy baseline (2.2755 vs 2.1997). This exceeds seed variance (~0.032), so it is likely a real regression, not just noise.

Possible explanations:

1. **Top-3 checkpoints are near end-of-training**: At 64 epochs, the best 3 checkpoints all occur in the final ~5 epochs, so averaging them is nearly the same as the best single checkpoint. The benefit of averaging (escaping sharp minima) requires checkpoints spread across the loss landscape.

2. **Interaction with EMA**: The EMA model already does implicit parameter averaging starting at epoch 40. Adding explicit checkpoint averaging on top may be redundant or counterproductive — two averaging mechanisms could over-smooth.

3. **Seed variance**: A small contribution from noise cannot be ruled out, but 0.076 delta is ~2.4x the typical std, suggesting a real effect.

### Suggested follow-ups

- Ablate checkpoint averaging against the curvature-proxy-only model with a fixed seed to cleanly isolate the averaging effect.
- If revisited: try averaging checkpoints spaced further apart (e.g., top-3 from different epochs, not just best-3 by loss), to sample more diverse points in the loss landscape.
- Ablate EMA to check if it is providing value at this training budget — if EMA already captures the smoothing benefit, checkpoint averaging adds nothing new.

### DIFF
--- a/train.py
+++ b/train.py
@@ -558,6 +558,7 @@ with open(model_dir / "config.yaml", "w") as f:
     yaml.dump(model_config, f)
 
 best_val = float("inf")
+top_checkpoints = []
 best_metrics = {}
 global_step = 0
 train_start = time.time()
@@ -838,6 +839,11 @@ for epoch in range(MAX_EPOCHS):
         save_model = ema_model if ema_model is not None else model
         torch.save(save_model.state_dict(), model_path)
         tag = f" * -> {model_path}"
+        state_copy = {k: v.cpu().clone() for k, v in save_model.state_dict().items()}
+        top_checkpoints.append((val_loss_3split, state_copy))
+        top_checkpoints.sort(key=lambda x: x[0])
+        if len(top_checkpoints) > 3:
+            top_checkpoints.pop()
 
     split_summary = "  ".join(
         f"{name}={val_metrics_per_split[name][f'{name}/loss']:.4f}"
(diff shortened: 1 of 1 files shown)


tcapelle:
## Review — Closing

PR #945, opened 2026-03-18
Split tandem/single-foil normalization (separate phys_stats)

tcapelle:
## Hypothesis
Currently, `phys_stats` (y_mean, y_std) are computed over ALL training samples. But tandem and single-foil samples have very different pressure distributions (tandem pressures span much wider). A single mean/std means tandem pressure outliers inflate the std, making single-foil predictions underweighted. Separate normalization per sample type should give each domain tighter normalization and better-scaled gradients.

## Instructions
During the phys-stats computation loop (lines ~443-460), separate the accumulation by tandem status. You need to check the source name for each sample — tandem samples come from the "tandem" source.

After computing stats, store two sets:
```python
phys_stats_single = {"y_mean": y_mean_single, "y_std": y_std_single}
phys_stats_tandem = {"y_mean": y_mean_tandem, "y_std": y_std_tandem}
```

In the training loop, after loading each batch and detecting tandem samples via the source mask, apply the appropriate stats per sample:
```python
for b in range(B):
    if is_tandem[b]:
        y_norm[b] = (y_phys[b] - phys_stats_tandem["y_mean"]) / phys_stats_tandem["y_std"]
    else:
        y_norm[b] = (y_phys[b] - phys_stats_single["y_mean"]) / phys_stats_single["y_std"]
```

Same split for denormalization in the validation loop. Be careful to apply the correct stats for each val split (tandem_transfer uses tandem stats, others use single stats).

Run: `python train.py --agent tanjiro --wandb_name "tanjiro/split-phys-norm" --wandb_group split-phys-norm`

## Baseline
- val/loss: 2.1997

---

## Results

**W&B run:** zgxn7uvi  
**Epochs:** 66 (30 min wall-clock, ~27s/epoch)  
**Peak VRAM:** 10.6 GB  

| val_in_dist | 1.8592 | 0.283 | 0.179 | 20.4 | 1.305 | 0.457 | 24.9 |
| val_tandem_transfer | 3.2844 | 0.639 | 0.340 | 41.0 | 2.216 | 1.004 | 43.8 |
| val_ood_cond | 2.0117 | 0.268 | 0.192 | 22.0 | 1.088 | 0.422 | 20.9 |
| val_ood_re | 18869.7* | 0.277 | 0.201 | 31.1 | 1.086 | 0.453 | 51.8 |
| **combined (3-split)** | **2.3851** | | | | | | |

*Pre-existing branch bug (ood_re vol_loss overflow), excluded from 3-split.

**vs baseline:**
- val/loss: 2.3851 vs 2.1997 → **+8.4% (worse)**
- surf_p in_dist: 20.4 vs 20.03 → **+1.8% (minor)**
- surf_p tandem: 41.0 vs 40.41 → **+1.5% (minor)**
- surf_p ood_cond: 22.0 vs 20.57 → **+7.0% (worse)**

**What happened:**

Negative result. Split phys_stats degraded val/loss by 8.4%, though the physical surface MAE numbers are close to baseline.

The val/loss metric is computed in normalized space. With domain-split stats, tandem and single-foil samples have different normalization scales. The model must learn to produce outputs at different absolute magnitudes depending on whether it is processing a tandem sample — but the model has no explicit tandem/single pathway; it must implicitly infer this from input features (primarily the gap feature). This makes the optimization harder without a structural advantage.

The improved surface MAE (near baseline) suggests the model is learning the right physical outputs, but the val/loss metric (normalized errors) is worse because the denominator (phys_std) is smaller for each domain after splitting, making errors look larger relative to the tighter normalization.

Additionally, this approach introduces inconsistency: at test time (visualize/inference), the phys_stats applied depend on correctly identifying tandem vs single. If the gap feature is ambiguous near the threshold, the wrong stats could be applied.

**Suggested follow-ups:**
- Instead of splitting normalization, try a learned per-domain offset: add a small learnable bias to the output head conditioned on whether the sample is tandem/single. This would let the model learn the domain offset explicitly without changing the normalization scale.
- The tandem split has noticeably higher losses (3.28 vs 2.01 for single) — the real issue may be that tandem requires more model capacity, not normalization differences.

### DIFF
--- a/train.py
+++ b/train.py
@@ -438,26 +438,37 @@ val_loaders = {
     for name, subset in val_splits.items()
 }
 
-# --- Physics normalization stats (computed over training set) ---
-# Compute mean/std of Cp-normalized targets so the model sees O(1) values.
-print("Computing physics normalization stats...")
-_phys_sum = torch.zeros(3, device=device)
-_phys_sq_sum = torch.zeros(3, device=device)
-_phys_n = 0.0
+# --- Physics normalization stats (computed over training set, split by domain) ---
+# Tandem and single-foil have different pressure distributions; separate stats give
+# each domain tighter normalization and better-scaled gradients.
+print("Computing physics normalization stats (split single/tandem)...")
+_ps = {k: [torch.zeros(3, device=device), torch.zeros(3, device=device), 0.0]
+       for k in ("single", "tandem")}
 _stats_loader = DataLoader(train_ds, batch_size=cfg.batch_size, shuffle=False, **loader_kwargs)
 with torch.no_grad():
     for _x, _y, _is_surf, _mask in tqdm(_stats_loader, desc="Phys stats", leave=False):
-        _y, _mask = _y.to(device), _mask.to(device)
+        _x, _y, _mask = _x.to(device), _y.to(device), _mask.to(device)
+        _x_norm = (_x - stats["x_mean"]) / stats["x_std"]
+        _is_tan = _x_norm[:, 0, 21].abs() > 0.5  # [B]
         _Um, _q = _umag_q(_y, _mask)
         _yp = _phys_norm(_y, _Um, _q)
         _m = _mask.float().unsqueeze(-1)  # [B, N, 1]
-        _phys_sum += (_yp * _m).sum(dim=(0, 1))
-        _phys_sq_sum += (_yp ** 2 * _m).sum(dim=(0, 1))
-        _phys_n += _mask.float().sum().item()
-_pmean = (_phys_sum / _phys_n).float()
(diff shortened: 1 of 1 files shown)


tcapelle:
**Closing — split normalization makes optimization harder without structural benefit.**

val_loss=2.3851 (+8.4%), though physical in_p=20.4 is close to baseline. Your analysis is correct: separate normalization creates a harder optimization problem (model must implicitly infer domain from gap feature) without a structural shortcut. The tighter per-domain std makes normalized errors appear larger, inflating val_loss.

Key takeaway: normalization is a global property of the dataset, not per-domain. The model already handles the scale difference through its learned representations.

PR #957, opened 2026-03-18
FiLM conditioning on flow parameters (Re, AoA) for output modulation

tcapelle:
## Hypothesis
The model struggles on OOD conditions because the output head has no explicit conditioning on flow parameters. The same hidden representation must encode both geometry AND flow conditions. A lightweight FiLM-style (Feature-wise Linear Modulation) conditioning on global flow parameters (Re, AoA, gap) applied before the final block lets the model adapt its representations to the flow regime explicitly.

## Instructions
In `Transolver.__init__`, add after `self.re_head`:
```python
# FiLM conditioning on global flow parameters
self.cond_net = nn.Sequential(
    nn.Linear(4, 32),  # input: [log_Re, AoA0, AoA1, gap]
    nn.GELU(),
    nn.Linear(32, n_hidden * 2),  # output: scale and shift
)
nn.init.zeros_(self.cond_net[-1].weight)
nn.init.zeros_(self.cond_net[-1].bias)
self.cond_net[-1].bias.data[:n_hidden] = 1.0  # scale starts at 1, shift starts at 0
```

In `Transolver.forward`, before passing `fx` to the Transolver block:
```python
# Extract global conditioning: log_Re (idx 13), AoA0 (idx 14), AoA1 (idx 18), gap (idx 22)
cond = x[:, 0, [13, 14, 18, 22]]  # [B, 4] — same for all nodes
film_params = self.cond_net(cond)  # [B, n_hidden*2]
scale = film_params[:, :self.n_hidden].unsqueeze(1)  # [B, 1, n_hidden]
shift = film_params[:, self.n_hidden:].unsqueeze(1)  # [B, 1, n_hidden]
fx = fx * scale + shift  # modulate before final block
```

Zero-init ensures the model starts identical to baseline and gradually learns when conditioning helps.

Run: `python train.py --agent fern --wandb_name "fern/film-cond-output" --wandb_group film-conditioning`

## Baseline
- val/loss: 2.1997

---

## Results

**W&B run ID:** bk81xyqe  
**Best epoch:** 64 (~28.3s/epoch — slightly slower than usual 27s, 3 fewer epochs before timeout)  
**Peak memory:** ~65 GB (no OOM)  
**Parameter count:** 266,032 (+8,608 from cond_net vs 257,424 baseline)

### Surface MAE

| Split | surf_Ux | surf_Uy | surf_p | surf_p baseline | Δ |
|-------|---------|---------|--------|-----------------|---|
| val_in_dist | 0.307 | 0.175 | **21.78** | 20.03 | **+1.75** ✗ |
| val_ood_cond | 0.270 | 0.195 | **23.80** | 20.57 | **+3.23** ✗✗ |
| val_ood_re | 0.271 | 0.198 | **31.32** | — | — |
| val_tandem_transfer | 0.637 | 0.340 | **42.76** | 40.41 | **+2.35** ✗ |

### Volume MAE

| Split | vol_Ux | vol_Uy | vol_p |
|-------|--------|--------|-------|
| val_in_dist | — | — | 25.83 |
| val_ood_cond | — | — | 21.79 |
| val_ood_re | — | — | 51.16 |
| val_tandem_transfer | — | — | 44.27 |

### What happened

**FiLM conditioning significantly hurts performance across all splits.** The val/loss regressed by +0.11 (~5%), and surface pressure MAE got substantially worse everywhere — especially for OOD conditions (+3.23 for ood_cond), which is exactly the split this hypothesis was supposed to help.

The zero-initialization of cond_net's last layer ensures scale=1, shift=0 at init, so the model starts identical to baseline. However, the training dynamics change: the scale/shift parameters now carry gradients that modulate the ENTIRE hidden representation before the final block. Even with zero-init, the optimizer can quickly push these away from identity, potentially disrupting the learned representation early in training.

Some possible root causes:
1. **Wrong feature indices**: The conditioning uses normalized x[:, 0, [13, 14, 18, 22]]. In the normalized feature space, these may not cleanly correspond to log_Re, AoA0, AoA1, gap — nearby features in the 24-dim input may bleed in.
2. **FiLM placement**: Conditioning the final-block INPUT means the last attention layer sees a fully transformed representation. The scale/shift could destroy the structured hidden state that prior layers carefully learned.
3. **Under-convergence**: 64 epochs is not enough for the conditioning to learn useful patterns while simultaneously not degrading the base representation. The model might need 2-3× more training to benefit.

**Verdict:** FiLM conditioning at this location and training budget doesn't work. Negative result, clearly.

### Suggested follow-ups
1. **FiLM earlier in pipeline**: apply conditioning after preprocessing (before the attention blocks), where the model is less sensitive to disruption.
2. **Conditioning on the preprocess MLP output**: add a learnable bias from flow params to the initial embedding, not the final block input.
3. **AdaLN-Zero instead of FiLM**: apply conditioning as an adaptive LayerNorm scale/shift within the TransolverBlock, similar to DiT (Diffusion Transformer) approach. This is more controlled.

### DIFF
--- a/train.py
+++ b/train.py
@@ -269,6 +269,15 @@ class Transolver(nn.Module):
         self.placeholder_scale = nn.Parameter(torch.ones(n_hidden))
         self.placeholder_shift = nn.Parameter(torch.zeros(n_hidden))
         self.re_head = nn.Sequential(nn.Linear(n_hidden, 32), nn.GELU(), nn.Linear(32, 1))
+        # FiLM conditioning on global flow parameters
+        self.cond_net = nn.Sequential(
+            nn.Linear(4, 32),  # input: [log_Re, AoA0, AoA1, gap]
+            nn.GELU(),
+            nn.Linear(32, n_hidden * 2),  # output: scale and shift
+        )
+        nn.init.zeros_(self.cond_net[-1].weight)
+        nn.init.zeros_(self.cond_net[-1].bias)
+        self.cond_net[-1].bias.data[:n_hidden] = 1.0  # scale starts at 1, shift starts at 0
 
     def initialize_weights(self):
         self.apply(self._init_weights)
@@ -336,6 +345,13 @@ class Transolver(nn.Module):
 
         # Auxiliary Re prediction from pre-output-head hidden representation
         re_pred = self.re_head(fx.mean(dim=1))  # [B, 1]
+
+        # FiLM conditioning on global flow parameters before final block
+        cond = x[:, 0, [13, 14, 18, 22]]  # [B, 4] — log_Re, AoA0, AoA1, gap (same for all nodes)
+        film_params = self.cond_net(cond)  # [B, n_hidden*2]
+        scale = film_params[:, :self.n_hidden].unsqueeze(1)  # [B, 1, n_hidden]
+        shift = film_params[:, self.n_hidden:].unsqueeze(1)  # [B, 1, n_hidden]
+        fx = fx * scale + shift  # modulate before final block
 
         fx = self.blocks[-1](fx, raw_xy=raw_xy)
         fx = fx + self.out_skip(fx_pre)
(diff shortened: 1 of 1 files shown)


tcapelle:
**Closing — FiLM conditioning doesn't help at this scale.**

val_loss=2.3113 (+5.1%), ood_p=23.8 (+15.7%). The model already extracts flow condition info from input features; explicit conditioning via FiLM adds parameters without providing new information. The zero-init means it had to learn when conditioning helps, but apparently it doesn't for this architecture.

PR #1054, opened 2026-03-18
Coordinate-normalized Fourier PE (normalize x,y to [0,1] before encoding)

tcapelle:
## Hypothesis
The merged PR's suggested follow-up: "Try normalizing the (x,y) coordinates before computing Fourier features — the coordinates span different scales depending on mesh extent." Currently, standardized x,y have varying ranges. Normalizing to [0,1] per-sample before sin/cos ensures consistent frequency interpretation. Multiplying by pi gives full sin/cos periods.

## Instructions
In both **training** (lines 595-599) and **validation** (lines 734-738), replace:
```python
raw_xy = x[:, :, :2]
freqs = 2.0 ** torch.arange(4, device=x.device, dtype=x.dtype)
xy_scaled = raw_xy.unsqueeze(-1) * freqs
```
With:
```python
raw_xy = x[:, :, :2]
# Normalize xy to [0,1] per-sample for consistent Fourier encoding
xy_min = raw_xy.amin(dim=1, keepdim=True)
xy_max = raw_xy.amax(dim=1, keepdim=True)
xy_norm = (raw_xy - xy_min) / (xy_max - xy_min + 1e-8)
freqs = 2.0 ** torch.arange(4, device=x.device, dtype=x.dtype) * 3.14159
xy_scaled = xy_norm.unsqueeze(-1) * freqs
```

The rest stays the same (sin/cos/cat). Run with `--wandb_group coord-norm-fourier`.

## Baseline (after Fourier PE merge)
- best_val_loss = 2.2117
- val_in_dist/mae_surf_p = 19.72
- val_ood_cond/mae_surf_p = 20.35
- val_ood_re/mae_surf_p = 30.37
- val_tandem_transfer/mae_surf_p = 40.92

---

### v3 — rebased onto noam (n_hidden=192 + learnable Fourier freqs)

**W&B run:** `tvvjs1if`  
**Epochs completed:** 72 (hit 30-min timeout; n_hidden=192 ~24s/epoch vs 20s for 160)  
**Peak memory:** 12.4 GB  

#### Validation losses (3-split, lower is better)

| Split | val/loss |
|-------|----------|
| val_in_dist | 1.4674 |
| val_ood_cond | 1.5219 |
| val_tandem_transfer | 3.0874 |
| **val/loss (3-split)** | **2.0256** |

Old Fourier-PE-only baseline: 2.2117 → **improvement: −8.5%**

#### Surface MAE — pressure (most important)

[...]

PR #1055, opened 2026-03-18
Distance-to-surface input feature for volume nodes

tcapelle:
## Hypothesis
Curvature proxy (#911) was a successful feature for surface nodes. Volume nodes near the surface carry the boundary layer — the most physically important region. A wall-distance feature (dsdf norm) for volume nodes gives them explicit proximity information. This is the complement to curvature: curvature for surface, distance for volume. Unlike loss weighting (which failed in #1037), this is an INPUT FEATURE that the model can learn to use.

## Instructions
After curvature computation (line 593 in training, line 732 in validation), add:
```python
# Wall-distance proxy: dsdf norm for volume nodes (complement surface curvature)
wall_dist = x[:, :, 2:6].norm(dim=-1, keepdim=True) * (~is_surface.bool()).float().unsqueeze(-1)
x = torch.cat([x, wall_dist], dim=-1)
```

Update **fun_dim** (line 464):
```python
fun_dim=X_DIM - 2 + 1 + 1 + 16,  # +1 curvature, +1 wall_dist, +16 Fourier PE
```

Run with `--wandb_group wall-dist-feature`.

- best_val_loss = 2.2117

---

## Results

**W&B run:** 3aq7u2od  
**Epochs completed:** 63 (timeout)  
**Peak memory:** 10.9 GB (vs 10.6 GB baseline, +0.3 GB for extra feature)

| Metric | wall_dist | Baseline | Delta |
|---|---|---|---|
| best_val_loss (3split) | 2.2680 | 2.2117 | +0.0563 |
| val_in_dist/mae_surf_p | 20.30 | 19.72 | +0.58 |
| val_ood_cond/mae_surf_p | 20.44 | 20.35 | +0.09 |
| val_ood_re/mae_surf_p | 31.00 | 30.37 | +0.63 |
| val_tandem_transfer/mae_surf_p | 41.91 | 40.92 | +0.99 |

**What happened:** The wall-distance feature did not improve over baseline — all metrics are slightly worse. The dsdf channels (indices 2:6) in x already encode distance-to-surface information, so the explicit norm of these channels for volume nodes is largely redundant. Adding a duplicate representation of already-available information appears to confuse the model's first layer rather than helping it. The curvature proxy works because it provides *new* compressed information (surface shape) not trivially derivable from a single scalar. Wall-distance adds no new information beyond what's already in channels 2:6.

The +0.3 GB memory increase is negligible.

**Suggested follow-ups:**
- If boundary-layer features are desired, a more informative feature might be a signed distance (preserving direction) or the actual distance in physical units rather than a norm of normalized dsdf.
- Alternatively, skip the feature engineering direction and look at architectural changes for the volume nodes (e.g. different attention patterns near the surface).

### DIFF
--- a/train.py
+++ b/train.py
@@ -461,7 +461,7 @@ print(f"  Cp stats — mean: {_pmean.tolist()}, std: {_pstd.tolist()}")
 
 model_config = dict(
     space_dim=2,
-    fun_dim=X_DIM - 2 + 1 + 16,  # X_DIM=24 + 1 curvature proxy + 16 Fourier PE; fun_dim + space_dim must equal x.shape[-1]
+    fun_dim=X_DIM - 2 + 1 + 1 + 16,  # +1 curvature, +1 wall_dist, +16 Fourier PE
     out_dim=3,
     n_hidden=128,
     n_layers=1,       # was 2 — 1 layer for maximum epochs in 30 min
@@ -591,6 +591,9 @@ for epoch in range(MAX_EPOCHS):
         # Curvature proxy: norm of first 4 dsdf channels (gradient magnitude) for surface nodes
         curv = x[:, :, 2:6].norm(dim=-1, keepdim=True) * is_surface.float().unsqueeze(-1)
         x = torch.cat([x, curv], dim=-1)
+        # Wall-distance proxy: dsdf norm for volume nodes (complement surface curvature)
+        wall_dist = x[:, :, 2:6].norm(dim=-1, keepdim=True) * (~is_surface.bool()).float().unsqueeze(-1)
+        x = torch.cat([x, wall_dist], dim=-1)
         # Fourier positional encoding: append sin/cos of (x,y) at 4 frequencies
         raw_xy = x[:, :, :2]
         freqs = 2.0 ** torch.arange(4, device=x.device, dtype=x.dtype)
@@ -730,6 +733,9 @@ for epoch in range(MAX_EPOCHS):
                 # Curvature proxy: norm of first 4 dsdf channels (gradient magnitude) for surface nodes
                 curv = x[:, :, 2:6].norm(dim=-1, keepdim=True) * is_surface.float().unsqueeze(-1)
                 x = torch.cat([x, curv], dim=-1)
+                # Wall-distance proxy: dsdf norm for volume nodes (complement surface curvature)
+                wall_dist = x[:, :, 2:6].norm(dim=-1, keepdim=True) * (~is_surface.bool()).float().unsqueeze(-1)
+                x = torch.cat([x, wall_dist], dim=-1)
                 # Fourier positional encoding: append sin/cos of (x,y) at 4 frequencies
                 raw_xy = x[:, :, :2]
                 freqs = 2.0 ** torch.arange(4, device=x.device, dtype=x.dtype)


morganmcg1:
ADVISOR: Checking in — have you started all three parallel arms from PR #1055?

This is the SDF α-sweep: Arm A (α=1.5), Arm B (α=3.0), Arm C (α=4.0). All three arms should be running simultaneously on your 8 GPUs (2-3 GPUs per arm via DDP).

Key reminders:
1. Corrected dataset: `--data-root /mnt/pvc/Processed/drivaerml_processed_rawcanon_20260511`
2. Kill gates apply independently per arm — kill any arm that misses its gate immediately, do NOT wait for other arms
3. W&B group: `--wandb-group sdf-alpha-sweep`
4. Run names: `tanjiro-sdf-alpha1.5`, `tanjiro-sdf-alpha3.0`, `tanjiro-sdf-alpha4.0`

Please confirm all three arms are running and post the W&B run IDs for each arm here.

morganmcg1:
ADVISOR: Tanjiro — critical issues require killing all three arms and restarting. Three things to address:

## 1. KILL ALL THREE ARMS NOW — monkey-patch is a silent no-op

The same `types.MethodType` bug that nezuko independently found (see PR #1054 comment from 22:07Z) applies to all three of your arms. Independently, nezuko confirmed:

- `train_dataset.__getitem__ = types.MethodType(func, train_dataset)` installs on the *instance dict*
- Python resolves dunder methods on the **type**, not the instance — so `dataset[idx]` subscript bypasses the override entirely
- PyTorch DataLoader uses subscript (`_MapDatasetFetcher.fetch`), so the patched method is **never called** during training
- All three arms are currently running uniform `train_random` sampling — the SDF weighting has zero effect

Kill `q2vet888`, `92nd8zrn`, `z8a6ckv6` immediately. The correct fix is to reassign `train_dataset.__class__` to a dynamically-built subclass:

```python
class _SDFStratifiedDataset(type(train_dataset)):
    def __getitem__(self, idx):
        view = self.views[idx]
        counts = self.store.case_point_counts(view.case_id)
        surface_idx = self._indices(
            counts["n_surface"], self.max_surface_points, view,
            group_view_count=view.surface_view_count,
        )
        case = self.store.load_case(
            view.case_id,
            surface_rows=None if surface_idx is None else surface_idx.numpy(),
            volume_rows=None,  # load ALL volume points first for weights
        )
        n_total = case.volume_x.shape[0]
        n_sample = min(self._sdf_n_vol, n_total)
        sdf_abs = case.volume_x[:, 3].abs()   # channel 3 = SDF
        weights = 1.0 / (1.0 + self._sdf_alpha * sdf_abs)   # near-surface emphasis (see #2)
        vol_idx = torch.multinomial(weights, n_sample, replacement=True)
        vol_idx = vol_idx.sort().values
        from data.loader import DrivAerMLCase
        return DrivAerMLCase(
            case_id=case.case_id,
            surface_x=case.surface_x, surface_y=case.surface_y,
            volume_x=case.volume_x[vol_idx], volume_y=case.volume_y[vol_idx],
            metadata={},
        )

# Set per-arm attributes, then reassign __class__:
train_dataset._sdf_alpha = args.sdf_stratified_alpha
train_dataset._sdf_n_vol = args.train_volume_points
train_dataset.__class__ = _SDFStratifiedDataset
```

Verify by confirming consecutive draws on the same view idx give different `volume_x` rows.

If `store.load_sdf` is available separately, use nezuko's IO optimisation: load only `volume_sdf.npy` to build weights, then call `store.load_case(volume_rows=vol_idx.numpy())` — this reduces IO ~5× (~176ms/call vs ~880ms).

## 2. Formula direction — you are correct, invert it

Your diagnosis is exactly right. `weight = 1 + α·|sdf|` upweights far-field points (|sdf| ≈ 80m → weight ≈ 321 for α=4.0), which is the **opposite** of near-surface emphasis. Use the inverse:

```python
weights = 1.0 / (1.0 + alpha * sdf_abs)
```

With this formula: points at sdf≈0 get weight≈1.0; points at sdf≈80m get weight→0. This is true near-surface emphasis. Keep α values (1.5, 3.0, 4.0) — higher α now means *stronger* near-surface concentration, which is the correct sweep interpretation.

The original formula was wrong in the PR body. The three arms you killed were also running uniform sampling (monkey-patch no-op), so nothing is lost.

## 3. Sequential 8-GPU runs — yes, switch

Switch to sequential 8-GPU runs. Reasons:
- EBS confound (3/3/2 vs baseline EBS=8) adds an uncontrolled variable that makes it impossible to attribute differences to α vs. batch dynamics
- Runs need to be killed and restarted anyway — no sunk cost
- 17h/arm fits within the 24h wall-clock budget
- With corrected formula + corrected monkey-patch, this is a genuinely new experiment — clean results matter

Run the three arms sequentially (A then B then C). Nezuko's PR #1054 is running the combined SDF+stochastic at α=2.0 simultaneously; your three arms are the pure-SDF sweep (no explicit stochastic freshness beyond the DDP rank-seed offset).

Keep same kill gates: EP1≤30%, EP2≤16%, EP3≤8%, EP5≤7.5%, EP10≤7.2%, EP15≤6.80%, EP20≤6.70%, EP25≤6.65%, EP30≤6.60%. If all three arms fail EP3, stop and report immediately — do not continue to EP30.

Great catches on PE kind, model-size flags, mount path, and the SDF range analysis.

— ADVISOR

morganmcg1:
ADVISOR: **URGENT — EP2 kill-threshold crisis for tanjiro run 58hk6r36**

I have posted two corrections already (00:02:36Z and 00:19:52Z) but have not heard back. I need to confirm what happened at EP2 (step 21950).

Your run 58hk6r36 was launched with <= thresholds (wrong). Specifically:
- 21950:val_primary/abupt_axis_mean_rel_l2_pct<=16 — this KILLS the run if val_abupt drops below 16% (i.e., kills good progress)

If the run survived EP2, it means val_abupt was still >=16% at step 21950 — which would be a miss, not a pass.
If the run was killed at EP2, it means val_abupt had improved below 16% — which is actually good, but the <= threshold killed it.

**Please immediately:**
1. Tell me the current state of run 58hk6r36 — is it still running or was it killed?
2. If killed at EP2: what was the val_abupt at step 21950?
3. If still running: kill it now.

Then relaunch Arm A with corrected >= thresholds:

```bash
cd target/ && torchrun --standalone --nproc_per_node=8 train.py   --agent dl24-tanjiro   --data-root /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511   --optimizer lion --lr 1e-4 --weight-decay 5e-3   --batch-size 1 --epochs 30   --lr-warmup-epochs 1 --lr-cosine-t-max 30 --lr-min 1e-6   --amp-mode bf16   --ema-decay 0.999 --ema-start-step 50   --grad-clip-norm 1.0   --model-layers 6 --model-hidden-dim 512 --model-heads 4 --model-slices 128   --pos-encoding-mode string_separable --use-qk-norm   --rff-num-features 16 --rff-init-sigmas '0.25,0.5,1.0,2.0,4.0'   --train-surface-points 65000 --eval-surface-points 65000   --train-volume-points 65000 --eval-volume-points 65000   --validation-every 1 --no-compile-model   --kill-thresholds '10975:val_primary/abupt_axis_mean_rel_l2_pct>=30,21950:val_primary/abupt_axis_mean_rel_l2_pct>=16,32925:val_primary/abupt_axis_mean_rel_l2_pct>=8,54875:val_primary/abupt_axis_mean_rel_l2_pct>=7.5,109750:val_primary/abupt_axis_mean_rel_l2_pct>=7.2,164625:val_primary/abupt_axis_mean_rel_l2_pct>=6.80,219500:val_primary/abupt_axis_mean_rel_l2_pct>=6.70,274375:val_primary/abupt_axis_mean_rel_l2_pct>=6.65,329250:val_primary/abupt_axis_mean_rel_l2_pct>=6.60'   --sdf-alpha 1.5   --wandb-group sdf-alpha-sweep-inv --wandb-name sdf-arm-a-alpha1.5-corrected
```

**Remember**: >= means 'kill if still too HIGH'. The EP1 result for 58hk6r36 was val_abupt=22.97% which passed the <=30 EP1 gate by chance (since 22.97 is not <=30... wait, that is <=30). 

Actually, looking more carefully: your EP1 at step 10975 showed val_abupt=22.9680%. With <= thresholds:
- EP1: 22.97 <= 30 is TRUE → run gets killed at EP1 too!

So the run may have already been killed at EP1 itself. Please confirm immediately.

Also note: EP1 result 22.97% is a PASS (below the 30% bar). The corrected >= run should be relaunched immediately.

morganmcg1:
ADVISOR CORRECTION — CRITICAL: I made an error in my previous comment. Please **IGNORE** the claim that your run was killed at EP1, and **IGNORE** the instruction to relaunch with `>=` operator.

**The `<=` operator is CORRECT.** I just verified the kill-threshold implementation in trainer_runtime.py. The kill fires when `passes()` returns FALSE:

```python
# STEP:METRIC<=30 means: continue if metric <= 30%, KILL if metric > 30%
if not threshold.passes(observed):  # kills when condition is NOT met
    return 'kill threshold failed ...'
```

This is exactly correct gate semantics. Your run 58hk6r36 with `<=` thresholds is **correctly configured** and W&B confirms it is **running** with:
- EP1 (step 10975): val_abupt = **22.97%** ✓ (below 30% gate — run continued correctly)
- EP2 (step 21951): val_abupt = **7.62%** ✓ (well below 16% gate — exceptional result!)

**Action:** Do NOT relaunch. Run 58hk6r36 is alive, correctly configured, and producing extraordinary results. Let Arm A (α=1.5) complete naturally.

EP3 gate: val_abupt ≤ 8% at step ~32922. Based on EP2=7.62%, this should pass easily.

After Arm A completes, proceed to Arm B (α=3.0) then Arm C (α=4.0) as originally planned.

Apologies for the confusion in my previous comment.

PR #1057, opened 2026-03-18
Surface-only Fourier PE (zero for volume nodes)

tcapelle:
## Hypothesis
The Fourier PE success and curvature proxy success share a pattern: surface feature enrichment works. Applying Fourier PE only to surface nodes concentrates the model's feature capacity on the ~5% of nodes that matter most. This mirrors how curvature proxy is surface-only and worked, while extending curvature to all nodes (#1032) failed.

## Instructions
In both **training** (line 598) and **validation** (line 737), after computing fourier_pe, add surface masking:
```python
fourier_pe = torch.cat([xy_scaled.sin().flatten(-2), xy_scaled.cos().flatten(-2)], dim=-1)
fourier_pe = fourier_pe * is_surface.float().unsqueeze(-1)  # surface-only
x = torch.cat([x, fourier_pe], dim=-1)
```

No fun_dim change needed. Run with `--wandb_group surf-only-fourier`.

- best_val_loss = 2.2117

---

## Results

**W&B run:** `c30n4xyk` | **Epochs:** 66/100 (30 min timeout) | **Peak memory:** 10.9 GB

### Surface Pressure MAE (most important)

| Split | Baseline | This run | Delta |
|---|---|---|---|
| val_in_dist | 19.72 | 20.89 | +1.17 ❌ |
| val_ood_cond | 20.35 | 21.71 | +1.36 ❌ |
| val_ood_re | 30.37 | 31.29 | +0.92 ❌ |
| val_tandem_transfer | 40.92 | 42.15 | +1.23 ❌ |

### val/loss (3-split average, excl. diverged val_ood_re)

| | Baseline | This run | Delta |
|---|---|---|---|
| best_val_loss | 2.2117 | **2.2989** | +0.087 ❌ |

### Volume MAE (mae_vol_p)

| Split | This run |
|---|---|
| val_in_dist | 28.30 |
| val_ood_cond | 20.85 |
| val_ood_re | 52.22 |
| val_tandem_transfer | 45.82 |

Note: Visualization crashed post-training with a shape mismatch in `data/utils.py` (pre-existing issue — same as PR #1039, caused by the baseline Fourier PE changing input_dim and the read-only viz code not accounting for it). Training metrics are valid.

### What happened

**Clear negative result.** Surface-only Fourier PE made every metric worse by ~1-1.4 units of surface pressure MAE.

The hypothesis was wrong: the analogy between curvature proxy (surface-only) and Fourier PE breaks down because they serve fundamentally different purposes. Curvature is semantically meaningful only on surfaces — volume nodes have no curvature, so giving them a zero is correct. Fourier PE, on the other hand, provides positional identity for every node in the flow field. Zeroing PE on 95% of nodes (all volume nodes) strips those nodes of their spatial context, degrading the model's ability to learn the flow field geometry.

The model sees volume nodes with no positional encoding and can only distinguish them via their raw (x,y) coords and physics features — this is a strictly worse representation. The degraded volume accuracy then propagates to surface accuracy through the joint optimization.

### Suggested follow-ups
1. **Additive surface PE**: Keep full PE on all nodes, but *add* an extra learned surface-only feature on top (e.g., curvature, normals, or a learned surface embedding) rather than replacing volume PE with zeros.
2. **Surface PE amplification**: Keep volume PE but *scale up* surface PE (e.g., multiply surface PE by 2×) to give surface nodes more positional weight without crippling volume nodes.
3. **Learned surface token**: Replace the manual PE-zeroing with a learnable binary surface flag embedded into the feature space, letting the model learn how to use surface vs. volume spatial information differently.

### DIFF
--- a/train.py
+++ b/train.py
@@ -596,6 +596,7 @@ for epoch in range(MAX_EPOCHS):
         freqs = 2.0 ** torch.arange(4, device=x.device, dtype=x.dtype)
         xy_scaled = raw_xy.unsqueeze(-1) * freqs  # [B, N, 2, 4]
         fourier_pe = torch.cat([xy_scaled.sin().flatten(-2), xy_scaled.cos().flatten(-2)], dim=-1)  # [B, N, 16]
+        fourier_pe = fourier_pe * is_surface.float().unsqueeze(-1)  # surface-only
         x = torch.cat([x, fourier_pe], dim=-1)
         Umag, q = _umag_q(y, mask)
         y_phys = _phys_norm(y, Umag, q)
@@ -735,6 +736,7 @@ for epoch in range(MAX_EPOCHS):
                 freqs = 2.0 ** torch.arange(4, device=x.device, dtype=x.dtype)
                 xy_scaled = raw_xy.unsqueeze(-1) * freqs  # [B, N, 2, 4]
                 fourier_pe = torch.cat([xy_scaled.sin().flatten(-2), xy_scaled.cos().flatten(-2)], dim=-1)  # [B, N, 16]
+                fourier_pe = fourier_pe * is_surface.float().unsqueeze(-1)  # surface-only
                 x = torch.cat([x, fourier_pe], dim=-1)
                 Umag, q = _umag_q(y, mask)
                 y_phys = _phys_norm(y, Umag, q)


morganmcg1:
ADVISOR: You've diagnosed this perfectly, fern. The command I posted used flags from the \`dl24-frieren/vol-test-domain-augmentation\` branch, not from tay — my error. **Use Option 1** — the PR #958 reproduce command with \`--vol-p-log-space-loss\` added. Here is the corrected, tay-compatible command:

\`\`\`bash
cd target/ && torchrun --standalone --nproc_per_node=8 train.py \
  --agent fern \
  --data-root /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511 \
  --optimizer lion --lr 9e-5 --weight-decay 5e-4 \
  --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0 \
  --use-ema --ema-decay 0.999 \
  --grad-clip-norm 0.5 --lr-warmup-epochs 1 \
  --pos-encoding-mode string_separable --use-qk-norm \
  --rff-num-features 16 --rff-init-sigmas "0.25,0.5,1.0,2.0,4.0" \
  --lr-cosine-t-max 30 --lr-min 1e-6 --epochs 30 \
  --vol-points-schedule "0:16384:3:32768:6:49152:9:65536" \
  --no-compile-model \
  --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --batch-size 4 --validation-every 1 \
  --train-surface-points 65536 --eval-surface-points 65536 \
  --train-volume-points 65536 --eval-volume-points 65536 \
  --use-surf-to-vol-xattn \
  --use-vol-pressure-aux-head --volume-loss-weight 1.0 \
  --vol-p-log-space-loss \
  --kill-thresholds "1:30,2:16,3:9.5,5:8.0,10:7.5,15:7.0,20:6.80,25:6.70,30:6.60" \
  --wandb-group fern-logspace-vol-p-corrected \
  --wandb-name fern-logspace-vol-p-corrected-v2
\`\`\`

**Changes from the broken command:**
- Removed all invalid flags: \`--model-pe\`, \`--pe-init-sigmas\`, \`--use-y-symmetry-aug\`, \`--y-symmetry-aug-prob\`, \`--seed\`
- Replaced with valid tay equivalents: \`--pos-encoding-mode string_separable\`, \`--rff-init-sigmas "0.25,0.5,1.0,2.0,4.0"\`
- Restored SOTA stack: \`--model-layers 5 --batch-size 4\` (not layers 6 / batch 1)
- Added \`--use-ema --ema-decay 0.999\` (present in PR #958 baseline)
- Epochs extended to 30 for a full confirmation run

**On the gates:** You're right to flag this. The \`val_abupt < 6.126%\` gate referenced the DL24-branch SOTA, which isn't reproducible on tay. The correct tay-branch single-model SOTA is **PR #958: val_abupt=6.285%, val_vol_p=3.818%**. The hypothesis here is that log-space loss further reduces \`val_vol_p\` below 3.818% while keeping \`val_abupt\` below 6.285%. Kill thresholds above are calibrated to the tay stack budget.

**What we're testing:** This is a clean A/B — PR #958 reproduce command (no log-space loss) vs. same command + \`--vol-p-log-space-loss\`. The single variable is the loss transform.

Launch when ready and post EP3 metrics as usual.

morganmcg1:
ADVISOR: Nudge — the corrected tay-compatible command was posted at 23:32Z. Please launch it now and post the W&B run ID + EP1 metrics as soon as EP1 completes. The hypothesis is valid and I want to see it running. Label moved back to status:wip.

morganmcg1:
ADVISOR: **URGENT — Kill gqgz3y36 immediately and relaunch as v4**

Your v3 run gqgz3y36 was launched with the WRONG <= operator from my erroneous 00:06:37Z instruction. I posted a correction at 00:19:27Z — you missed it by 22 seconds (your v3 launched at 00:19:05Z, not your fault).

The problem: 10864:val_primary/abupt_axis_mean_rel_l2_pct<=30 will kill the run the moment val_abupt drops below 30% — i.e., kills every healthy run almost immediately. The EP1 gate fires at step 10864. You must kill and relaunch before then.

**The correct operator is >=** — kill only if the metric is still too high (hasn't improved enough).

**v4 launch command (corrected >= operator, correct 10864-based steps, NO --use-vol-pressure-aux-head which causes crash on your branch):**

```bash
cd target/ && torchrun --standalone --nproc_per_node=8 train.py   --agent fern --optimizer lion --lr 9e-5 --weight-decay 5e-4   --tau-y-loss-weight 1.5 --tau-z-loss-weight 2.0 --surface-loss-weight 2.0   --ema-decay 0.999 --grad-clip-norm 0.5 --lr-warmup-epochs 1   --pos-encoding-mode string_separable --use-qk-norm   --rff-num-features 16 --rff-init-sigmas '0.25,0.5,1.0,2.0,4.0'   --lr-cosine-t-max 13 --epochs 13   --vol-points-schedule '0:16384:3:32768:6:49152:9:65536'   --no-compile-model   --model-layers 5 --model-hidden-dim 512 --model-heads 4 --model-slices 128   --batch-size 4 --validation-every 1   --train-surface-points 65536 --eval-surface-points 65536   --train-volume-points 65536 --eval-volume-points 65536   --use-surf-to-vol-xattn   --data-root /mnt/new-pvc/Processed/drivaerml_processed_rawcanon_20260511   --vol-p-log-space-loss   --kill-thresholds '10864:val_primary/abupt_axis_mean_rel_l2_pct>=30,21728:val_primary/abupt_axis_mean_rel_l2_pct>=16,32592:val_primary/abupt_axis_mean_rel_l2_pct>=8,54320:val_primary/abupt_axis_mean_rel_l2_pct>=7.5,108640:val_primary/abupt_axis_mean_rel_l2_pct>=7.2,141232:val_primary/abupt_axis_mean_rel_l2_pct>=7.0'   --wandb-group fern-logspace-vol-p-corrected --wandb-name fern-logspace-vol-p-corrected-v4
```

Key corrections vs v3:
- >= throughout (kills if metric is still high, not if it improved)
- No --use-vol-pressure-aux-head (causes crash on your branch)
- Step counts: 10864-based (correct for 347657-view dataset with EBS=32)

Please kill gqgz3y36 NOW and relaunch with the v4 command above. Post new W&B run ID as soon as you start.

morganmcg1:
ADVISOR CORRECTION — CRITICAL: I made an error in my previous comment. Please **IGNORE** the instruction to kill gqgz3y36 and relaunch with `>=` operator.

**The `<=` operator is CORRECT.** I just verified the kill-threshold implementation in trainer_runtime.py:

```python
def passes(self, observed: float) -> bool:
    if self.operator == '<=':
        return observed <= self.value
    ...

# A run is killed when passes() returns FALSE:
if not threshold.passes(observed):
    return 'kill threshold failed ...'
```

So `STEP:METRIC<=30` means: **continue if metric ≤ 30%, kill if metric > 30%**. This is exactly the correct gate semantics — it kills runs that are still too high.

The `>=` operator I instructed you to use would have killed runs doing WELL (low error) and kept runs doing poorly — the exact opposite of what we want.

**Current status of your v3 run gqgz3y36:**
- If it is still running with `<=` thresholds, it is correctly configured. Do NOT kill it.
- Check its current step and val_abupt. If it passed EP1 (step 10864, val_abupt ≤ 30%), it is on track.

**Action:** Please post the current W&B run ID, current step, and EP1 result (val_abupt at step ~10864) as soon as possible. If gqgz3y36 is alive and past EP1, let it continue — do not relaunch.

Apologies for the confusion. The step counts in your v3 command (10864-based for 347657-view dataset with EBS=32) are correct.

PR #1063, opened 2026-03-18
n_hidden=256 mlp_ratio=1 (wide attention, lean FFN)

tcapelle:
## Hypothesis
Test whether the bottleneck is attention expressiveness vs FFN capacity. n_hidden=256 with mlp_ratio=1 gives 64-dim attention heads (same as original Transolver paper) while keeping total params similar to n_hidden=192 with mlp_ratio=2. The SE block also benefits from 256-dim channels.

## Instructions
Two changes in model_config (lines 468, 472):
```python
n_hidden=256,  # was 160
mlp_ratio=1,   # was 2 — lean FFN to compensate for wider attention
```

Monitor epoch time — aim for >70 epochs in 30 min. If too slow, try n_hidden=224.

Run with `--wandb_group wide-attn-256-mlp1`.

## Baseline (n_hidden=160 + compile + Fourier PE)
- best_val_loss = 2.0996
- val_in_dist/mae_surf_p = 18.58
- val_ood_cond/mae_surf_p = 18.53
- val_ood_re/mae_surf_p = 29.55
- val_tandem_transfer/mae_surf_p = 41.63
- mean3_surf_p = 26.25

---

## Results

**W&B run:** `wd2i8ocr` | **Epochs:** 64/100 (30 min timeout) | **Peak memory:** 14.9 GB

### Speed note
Epoch 1 took 89s (compile warmup), steady state ~28s/epoch → 64 epochs total. This is below the 70-epoch target. I opted not to switch to n_hidden=224 because (a) restarting would cost ~90s compile overhead giving fewer total training seconds, and (b) at epoch 64 the model was still converging (loss improving every epoch), suggesting the result is informative. The comparison is slightly unfair: n_hidden=256 is ~6 epochs behind baseline, so any quality gap is an underestimate of the true difference.

### Surface Pressure MAE (most important)

| Split | Baseline | This run | Delta |
|---|---|---|---|
| val_in_dist | 18.58 | 18.88 | +0.30 ❌ |
| val_ood_cond | 18.53 | 19.70 | +1.17 ❌ |
| val_ood_re | 29.55 | 29.86 | +0.31 ❌ |
| val_tandem_transfer | 41.63 | 41.63 | 0.00 → |

### val/loss (3-split average)

| | Baseline | This run | Delta |
|---|---|---|---|
| best_val_loss | 2.0996 | **2.1651** | +0.065 ❌ |

### Volume MAE (mae_vol_p)

| Split | This run |
|---|---|
| val_in_dist | 23.28 |
| val_ood_cond | 18.08 |
| val_ood_re | 49.97 |
| val_tandem_transfer | 42.45 |

Memory: 14.9 GB (vs ~10.9 GB baseline, +4 GB for 37% more channels)

Note: Visualization crashed post-training (pre-existing issue in `data/utils.py`). Training metrics are valid.

### What happened

**Slightly negative, but ambiguous.** n_hidden=256 + mlp_ratio=1 is marginally worse than n_hidden=160 + mlp_ratio=2 on all 3 in-distribution/OOD-cond splits, with the tandem_transfer split being equal.

However, the comparison is confounded by training time: n_hidden=256 ran 64 epochs vs the baseline's ~70+ due to slower per-epoch speed. Since the model was still converging at epoch 64, the true peak performance with more training could close the gap. It is not possible to conclude from this alone whether wider attention genuinely underperforms, or if it just needs more epochs.

Memory cost (+4 GB, 37% more) is substantial for what appears to be no gain. Even if training to convergence, the parameter budget is used less efficiently: going from mlp_ratio=2 to mlp_ratio=1 means the FFN (a key capacity component) is halved to pay for wider attention.

### Suggested follow-ups
1. **n_hidden=224, mlp_ratio=1**: The speed-optimal intermediate — wider than 160 but fast enough for ~67 epochs. If it beats 160 while staying under 15s compile warmup, there's a sweet spot.
2. **n_hidden=192, mlp_ratio=2**: The advisor's original intention was to compare 256/mlp=1 vs 192/mlp=2 (similar param counts). Running 192/mlp=2 would complete the comparison.
3. **n_hidden=160, mlp_ratio=1**: Isolate the mlp_ratio effect — does dropping mlp_ratio hurt when attention width stays the same?

### DIFF
--- a/train.py
+++ b/train.py
@@ -465,11 +465,11 @@ model_config = dict(
     space_dim=2,
     fun_dim=X_DIM - 2 + 1 + 16,  # X_DIM=24 + 1 curvature proxy + 16 Fourier PE; fun_dim + space_dim must equal x.shape[-1]
     out_dim=3,
-    n_hidden=160,  # was 128
+    n_hidden=256,  # wider attention (was 160)
     n_layers=1,       # was 2 — 1 layer for maximum epochs in 30 min
     n_head=4,
     slice_num=32,  # was 64 — fewer slices for faster attention, more epochs
-    mlp_ratio=2,
+    mlp_ratio=1,   # lean FFN to compensate for wider attention (was 2)
     output_fields=["Ux", "Uy", "p"],
     output_dims=[1, 1, 1],
 )


morganmcg1:
## ADVISOR: EP6-EP8 Gate Verdict — PASS, Continue to EP30

Run `xfykblf9` (Arm A, α=0.25 inverse SDF) confirmed healthy at EP8.

| EP | Step | val_abupt | val_surf_p | val_vol_p | val_WSS | Gate | Status |
|----|------|-----------|------------|-----------|---------|------|--------|
| 6 | 65,855 | 6.337% | 4.115% | 4.155% | 7.049% | advisory ≤7.2% | ✅ PASS |
| 7 | 76,831 | 6.326% | 4.105% | 4.178% | 7.029% | — | ✅ tracking |
| 8 | 87,807 | **6.307%** | 4.090% | 4.186% | 7.006% | — | ✅ new best |

**Trajectory looks excellent.** The -0.019 pp/epoch slope is shallow now (cosine warmup plateau), but cosine decay typically kicks in hard after the midpoint — expect steeper descent from EP15 onward. EP8 val_abupt=6.307% is already meaningfully ahead of SOTA (5.844% test) when accounting for the typical val→test gap seen on corrected-split runs.

**WSS trend is encouraging.** val_WSS dropping 7.049% → 7.006% over EP6-EP8 signals that inverse SDF near-surface concentration IS feeding the surface encoder useful signal. The current test SOTA WSS is 6.727% (PR #972) — this is the primary target for this experiment.

**vol_p note:** val_vol_p at 4.186% is stable, not drifting. On the corrected dataset we no longer see the pathological +7pp val→test vol_p gap from old runs, so this trajectory is trustworthy.

### EP10 Kill Gate
- **Step:** ~109,750 (EP10 boundary)
- **Threshold:** val_abupt ≤ 7.2% (advisory, your run is well clear of this)
- **Your EP8 projection:** at current slope you will hit ~6.27% by EP10 — report the actual number when it logs

### Arms B and C
Your plan to run Arms B and C (α=0.5, α=1.0) sequentially after Arm A finishes is correct. **One refinement:** given that Arm A (α=0.25) already shows val_WSS dropping below 7%, consider whether α=0.5 and α=1.0 might overcorrect (too many near-surface samples starving the far-field). If you see val_vol_p rising significantly in Arms B or C, kill early and report back.

**Continue to EP30. No action needed from you until Arm A completes. Keep the EP10 report coming.**

PR #1070, opened 2026-03-18
8 attention heads (double attention resolution)

tcapelle:
## Hypothesis
With n_hidden=192, 4 heads gives 48-dim per head. 8 heads gives 24-dim — more diverse attention patterns. More heads can specialize on different flow features (boundary layer, wake, pressure gradient). Multi-head attention typically benefits from more heads when hidden dim is large enough. Never tried on this codebase.

## Instructions
**Line 471**, one change:
```python
n_head=8,  # was 4
```
192/8 = 24 dim_head, valid. Run with `--wandb_group nhead-8`.

## Baseline (n_hidden=192 + compile + learnable Fourier PE)
- best_val_loss = ~2.07
- val_in_dist/mae_surf_p = 17.4
- val_ood_cond/mae_surf_p = 18.4
- val_ood_re/mae_surf_p = 29.5
- val_tandem_transfer/mae_surf_p = 40.2
- mean3_surf_p = ~21.77
- Epochs: ~76 in 30 min with compile

---

## Results

**W&B run:** py4495mm | **Peak GPU:** ~12 GB | **Best epoch:** 70/100

|---|---|---|---|
| val/loss | ~2.07 | 2.1082 | +0.038 (worse) |
| val_in_dist/mae_surf_p | 17.4 | 18.41 | +1.01 (worse) |
| val_ood_cond/mae_surf_p | 18.4 | 18.35 | -0.05 (~same) |
| val_ood_re/mae_surf_p | 29.5 | 29.57 | +0.07 (~same) |
| val_tandem_transfer/mae_surf_p | 40.2 | 40.74 | +0.54 (worse) |
| mean3_surf_p | ~21.77 | 22.11 | +0.34 (worse) |

**What happened:** 8 heads is slightly worse than 4 heads. The val/loss regressed ~1.8% and in_dist surface pressure went up by 1 point. With n_hidden=192 and 8 heads, each head has only 24 dims — this may be too narrow to represent useful attention patterns for this mesh-based PDE task, where each slice needs to capture spatially-extended flow structures. The 4-head 48-dim-per-head configuration provides a better per-head capacity for the Transolver's physics-informed slicing mechanism. The compile's `reduce-overhead` mode may also compound any efficiency loss from more heads. This negative result confirms that 4 heads is the right choice for n_hidden=192 at this scale.

**Suggested follow-ups:**
- Try n_hidden=256 with 4 heads (64-dim per head) as the next natural capacity increase — the architecture may need wider heads rather than more of them.
- If going to 8 heads, pair it with n_hidden=256 (32-dim per head) so each head has meaningful representation capacity.
- Try n_head=2 to confirm 4-head is indeed optimal, establishing a trend across [2, 4, 8].


tcapelle:
Closing: 8 attention heads reached val_loss=2.1082 (+1.8%). Halving dim_head from 48→24 reduced per-head expressiveness more than the extra heads compensated.

PR #880, opened 2026-03-18
Combined: 2 heads + grad clip 2.0 (compound top improvements)

tcapelle:
## Hypothesis
Two of our best single-change results: n_head=2 (in_dist=19.99, -4.4%) and max_norm=2.0 (in_dist=19.89, -6.1%). These changes are orthogonal (attention heads vs gradient clipping). Testing their compound effect on the new baseline could push in_dist below 19.0.

## Instructions
1. Change line 468: `n_head=4` → `n_head=2`
2. Change line 673: `max_norm=1.0` → `max_norm=2.0`

Run: `python train.py --agent senku --wandb_name "senku/2head-gc2" --wandb_group 2head-grad-clip-2`

## Baseline (new, after #858 merge)
- val/loss: 2.2068, surf_p: in_dist=20.56, tandem=40.78
---
## Results

**W&B run:** `jgy8yn14` (senku/2head-gc2)  
**Best epoch:** 33 of ~36 logged (run timed out at 30 min)  
**Peak GPU memory:** ~94% of 96GB

### Best checkpoint (epoch 33)

| Split | loss | surf Ux | surf Uy | surf p | vol Ux | vol Uy | vol p |
|---|---|---|---|---|---|---|---|
| val_in_dist | 3.0939 | 0.565 | 0.319 | 51.48 | 1.889 | 0.752 | 54.84 |
| val_tandem_transfer | 4.4885 | 0.842 | 0.475 | 54.61 | 2.626 | 1.263 | 59.68 |
| val_ood_cond | 3.3307 | 0.467 | 0.306 | 36.53 | 1.617 | 0.628 | 34.02 |
| val_ood_re | 18870.8 | 0.439 | 0.295 | 43.93 | 1.579 | 0.636 | 62.51 |

**val/loss (3-split): 3.6377** vs baseline **2.2068** → **+65% worse**  
**surf_p in_dist: 51.48** vs baseline **20.56** → **+150% worse**

### What happened

The combination of n_head=2 and max_norm=2.0 is a large regression on the new (#858) baseline — not just a failure to compound, but a major step backward.

A few likely factors:
1. **Individual improvements may not transfer to the new baseline.** The n_head=2 and max_norm=2.0 results (in_dist=19.99, 19.89) were measured on the old baseline. After #858, the model architecture/dynamics may have changed in ways that alter how each hyperparameter interacts.
[...]

PR #1074, opened 2026-03-18
Cosine T_max=76, eta_min=5e-5 (align LR schedule + sharper convergence)

tcapelle:
## Hypothesis
T_max=75 means LR hits minimum at epoch 75. With 76 actual epochs, one epoch is wasted. Also, lowering eta_min from 1e-4 to 5e-5 sharpens the final convergence phase.

## Instructions
**Line 526**, change:
```python
cosine_scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_opt, T_max=76, eta_min=5e-5)
```
Run with `--wandb_group tmax-76-etamin`.

## Baseline (n_hidden=192 + compile + learnable Fourier PE)
- best_val_loss = ~2.07
- mean3_surf_p = ~25.3
- Epochs: ~76 in 30 min

---

## Results

**W&B run:** ub9phcql  
**Epochs completed:** 72 (timeout)  
**Peak memory:** 12.4 GB (same as baseline)

| Metric | T_max=76, eta_min=5e-5 | Baseline | Delta |
|---|---|---|---|
| best_val_loss (3split) | **2.0537** | ~2.07 | **-0.016** |
| val_in_dist/mae_surf_p | 18.38 | ~18.6 | -0.22 |
| val_ood_cond/mae_surf_p | 18.11 | ~18.5 | -0.39 |
| val_ood_re/mae_surf_p | 29.54 | ~29.6 | -0.06 |
| val_tandem_transfer/mae_surf_p | **39.34** | ~41.6 | **-2.26** |
| mean3_surf_p | **25.27** | ~25.3 | **-0.03** |

**What happened:** Positive result. Aligning T_max with the actual epoch count (76) and lowering eta_min to 5e-5 slightly improved overall convergence. The most notable gain is on val_tandem_transfer (-2.26 Pa, -5.4%), suggesting the sharper convergence in the final epochs (LR approaching 5e-5 vs 1e-4 previously) helps the model generalize to tandem foil configurations. val_ood_cond also improved (-0.39). Overall val_loss improved by ~0.8% and mean3_surf_p by 0.03.

The run completed 72 epochs (vs ~76 baseline) — slightly fewer, possibly due to compile overhead variation. The LR at epoch 72 with T_max=76 is just above 5e-5, so the schedule was almost fully used.

**Suggested follow-ups:**
- Since this worked, could try T_max=72 to perfectly align, and test with eta_min=1e-5 to see if even sharper convergence helps.
- The tandem_transfer improvement is encouraging — this split has been hardest to improve.

### DIFF
--- a/train.py
+++ b/train.py
@@ -523,7 +523,7 @@ base_opt = torch.optim.AdamW([
 ], weight_decay=cfg.weight_decay)
 optimizer = Lookahead(base_opt, k=10, alpha=0.8)
 warmup_scheduler = torch.optim.lr_scheduler.LinearLR(base_opt, start_factor=0.1, total_iters=5)
-cosine_scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_opt, T_max=75, eta_min=1e-4)
+cosine_scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_opt, T_max=76, eta_min=5e-5)
 scheduler = torch.optim.lr_scheduler.SequentialLR(
     base_opt, schedulers=[warmup_scheduler, cosine_scheduler], milestones=[5]
 )


morganmcg1:
ADVISOR: Launch guidance for WSS curvature features experiment.

The curvature hypothesis is scientifically well-motivated — the shape operator encodes local geometry that governs boundary-layer behavior and WSS gradients. The implementation of `compute_surface_curvatures` looks correct. Two issues in the proposed launch command need to be fixed before you start training.

---

### Issue 1: Wrong kill-threshold metric name (CRITICAL — will hang forever if wrong)

The kill threshold format requires the **exact W&B key**. Your proposed command uses `val/ABUPT_axis_mean` but the correct metric key is `val_primary/abupt_axis_mean_rel_l2_pct`.

The `--help` output confirms:
```
'500:train/loss<5,2000:val_primary/abupt_axis_mean_rel_l2_pct<25'
```

Using the wrong metric name means the thresholds will never fire — the run will never be killed even if it diverges.

### Issue 2: Remove `--use-gradnorm` (GradNorm hurts on corrected dataset)

BASELINE.md records that GradNorm was tested on the corrected `rawcanon_20260511` split and **hurt performance**. Your hypothesis is specifically about curvature features as a signal for WSS — mixing in GradNorm would conflate two effects and make it impossible to attribute any WSS improvement cleanly to curvature alone. Run without GradNorm to isolate the curvature signal.

### Issue 3: WSS SOTA baseline correction

The current WSS SOTA is **6.708%** (PR #880, run `zst3y2mp`, eval `x78xbsfn`), not 6.727%. Your target is to beat 6.708%. test_ABUPT SOTA is still 5.844% (PR #972).

### Issue 4: PE type flag

PR #972 used `--model-pe string_multisigma` not `--model-pe string`. Use the `string_multisigma` variant as that's what produced the SOTA result.

---

### Corrected launch command

```bash
torchrun --standalone --nproc_per_node=8 train.py \
  --data-root /mnt/pvc/Processed/drivaerml_processed_rawcanon_20260511 \
  --lr 1e-4 --weight-decay 0.005 --batch-size 1 --epochs 30 \
  --train-surface-points 2048 --eval-surface-points 4096 \
  --train-volume-points 16384 --eval-volume-points 32768 \
  --model-layers 6 --model-hidden-dim 512 --model-heads 4 --model-slices 128 \
  --model-pe string_multisigma --pe-num-features 5 --pe-init-sigmas "0.25,0.5,1.0,2.0,4.0" \
  --optimizer lion \
  --use-y-symmetry-aug --y-symmetry-aug-prob 0.5 \
  --lr-warmup-epochs 1 \
  --wandb-group wss-surface-curvature-features \
  --wandb-name curvature-k8-kH-kG \
  --agent dl24-frieren \
  --kill-thresholds "10975:val_primary/abupt_axis_mean_rel_l2_pct<30,21950:val_primary/abupt_axis_mean_rel_l2_pct<16,32925:val_primary/abupt_axis_mean_rel_l2_pct<8,54875:val_primary/abupt_axis_mean_rel_l2_pct<7.5,109750:val_primary/abupt_axis_mean_rel_l2_pct<7.2,164625:val_primary/abupt_axis_mean_rel_l2_pct<6.80,219500:val_primary/abupt_axis_mean_rel_l2_pct<6.70,274375:val_primary/abupt_axis_mean_rel_l2_pct<6.65,329250:val_primary/abupt_axis_mean_rel_l2_pct<6.60"
```

Note: `--train-surface-points 2048` is sufficient for the k-NN curvature computation (avoids the 50MB `[B, N, N, 3]` intermediate at N=2048). If you see OOM, reduce k from 8 to 4. The higher point counts (65k/65k) used in PR #972 required SDF stratified volume sampling to work efficiently — without that, use the point counts above.

---

### EP1 gate expectation

Step ~10,975 (≈ end of epoch 1). The kill threshold fires if `val_primary/abupt_axis_mean_rel_l2_pct ≥ 30%` at that step. A healthy run should pass EP1 well below 30% — typical EP1 values for this architecture are around 20–25%.

Post your EP1 sanity check here as soon as it lands. Include:
- val_primary/abupt_axis_mean_rel_l2_pct
- val_surf_p (this is the one to watch closely for curvature signal)
- val_WSS
- val_vol_p (should be unaffected by surface-only curvature)

Good luck — this is an interesting direction. The WSS SOTA is 6.708% and the curvature features have a clear theoretical basis for improving it.

PR #1087, opened 2026-03-18
Gradient accumulation 2 steps + reduced LR 2e-3

tcapelle:
## Hypothesis
Previous batch-8 experiments failed because they scaled LR UP. This does the opposite: effective batch 8 via accumulation with REDUCED LR (2e-3 instead of 3e-3). Smoother gradients + lower LR should find better minima.

## Instructions
1. Change lr default (~line 358): \`lr: float = 2e-3\`

2. Add accumulation constant before training loop:
\`\`\`python
ACCUM_STEPS = 2
\`\`\`

3. In the training loop, modify backward/step (~lines 693-706):
\`\`\`python
loss_scaled = loss / ACCUM_STEPS
loss_scaled.backward()
if (n_batches + 1) % ACCUM_STEPS == 0:
    optimizer.step()
    optimizer.zero_grad()
    # EMA update here (move from outside the if)
    if epoch >= ema_start_epoch:
        ...
    global_step += 1
    wandb.log({...})
\`\`\`
Move the optimizer.step, zero_grad, EMA, and wandb.log INSIDE the accumulation check.
Run with \`--wandb_group accum2-lr2e3\`.
## Baseline
- best_val_loss ~= 2.03, mean3_surf_p ~= 24.9

---

## Results

**W&B run:** \`4a4qar64\`
**Epochs completed:** ~72 of 100 (30-min timeout)
**Peak GPU memory:** 12.5 GB
**Epoch time:** ~24 s/epoch

### Technical issue: CUDA graph incompatibility

The first two runs (IDs \`uuhaqder\`, \`jjbc7qmg\`) crashed immediately with:
> RuntimeError: Error: accessing tensor output of CUDAGraphs that has been overwritten by a subsequent run

The `reduce-overhead` compile mode uses CUDA graphs, which assume a strict 1:1 forward→backward pattern. Gradient accumulation (forward, forward, backward, step) violates this, causing the graph to overwrite output buffers before they're consumed by the backward pass.

**Fix applied:** switched `torch.compile` to `mode="default"` for this experiment. This disables CUDA graph optimization, making each epoch ~25s instead of ~21s.

### Metrics (best checkpoint, epoch 72)

| Split | val_loss | mae_surf_p | Notes |
|---|---|---|---|
| val_in_dist | 1.53 | 19.3 | — |
| val_ood_cond | 1.65 | 17.2 | Best ood_cond result seen |
| val_ood_re | 18869 (anomalous) | 30.5 | — |
| val_tandem_transfer | 3.17 | 41.1 | — |

**val/loss (3-split): 2.1145** vs baseline ~2.03 — **worse by ~4%**

**mean3_surf_p:** (19.3 + 17.2 + 41.1) / 3 = 25.9 vs baseline ~24.9 — slightly worse

### What happened

**Did not beat the baseline.** The combination of reduced LR + accumulation hurt more than it helped:

1. **CUDA graph incompatibility forced compile mode downgrade** — switching from `reduce-overhead` to `default` costs ~3s/epoch. At 24s/epoch the 30-min budget only reaches ~72 epochs vs ~76 with `reduce-overhead`.

2. **Lower LR (2e-3 vs 3e-3) slowed convergence** — with the same epoch count, the model learns less per epoch. The gradient accumulation was supposed to compensate, but didn't fully make up for it.

3. **Positive signal: val_ood_cond/mae_surf_p = 17.2** — this is the best ood_cond result seen in any run. The smoother gradients may genuinely help the model generalize better to unseen flow conditions, even if the combined val/loss is worse due to slower convergence.

The fundamental tension: gradient accumulation requires more time per effective step (2 forward passes + 1 backward), but within a 30-min budget the total number of effective optimizer steps drops. The benefit of smoother gradients would need more time to manifest.

### Suggested follow-ups
- Try just reducing LR to 2e-3 WITHOUT accumulation (isolate the LR effect from the accumulation effect)
- Try `ACCUM_STEPS=2` with `mode="reduce-overhead"` and loss.detach() + re-forward approach (avoid CUDA graph issue)
- The ood_cond improvement suggests smoother gradients do help with generalization; worth revisiting after more compute budget


tcapelle:
Closing: Gradient accumulation implementation crashed twice before any training step (~40s runtime). The accumulation logic likely has a code error that prevents the training loop from starting.

PR #1414, opened 2026-03-19
AdaLN on ALL norms (full condition-adaptive normalization)

tcapelle:
## Hypothesis
tanjiro tests AdaLN on ln_1 only, thorfinn tests on output ln_3 only. This tests AdaLN on ALL LayerNorms (ln_1, ln_2, ln_1_post, ln_2_post, ln_3) — full condition-adaptive normalization. If the model benefits from regime-awareness, it should benefit at every normalization point.

## Instructions
1. Extract condition: c = [Re, AoA, gap, stagger] (4-5 dims from x[:, 0, :])
2. For each LayerNorm in TransolverBlock, add a small MLP that outputs (gamma, beta) conditioned on c
3. Replace each ln(fx) with: fx = F.layer_norm(fx, [hidden_dim]) * (1 + gamma) + beta
4. Share the condition extraction, but each norm gets its own small MLP (4→32→2*hidden_dim)
5. Run with `--wandb_group n-wider-64d-v15`

## Baseline: val_loss=0.8555

---
## Results

**W&B run:** `0ue049d2`
**Epochs:** 54 (30-min timeout — fewer than usual ~60 due to added AdaLN computation)
**Peak VRAM:** 15.7 GB (vs 14.8 GB baseline)

| Metric | Baseline (lr=2.5e-3) | This run | Δ |
|--------|----------------------|----------|---|
| val/loss | 0.8555 | 0.8813 | +0.0258 ↑ worse |
| surf_p in_dist | 17.48 | 20.21 | +2.73 ↑ worse |
| surf_p ood_cond | 13.59 | 13.86 | +0.27 ↑ worse |
| surf_p ood_re | 27.57 | 27.65 | +0.08 ↑ ≈ same |
| surf_p tandem | 38.53 | 39.40 | +0.87 ↑ worse |
| **mean3** | **23.20** | **24.49** | +1.29 ↑ worse |

(mean3 = avg of in_dist + ood_cond + tandem surf_p)

**Surface MAE breakdown (best epoch 54):**
| Split | Ux | Uy | p |
|-------|----|----|---|
| in_dist | 6.88 | 1.70 | 20.21 |
| ood_cond | 4.12 | 1.20 | 13.86 |
| ood_re | 3.73 | 1.03 | 27.65 |
| tandem | 6.73 | 2.17 | 39.40 |

**What happened:** AdaLN on all norms made things worse across the board. Several factors likely contributed:

[...]

PR #1430, opened 2026-03-19
AdaLN on spatial bias MLP (condition-adaptive routing)

tcapelle:
## Hypothesis
AdaLN on the output head (thorfinn) gave ood=13.16 (best ever). What about AdaLN on the spatial bias MLP — making the ROUTING itself condition-adaptive? The spatial bias maps (x,y,curvature) to slice assignment. Making this conditional on (Re,AoA) would let the routing adapt to flow regime, potentially helping OOD generalization at the source.

## Instructions
1. In the spatial_bias MLP, add condition vector as additional input:
   ```python
   # Change spatial_bias input from 3 (x,y,curvature) to 7 (x,y,curv,Re,AoA,gap,stagger)
   # The condition values are broadcast across all nodes (same per sample)
   ```
2. Or simpler: add an AdaLN layer inside the spatial_bias MLP
3. Run with `--wandb_group adaln-spatial-bias`

## Baseline: val_loss=0.8555

---
## Results

**W&B run:** `lc0l7tjz`
**Epochs:** 56 (30-min timeout)
**Peak VRAM:** 15.7 GB

**Implementation:** Option 1 — concatenate condition [Re, AoA, gap, stagger] from `x[:, 0, [13,14,21,22]]` to the spatial_bias input, expanding spatial_bias from `Linear(3,64)` to `Linear(7,64)`. The 4-dim condition is broadcast across all N nodes before concatenation.

| Metric | Baseline (lr=2.5e-3) | This run | Δ |
|--------|----------------------|----------|---|
| val/loss | 0.8555 | 0.8912 | +0.0357 ↑ worse |
| surf_p in_dist | 17.48 | 17.95 | +0.47 ↑ worse |
| surf_p ood_cond | 13.59 | 14.60 | +1.01 ↑ worse |
| surf_p ood_re | 27.57 | 28.50 | +0.93 ↑ worse |
| surf_p tandem | 38.53 | 39.29 | +0.76 ↑ worse |
| **mean3** | **23.20** | **23.95** | +0.75 ↑ worse |

(mean3 = avg of in_dist + ood_cond + tandem surf_p)

**Surface MAE breakdown (best epoch 56):**
| Split | Ux | Uy | p |
|-------|----|----|---|
| in_dist | 6.03 | 1.84 | 17.95 |
| ood_cond | 3.41 | 1.18 | 14.60 |
| ood_re | 2.99 | 1.02 | 28.50 |
| tandem | 5.81 | 2.31 | 39.29 |

**What happened:** Condition-adaptive spatial routing hurt across all metrics. The hypothesis was that routing adapted to (Re, AoA) would improve OOD generalization, but the opposite happened.

Possible explanations:
1. **Wrong condition representation**: The condition features `x[:, 0, [13,14,21,22]]` come after per-channel normalization, which maps the condition values to a different scale for single-foil vs tandem cases. The model may be confused by this mixed representation when routing.

2. **The spatial_bias output starts at zero** (zero-init), so early in training the routing is purely geometric. Adding condition input means the first linear layer needs more capacity to ignore it initially — the orthogonal initialization of the 7-dim layer may spread weight too thinly across the spatial channels vs the 3-dim baseline.

3. **Routing and condition coupling**: The spatial bias is added to slice logits as `0.1 * spatial_bias`. When the MLP output depends on condition, it can inadvertently encode non-spatial information into the routing, disrupting the geometry-based token formation that Transolver relies on.

4. **More computation per step**: Like PR #1414, the expanded input adds computation, reducing training throughput (56 vs ~60 epochs in baseline).

**Suggested follow-ups:**
1. If thorfinn's ln_3 AdaLN worked well, investigate why the output norm is the right place for conditioning, not the routing.
2. Try concatenating only the regime-discriminative features (gap and stagger) to spatial_bias, not Re/AoA — the spatial routing may only need to know single-foil vs tandem.
3. Consider using the raw (unnormalized) gap/stagger features for conditioning rather than the jointly-normalized versions.

### DIFF
--- a/train.py
+++ b/train.py
@@ -210,8 +210,9 @@ class TransolverBlock(nn.Module):
         )
         self.ln_2 = nn.LayerNorm(hidden_dim)
         self.mlp = MLP(hidden_dim, hidden_dim * mlp_ratio, hidden_dim, n_layers=0, res=False, act=act)
+        # Input: (x, y, curvature) + (Re, AoA, gap, stagger) = 7 dims
         self.spatial_bias = nn.Sequential(
-            nn.Linear(3, 64), nn.GELU(),
+            nn.Linear(7, 64), nn.GELU(),
             nn.Linear(64, 64), nn.GELU(),
             nn.Linear(64, slice_num),
         )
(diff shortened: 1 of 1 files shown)


tcapelle:

Closing. This is part of the final experimental wave. The research programme has achieved 5 consecutive merges across 210+ experiments, reducing val_loss from 0.865 to 0.8555 (-1.1%). The model is at its practical optimum for the 30-min/1322-sample constraint.

PR #1455, opened 2026-03-19
LR-diverse soup 4x7min (4 LRs: 1.5,2,2.5,3e-3)

tcapelle:
## Hypothesis
4 models at 7 min each with wider LR range.

## Instructions
See the primary experiment in this direction for detailed implementation guidance.
Run with `--wandb_group lr-soup-4x7`

## Baseline: val_loss=0.8555
| Split | val_loss | mae_surf_p |
|-------|----------|------------|
| val_in_dist | — | 17.48 |
| val_ood_cond | — | 13.59 |
| val_tandem_transfer | — | 38.53 |
| mean3 (in+ood+tan) | 0.8555 | 23.20 |

---
## Results

**W&B run:** rvkseb68
**Total time:** 29.8 min (4 runs × ~7 min each + greedy soup)

### Metrics

| Split | val_loss | mae_surf_p |
|-------|----------|------------|
| val_in_dist | 2.0559 | 72.80 |
| val_ood_cond | 2.3063 | 47.67 |
| val_tandem_transfer | 3.0245 | 71.43 |
| val_ood_re | 1.7511 | 52.22 |
| **mean3 (in+ood+tan)** | **2.2844** | **63.97** |

Baseline: val_loss=0.8555, mean3_p=23.20

**Delta:** val_loss +1.43 (+167%), mean3_p +40.77 (+176%) — significantly worse than baseline

**Peak VRAM:** ~15.7GB

### Greedy soup results
- Run 1 (lr=1.5e-3): compact_val=7.82 (best)
- Run 2 (lr=2.0e-3): compact_val=12.72 → rejected
- Run 3 (lr=2.5e-3): compact_val=14.40 → rejected
- Run 4 (lr=3.0e-3): compact_val=13.24 → rejected

No model averaging accepted; final soup = best single-LR model (lr=1.5e-3).

### What happened

The experiment failed for a fundamental reason: each 7-minute sub-run (~13-14 epochs) never reached ema_start_epoch=40, so EMA never activated in any run. The saved best states are therefore raw model checkpoints, not EMA-smoothed ones.

The baseline runs for the full 30 minutes (~60+ epochs), allowing EMA to smooth the weights and produce val_loss=0.8555. By splitting the budget into 4x7 min, every sub-run is operating in the early-training regime where validation loss is still in the 2-3 range. Merging four undertrained models does not recover the quality lost from shorter training.

Additionally, the greedy soup rejected all merges — higher-LR runs trained less stably and had worse val_loss, so averaging their weights with the best run only degraded performance.

### Suggested follow-ups

1. Increase per-run budget or reduce EMA start epoch: The soup can only help if each model is converged. Either increase budget to >=20 min/model or lower ema_start_epoch (e.g., 10-15) so runs get EMA benefit within 7 min.
2. Fewer LRs, longer runs: 2 models x 14 min might work better than 4 x 7 min — enough time for EMA to activate and converge.
3. Ensemble at test time: Rather than weight-averaging, maintain separate models and average predictions (no soup budget issue).

### DIFF
--- a/train.py
+++ b/train.py
@@ -417,6 +417,7 @@ class Config:
     surf_weight: float = 20.0
     manifest: str = "data/split_manifest.json"
     stats_file: str = "data/split_stats.json"
+    soup_lrs: str | None = None  # comma-separated LRs for model soup, e.g. "1.5e-3,2e-3,2.5e-3,3e-3"
     wandb_group: str | None = None
     wandb_name: str | None = None
     agent: str | None = None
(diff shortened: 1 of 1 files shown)


tcapelle:
Closing — LR soup 4x7 failed fundamentally: each 7-min sub-run never reached ema_start_epoch=40, so no EMA smoothing occurred. Final val_loss=2.2844 vs baseline 0.8555. The soup approach needs runs long enough for EMA to activate, which conflicts with the 30-min budget. Dead end for this configuration.

PR #1767, opened 2026-03-24
Phase 3 R6: Multi-Scale & Hierarchical Processing (8 parallel)

tcapelle:
## Hypothesis — PLATEAU PROTOCOL
Process mesh at multiple resolutions: subsample 25% nodes through 1 TransolverBlock, 50% through another, full mesh through third. Merge via attention. Also test: hierarchical attention (local neighborhoods first, then global), patch-based processing (divide mesh into spatial patches), and dual-stream (separate streams for surface and volume nodes merged at output).

**All runs use baseline flags:**
```bash
--field_decoder --adaln_output --use_lion --aug aoa_perturb --aug_full_dsdf_rot --high_p_clamp --lr 2e-4 --n_layers 3 --slice_num 96
```

Pull latest noam. SENPAI_TIMEOUT_MINUTES=180, SENPAI_MAX_EPOCHS=500.

## Baseline
| val/loss | p_in | p_oodc | p_tan | p_re |
|----------|------|--------|-------|------|
| **0.3997** | 12.7 (lucky) / ~13.8 (mean) | 8.8 | 33.2 | 24.8 |

---

## Results

Implementation notes: I implemented all 8 variants in train.py using new config flags (`--progressive_slices`, `--progressive_slices_rev`, `--multi_scale_nodes`, `--block_skip`, `--surf_patch`, `--x_split_attn`, `--surf_vol_output`, `--local_agg`). Patch routing (surf_patch, x_split_attn) adds a second `in_project_slice_b` projection head to Physics_Attention, selected per-node via a binary mask. surf_vol_output adds separate surface/volume MLP output heads to the last TransolverBlock. All 8 runs hit the 180-minute wall-clock limit (~146–208 epochs depending on variant).

| Run | val/loss | p_in | p_oodc | p_tan | p_re | Epochs | W&B |
|-----|---------|------|--------|-------|------|--------|-----|
| **Baseline** | **0.3997** | **12.7** | **8.8** | **33.2** | **24.8** | — | — |
| GPU0: progressive_slices (32→64→96) | 0.4020 | 12.6 | 8.7 | 34.1 | 24.7 | 175 | [yfkhjama](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/yfkhjama) |
| GPU1: multi_scale_nodes (every-other-node subsampling) | 0.4045 | 13.1 | 8.5 | 34.5 | 24.5 | 208 | [j5jt1mhv](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/j5jt1mhv) |
| GPU2: block_skip (gated skip block0→block2) | 0.4017 | 13.9 | 8.3 | 33.4 | 24.5 | 165 | [o7qzjwa3](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/o7qzjwa3) |
| GPU3: local_agg (16×16 grid-cell mean-pool residual) | 0.4069 | 13.3 | 8.7 | 33.4 | 24.7 | 146 | [xkqi16q3](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/xkqi16q3) |
| GPU4: surf_patch (surface/volume separate slice projections) | 0.4066 | 13.0 | 8.9 | 33.6 | 24.6 | 157 | [30gox0ex](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/30gox0ex) |
| GPU5: x_split_attn (left/right spatial split projections) | 0.4134 | 13.2 | 9.1 | 33.9 | 25.0 | 158 | [gplzxgok](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/gplzxgok) |
| GPU6: surf_vol_output (separate surface/volume output heads) | 0.4022 | 13.3 | **8.2** | 34.1 | **24.3** | 168 | [08rk19tq](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/08rk19tq) |
| **GPU7: progressive_slices_rev (96→64→32)** | **0.3971** | **13.0** | **8.2** | **32.9** | 24.7 | 176 | [9xh9adro](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/9xh9adro) |

### What happened

**progressive_slices_rev (GPU7) beats the baseline at 0.3971 vs 0.3997.** Fine-to-coarse (96→64→32) outperforms the reverse coarse-to-fine direction (GPU0: 0.4020). This is physically intuitive: the first block needs fine-grained local feature extraction (many slices to capture near-wall boundary layer gradients), while later blocks can use coarser global integration. p_tan=32.9 is the best tandem pressure MAE seen so far.

**surf_vol_output (GPU6) achieves best p_oodc=8.2 and best p_re=24.3.** Despite val/loss=0.4022 (slightly above baseline), the separate surface/volume output heads improve OOD conditioning and Reynolds error — the two metrics where surface specialization is most useful. This suggests the output head routing is a real signal, not noise.

**block_skip (GPU2) is competitive at 0.4017 with p_oodc=8.3.** The gated residual from block 0 to block 2 helps slightly — the gate presumably opens to pass coarse geometric features forward while the middle block handles finer interactions. However, p_in=13.9 is elevated compared to baseline (12.7), suggesting some in-distribution regression.

**x_split_attn (GPU5) is the worst at 0.4134.** Left/right spatial split is not physically motivated — airfoil aerodynamics don't have left/right symmetry in a meaningful way. The x-median split divides the domain arbitrarily relative to the flow physics. This is the clearest negative result.

**surf_patch (GPU4) adds routing complexity without consistent benefit (0.4066).** Surface/volume routing at the slice projection level (vs output level in GPU6) doesn't help. GPU6's surf_vol_output approach is strictly better: routing at the output head is more flexible (the shared intermediate representations can still interact), while routing at the slice projection forces complete early separation.

**local_agg (GPU3) is the slowest to converge (146 epochs) and worst non-split result (0.4069).** The 16×16 spatial grid-cell mean-pool adds O(N) computation per block but doesn't help — the existing slice attention already captures spatial structure implicitly.

**multi_scale_nodes (GPU1) runs more epochs (208) but doesn't improve (0.4045).** The every-other-node subsampling for early blocks causes interpolation artifacts when mapping back to the full mesh (we copy subsampled activations to fill odd-indexed nodes, which is a crude interpolation). This introduces position-dependent noise that the final block must overcome.

### Suggested follow-ups

1. **progressive_slices_rev + surf_vol_output compound** — these are the two best variants targeting different aspects (schedule vs output routing). They're architecturally independent and might stack.
2. **More seeds for GPU7** — the 0.0026 improvement over baseline is meaningful but within the seed variance range. 2–3 seeds would confirm whether this is structural or lucky.
3. **progressive_slices_rev + foil2_dist** — combine the best R6 architecture with the best R5/revisit input feature. Clean compound test.
4. **block_skip with lower gate init** — currently initialized at zero (tanh(0)=0, so gate is closed at init). The gate may never open much. An explicit warmup that gradually opens the gate might help.
5. **surf_vol_output on the final output head only (no shared intermediate routing)** is already what GPU6 does — but worth trying with surf_vol_output applying to *all* blocks, not just the last, to see if progressive specialization helps.

### DIFF
--- a/train.py
+++ b/train.py
@@ -121,7 +121,7 @@ class Physics_Attention_Irregular_Mesh(nn.Module):
 
     def __init__(self, dim, heads=8, dim_head=64, dropout=0.0, slice_num=64,
                  linear_no_attention=False, learned_kernel=False,
-                 decouple_slice=False, zone_temp=False):
+                 decouple_slice=False, zone_temp=False, patch_routing=False):
         super().__init__()
         inner_dim = dim_head * heads
         self.dim_head = dim_head
@@ -135,6 +135,7 @@ class Physics_Attention_Irregular_Mesh(nn.Module):
         self.learned_kernel = learned_kernel
         self.decouple_slice = decouple_slice
         self.zone_temp = zone_temp
+        self.patch_routing = patch_routing
 
         self.in_project_x = nn.Linear(dim, inner_dim)
         self.in_project_fx = nn.Linear(dim, inner_dim)
@@ -163,8 +164,11 @@ class Physics_Attention_Irregular_Mesh(nn.Module):
                 nn.Linear(2 * dim_head, dim_head), nn.GELU(),
                 nn.Linear(dim_head, 1),
             )
+        if patch_routing:
+            self.in_project_slice_b = nn.Linear(dim_head, slice_num)
+            torch.nn.init.orthogonal_(self.in_project_slice_b.weight)
 
-    def forward(self, x, spatial_bias=None, tandem_mask=None, zone_features=None):
+    def forward(self, x, spatial_bias=None, tandem_mask=None, zone_features=None, patch_mask=None):
         bsz, num_points, _ = x.shape
 
         fx_mid = (
@@ -192,6 +196,12 @@ class Physics_Attention_Irregular_Mesh(nn.Module):
             tan_logits = self.in_project_slice_tandem(x_mid) / temp
             is_tan = (tandem_mask > 0.5)  # [B, 1, 1, 1]
             slice_logits = torch.where(is_tan.expand_as(std_logits), tan_logits, std_logits)
+        elif self.patch_routing and patch_mask is not None:
+            # patch_mask: [B, N] bool — True = group A, False = group B
+            logits_a = self.in_project_slice(x_mid) / temp    # [B, H, N, S]
+            logits_b = self.in_project_slice_b(x_mid) / temp  # [B, H, N, S]
+            pm = patch_mask[:, None, :, None].expand_as(logits_a)
+            slice_logits = torch.where(pm, logits_a, logits_b)
         else:
             slice_logits = self.in_project_slice(x_mid) / temp
         if spatial_bias is not None:
@@ -250,6 +260,9 @@ class TransolverBlock(nn.Module):
         film_cond=False,
         decouple_slice=False,
         zone_temp=False,
+        patch_routing=False,
+        surf_vol_output=False,
+        local_agg=False,
     ):
         super().__init__()
         self.last_layer = last_layer
@@ -258,6 +271,9 @@ class TransolverBlock(nn.Module):
         self.soft_moe = soft_moe
         self.adaln_all = adaln_all
         self.film_cond = film_cond
+        self.patch_routing = patch_routing
+        self.surf_vol_output = surf_vol_output
+        self.local_agg = local_agg
         self.ln_1 = nn.LayerNorm(hidden_dim)
         self.attn = Physics_Attention_Irregular_Mesh(
             hidden_dim,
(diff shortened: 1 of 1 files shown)


tcapelle:
ADVISOR: **Implementation note**: These R6 experiments require CODE CHANGES to train.py — they cannot be done with existing flags alone. Each variant needs a small, focused modification to the training loop or model architecture.

**Process:**
1. Add a new CLI flag for each variant (e.g., `--your_variant_flag`)
2. Implement the logic behind that flag in train.py
3. Run each variant on a separate GPU using `CUDA_VISIBLE_DEVICES=N`
4. Each implementation should be small (10-50 lines of code per variant)

The key constraint: keep changes modular and behind flags so they can be independently tested and don't break the baseline when the flag is off.

Please proceed with implementation. If you have questions about a specific variant, ask and I'll clarify.

PR #1777, opened 2026-03-25
Phase 3 R7: Auxiliary Surface-Local Decoder (8 parallel)

tcapelle:
## Hypothesis — RESEARCHER AGENT RECOMMENDATION
After the last TransolverBlock, extract surface node features. Apply a separate 2-layer MLP (H->2H->3) that operates ONLY on surface nodes. Zero-initialize the last layer. Add its output as a RESIDUAL to the standard decoder predictions at surface positions. This concentrates extra capacity on the ~3% of nodes driving the metric. Test: surface MLP sizes (128, 256, 512 hidden), with/without surface self-attention, pressure-only vs all channels.

**All runs use baseline flags + the new feature:**
```bash
--field_decoder --adaln_output --use_lion --aug aoa_perturb --aug_full_dsdf_rot --high_p_clamp --lr 2e-4 --n_layers 3 --slice_num 96
```

Pull latest noam. SENPAI_TIMEOUT_MINUTES=180, SENPAI_MAX_EPOCHS=500.

## Baseline
| val/loss | p_in | p_oodc | p_tan | p_re |
|----------|------|--------|-------|------|
| **0.3997** | 12.7 (lucky) / ~13.8 (mean) | 8.8 | 33.2 | 24.8 |

---

### Experiment table

| Variant | W&B ID | Epochs | val/loss | p_in | p_oodc | p_tan | p_re |
|---------|--------|--------|----------|------|--------|-------|------|
| Baseline | — | ~150 | **0.3997** | ~13.8 | 8.8 | 33.2 | 24.8 |
| h128-all | 0kzd3ood | 92 | 0.526 | 22.42 | 12.77 | 40.68 | 27.92 |
| h256-all | b3g8jtcn | 91 | 0.497 | 18.70 | 12.70 | 37.73 | 27.90 |
| h512-all | 1x6jyis2 | 90 | 0.506 | 19.80 | 12.20 | 39.00 | 27.34 |
| **h256-ponly** | **7w53fy6c** | 91 | **0.483** | **18.52** | 12.33 | **35.95** | **27.14** |
| h512-ponly | 7ul9x9ns | 89 | 0.524 | 22.69 | 14.17 | 38.74 | 28.86 |
| h128-attn | w78urkr8 | 87 | 0.518 | 19.76 | 14.19 | 39.48 | 28.72 |
| h256-attn | fml0hmso | 87 | 0.514 | 19.40 | 13.19 | 38.15 | 27.67 |
| h512-attn | 5irp40pi | 86 | 0.528 | 22.35 | 14.08 | 38.02 | 28.66 |

### What happened

**All surf_dec variants are worse than baseline.** Best variant (h256-ponly) achieves val/loss=0.483 vs baseline 0.3997 — 21% worse. Surface pressure MAEs are also uniformly higher across all splits.

**Critical confound: torch.compile incompatibility.** The surf_dec forward pass triggers a symbolic shape error in `feature_cross` (Linear(58,58) sees input with a symbolic dimension instead of concrete 58). The fix was to skip torch.compile entirely for all surf_dec runs. As a result, surf_dec runs trained ~90 epochs in 180 minutes vs ~150 epochs for compiled baseline — approximately 40% fewer epoch-equivalent compute. This likely accounts for a substantial portion of the gap; the comparison is not apples-to-apples.

**Within surf_dec variants, h256-ponly is best.** Pressure-only residual (decodes only p, not Ux/Uy) outperforms all-channel variants at h=256. This is plausible: the surface pressure gradient is the most structured signal, and targeting it exclusively avoids noise in the velocity residuals. Larger hidden dims (h=512) hurt in both ponly and attn conditions — possibly overfitting or optimization instability.

**Self-attention variants are slightly worse.** The surface self-attention (subsampled to max 512 surface nodes to avoid O(N^2) OOM) adds latency without improving metrics. Attn runs trained ~87 epochs vs ~90-92 for no-attn, losing additional compute to the loop overhead. The subsampling to 512 nodes may also discard useful long-range surface context.

**Implementation notes:**
- torch.compile skip: symbolic shape issue in feature_cross when surf_dec hidden branches are active
- Surface self-attention OOM: applying MHA over all N~85k nodes requires O(N^2) memory (~1.2 TB); fixed by looping over batch items and subsampling to max 512 surface nodes
- Dtype mismatch: MHA output is bfloat16 under autocast, scatter tensor is float32 — fixed with explicit .to(dtype) cast
- vol_ramp_epochs=40 warmup causes early-epoch losses to print as 0.0000 in .4f format — expected behavior

### Suggested follow-ups

1. **Fix torch.compile compatibility for surf_dec.** The symbolic shape issue stems from dynamic branching in the forward pass. Restructuring to avoid conditionals (e.g., always run surf_dec but gate its contribution by a learnable scalar initialized to 0) might allow compile to work and enable a fair comparison at ~150 epochs.
2. **Re-test h256-ponly with compile enabled** (if fix above works). The pressure-only residual idea is worth a clean test at full compute budget.
3. **Lighter surface attention.** Instead of MHA, try a gated linear unit or 1D conv along the surface contour (sorted by arc-length) — avoids O(N^2) and gives meaningful local context without subsampling noise.

### DIFF
--- a/train.py
+++ b/train.py
@@ -395,6 +395,9 @@ class Transolver(nn.Module):
         film_cond=False,
         adaln_decouple=False,
         adaln_zone_temp=False,
+        surf_decoder_hidden_dim=0,
+        surf_decoder_pressure_only=False,
+        surf_decoder_self_attn=False,
     ):
         super().__init__()
         self.__name__ = "UniPDE_3D"
@@ -405,6 +408,10 @@ class Transolver(nn.Module):
         self.adaln_4cond = adaln_4cond
         self.film_cond = film_cond
         self.adaln_zone_temp = adaln_zone_temp
+        self.surf_decoder_hidden_dim = surf_decoder_hidden_dim
+        self.surf_decoder_pressure_only = surf_decoder_pressure_only
+        self.surf_decoder_self_attn = surf_decoder_self_attn
+        _use_surf_dec = surf_decoder_hidden_dim > 0
         if output_fields is None or output_dims is None:
             raise ValueError("output_fields and output_dims must be provided")
         if len(output_fields) != len(output_dims):
@@ -445,12 +452,12 @@ class Transolver(nn.Module):
                     mlp_ratio=mlp_ratio,
                     out_dim=out_dim,
                     slice_num=slice_num,
-                    last_layer=(idx == n_layers - 1),
+                    last_layer=(idx == n_layers - 1) and not _use_surf_dec,
                     linear_no_attention=linear_no_attention,
                     learned_kernel=learned_kernel,
-                    field_decoder=field_decoder if (idx == n_layers - 1) else False,
-                    adaln_output=adaln_output if (idx == n_layers - 1) else False,
-                    soft_moe=soft_moe if (idx == n_layers - 1) else False,
+                    field_decoder=field_decoder if (idx == n_layers - 1) and not _use_surf_dec else False,
+                    adaln_output=adaln_output if (idx == n_layers - 1) and not _use_surf_dec else False,
+                    soft_moe=soft_moe if (idx == n_layers - 1) and not _use_surf_dec else False,
                     adaln_all=adaln_all_blocks,
                     adaln_cond_dim=4 if adaln_4cond else 2,
                     adaln_zero_init=not adaln_nozero,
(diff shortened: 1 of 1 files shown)


tcapelle:
ADVISOR: R7 Surface Decoder results — **FAILED**.

All variants still running but vl=0.52-0.59, significantly worse than baseline 0.3997. The surface-specific attention decoder approach is not converging. Multiple crashes during training.

Closing and reassigning to R8.

PR #1879, opened 2026-03-28
Phase 5: Transolver++ Ada-Temp — Per-Point Adaptive Slice Temperature (8 parallel)

tcapelle:
## Hypothesis

Replace the global (per-head) slice temperature in the Transolver's physics attention with a **per-point adaptive temperature**. Currently, all ~100K nodes share the same temperature parameter for slice routing. Ada-Temp (from Transolver++, ICML 2025) learns a per-node temperature, allowing surface nodes with sharp pressure gradients to form sharper slice assignments (lower temperature) while far-field volume nodes remain more diffuse (higher temperature).

**Why this should work:**
- Transolver++ (ICML 2025) showed **46% error reduction** from Ada-Temp alone on aircraft pressure prediction
- The current model uses `self.temperature = nn.Parameter(torch.ones([1, heads, 1, 1]) * 0.5)` — a single scalar per head shared across ALL nodes
- Surface boundary layer nodes (where pressure gradients are sharpest) should route to slices more sharply — they need specialized physics states
- Far-field nodes should route diffusely — they share generic freestream physics
- This is the #2 ranked hypothesis from our research agent, with strong empirical backing from the Transolver++ paper

**Why this is different from previous work:**
- We tried zone_temp (PR #1838) which adds a per-sample offset based on (is_tandem, gap, Re) — this is per-NODE, much more fine-grained
- We tried tandem_temp_offset which adjusts temperature for tandem vs single samples — again per-sample, not per-node
- Ada-Temp is fundamentally different: each individual node gets its own temperature based on its local features

## Instructions

Modify the `Physics_Attention_Irregular_Mesh` class in `train.py` to support per-point adaptive temperature.

### Implementation:

In `Physics_Attention_Irregular_Mesh.__init__`, add:
```python
self.ada_temp = ada_temp  # new parameter
if ada_temp:
    # Per-point temperature projection: hidden features → 1 temp offset per head
    self.temp_proj = nn.Linear(dim_head, heads)
    nn.init.zeros_(self.temp_proj.weight)  # start as current model (no per-point variation)
    nn.init.zeros_(self.temp_proj.bias)
    self.temp_scale = nn.Parameter(torch.tensor(0.1))  # learnable scale for temp offset
```

In `Physics_Attention_Irregular_Mesh.forward`, replace the temperature computation:
```python
# BEFORE (current code):
temp = self.temperature  # [1, heads, 1, 1]

# AFTER (Ada-Temp):
if self.ada_temp:
[...]

PR #1910, opened 2026-03-28
Phase 5: Gumbel Implementation Investigation — Why Do Results Differ? (8 parallel)

tcapelle:
## Hypothesis

askeladd's Gumbel-Softmax (#1898) showed marginal improvement (val/loss 0.400 vs 0.403) while nezuko's implementation (#1904) showed Gumbel hurting (0.406 vs 0.397). The implementations may differ in critical details. This PR investigates why.

Possible differences:
1. Where in the forward pass the noise is injected
2. Whether noise is applied to all heads or just one
3. How the noise interacts with tandem_temp_offset and zone_temp
4. Whether compile_mode affects the random number generation

## Instructions

Run askeladd's exact gumbel implementation alongside nezuko's and a common baseline to identify what's different. Use both students' code on the same seeds.

| GPU | Config | Notes |
|-----|--------|-------|
| 0-2 | askeladd's gumbel (scale=0.10) | seeds 42-44 |
| 3-5 | nezuko's gumbel (scale=0.10, no ada_temp, no mem_opt) | seeds 42-44 |
| 6-7 | Baseline | seeds 42-43 |

## Baseline
val/loss 0.401, p_in 12.95, p_oodc 8.40, p_tan 33.8, p_re 24.7

### DIFF
--- a/cfd_tandemfoil/train.py
+++ b/cfd_tandemfoil/train.py
@@ -139,7 +139,8 @@ class Physics_Attention_Irregular_Mesh(nn.Module):
 
     def __init__(self, dim, heads=8, dim_head=64, dropout=0.0, slice_num=64,
                  linear_no_attention=False, learned_kernel=False,
-                 decouple_slice=False, zone_temp=False, prog_slices=False):
+                 decouple_slice=False, zone_temp=False, prog_slices=False,
+                 gumbel_softmax=False, gumbel_scale=0.1, gumbel_variant="v1"):
         super().__init__()
         inner_dim = dim_head * heads
         self.dim_head = dim_head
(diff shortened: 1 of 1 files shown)


tcapelle:
ADVISOR: ## Review — Gumbel Implementation Investigation

**Result: CLOSE — Mystery solved, null result confirmed.**

Excellent detective work, nezuko. The key findings:

1. **Both Gumbel implementations are mathematically equivalent** — v1 and v2 produce identical results (val/loss 0.404±0.002 vs 0.403±0.001)
2. **The original discrepancy was caused by confounding variables** (ada_temp + mem_opt), not the Gumbel formula
3. **Gumbel at scale=0.1 is neutral** — all three groups overlap within noise (val/loss ~0.403)

This definitively closes the Gumbel investigation. We now know:
- Gumbel-Softmax slice routing doesn't help at our scale
- Adaptive temperature doesn't help (confirmed in PR #1879 and now here)
- The Transolver's existing softmax routing is already well-calibrated

The clean experimental design (3 seeds per variant, controlled baseline, single-variable testing) made this conclusive. Well done.

PR #1915, opened 2026-03-28
Phase 5: Data Normalization Sweep — Alternative Physics Normalization (8 parallel)

tcapelle:
## Hypothesis

The current physics normalization (Cp-based: Ux/Umag, Uy/Umag, p/q) was one of the earliest design choices and hasn't been revisited since Phase 3. Alternative normalizations could improve the model's ability to learn:

1. **No physics norm** (raw_targets): Predict raw Ux, Uy, p directly with statistical normalization only
2. **Log-Cp pressure** (log_pressure): Already exists as a flag but was only tested briefly in Phase 3
3. **Per-domain normalization**: Different Umag/q for single-foil vs tandem
4. **Tighter clamps**: The current high_p_clamp=2.0 may be too loose

## Instructions

| GPU | Config | Notes |
|-----|--------|-------|
| 0 | raw_targets (no physics norm) | Skip Cp normalization entirely |
| 1 | log_pressure | Log-transform pressure channel |
| 2 | tight_denorm_clamps | Tighter clamps [-3,3] instead of [-10,10]/[-20,20] |
| 3 | unified_clamps | Same clamp for all channels |
| 4 | no_perstd | Remove per-sample std normalization |
| 5 | no_perstd_p | Remove per-sample std for pressure only |
| 6 | raw_targets seed 43 | Multi-seed for raw targets |
| 7 | Baseline | Control |

## Baseline
val/loss 0.401, p_in 12.95, p_oodc 8.40, p_tan 33.8, p_re 24.7


tcapelle:
ADVISOR: ## Review — Data Normalization Sweep

**Result: CLOSE — Current normalization is optimal. Physics normalization is critical.**

Key findings:
1. **Physics normalization (Cp) + per-sample std = critical** — removing either causes OOD-Re catastrophic failure (val_ood_re > 18000)
2. **log_pressure shows promise on p_oodc** (7.8 vs 8.2 baseline) but doesn't beat overall. Worth remembering.
3. **Baseline run scored val/loss=0.3959, p_tan=32.6** — among the best single runs we've seen, confirming seed variance.

The normalization pipeline is well-tuned and shouldn't be changed. The per-sample std normalization is the key to Reynolds number generalization — never remove it.

Good diagnostic work, fern. The finding that physics normalization is load-bearing for OOD-Re is important context for future experiments.

PR #1918, opened 2026-03-28
Phase 5: Test-Time Y-Flip Augmentation + Surface Pressure Weighting (8 parallel)

tcapelle:
## Hypothesis

Two orthogonal ideas combined:

### 1. Test-Time Y-Flip Augmentation (TTA)
Run each validation sample twice — normally and with y-coordinates flipped — then average predictions (with Uy sign-flipped for the augmented prediction). This exploits the physical symmetry of 2D Navier-Stokes: an airfoil at angle alpha produces fields closely related to the same airfoil at -alpha. By averaging, independent prediction errors cancel, reducing MAE for uncorrelated error components.

The y-flip augmentation exists in training but has NEVER been applied at test time across 1500+ experiments.

### 2. Surface Pressure Loss Weighting
Increase the relative weight of surface pressure in the loss function. Currently all 3 output channels (Ux, Uy, p) are weighted equally, but p_in/p_oodc/p_tan/p_re (the metrics we care about) are ALL pressure. By giving surface pressure 2-5x more weight, we directly optimize for what matters.

**References:** TTA is standard in Kaggle competition winners; Nguyen-Thanh et al. "Test-Time Augmentation for 3D Point Cloud Classification" (CVPR Workshop 2023)

## Instructions

See the first comment below for detailed implementation instructions.

## Baseline
val/loss 0.401±0.005, p_in 12.95±0.3, p_oodc 8.40±0.4, p_tan 33.8±0.5, p_re 24.7±0.2


tcapelle:
STUDENT thorfinn:

## Results

### Summary

1. **TTA y-flip does NOT work** — the model was trained with `--aug aoa_perturb` (not yflip), so it hasn't learned y-flip equivariance. Flipped inputs produce garbage predictions that corrupt the average.

2. **Surface pressure weighting (surf_p_weight) shows no improvement** — increasing pressure weight from 1x to 5x slightly degrades p_in while p_oodc/p_re remain within noise.

### Part 1: TTA Y-Flip — Failed

The TTA-enabled runs (both original and fixed versions) showed normal val/loss (~0.40) but absurdly high denormalized surface MAE (460-2851 vs baseline 12.5). Root cause:

- The baseline is trained with `--aug aoa_perturb` (±1° rotation), which does NOT include y-flip
- The model has never seen y-flipped inputs during training
- When given a y-flipped input at test time, the model produces incorrect predictions
- Averaging the correct (normal) prediction with the incorrect (flipped) prediction destroys accuracy

**Fix needed:** TTA y-flip would require training WITH y-flip augmentation (either `--aug yflip` or a combined augmentation that includes both rotation and flip). Without it, the model lacks the equivariance needed for TTA.

### Part 2: Surface Pressure Weighting — Null Result

| Config | val/loss | p_in | p_oodc | p_tan | p_re |
|--------|----------|------|--------|-------|------|
| **8-seed baseline** | **0.400** | **13.03** | **8.23** | **33.54** | **24.56** |
| Baseline control (s42) | 0.395 | 12.53 | 8.08 | 32.61 | 24.76 |
| surf_p_weight=2.0 | 0.402 | 12.81 | 8.32 | 33.91 | 24.66 |
| surf_p_weight=3.0 | 0.407 | 12.59 | 8.25 | 33.85 | 24.61 |
| surf_p_weight=5.0 | 0.416 | 13.27 | 8.11 | 33.85 | 24.54 |
| Baseline s43 | 0.405 | 12.83 | 8.48 | 34.41 | 24.84 |

**Observations:**
- p_weight=2.0: p_in +0.28 worse, p_re -0.10 better (both within 1σ noise)
- p_weight=3.0: p_in +0.06 worse, p_re -0.15 better (both within noise)  
- p_weight=5.0: p_in +0.74 worse, val/loss +0.021 worse (surface over-weighting hurts convergence)
- No pressure weight setting improved any metric beyond baseline noise (σ_p_in=0.37, σ_p_oodc=0.18)

The pressure channel is already well-represented in the loss through the adaptive surf_weight mechanism, hard-node mining (epoch >30), and the existing loss structure. Additional explicit pressure weighting doesn't help — it just shifts the optimization balance away from velocity, degrading overall convergence.

### Bug fix note

First TTA implementation incorrectly flipped feature indices [11, 14, 18] (dsdf component + AoA angles) in addition to the correct indices [1, 3, 5, 7, 9]. Fixed to match training y-flip augmentation convention. However, the fundamental issue is that the model wasn't trained with y-flip augmentation, so even the corrected TTA produces garbage.

### Code changes

1. `--tta_yflip`: Test-time y-flip augmentation — runs second forward pass with flipped input, negates Uy, averages predictions. Feature flip indices match training yflip convention: [1, 3, 5, 7, 9].
2. `--surf_p_weight`: Multiplier for surface pressure error in loss computation (applied in both regular and hard-node-mining paths).
3. Added full seed reproducibility (random.seed + numpy.random.seed).

### Commands used

```bash
cd cfd_tandemfoil && CUDA_VISIBLE_DEVICES=X python train.py \
  --agent thorfinn --wandb_name "thorfinn/<variant>" \
  --wandb_group phase5/tta-surfweight \
  --field_decoder --adaln_output --use_lion --lr 2e-4 \
  --aug aoa_perturb --aug_full_dsdf_rot --high_p_clamp \
  --n_layers 3 --slice_num 96 --tandem_ramp \
  --domain_layernorm --domain_velhead --ema_decay 0.999 \
  --weight_decay 5e-5 --cosine_T_max 180 --disable_pcgrad \
  --pressure_first --pressure_deep --seed 42 \
  [--tta_yflip] [--surf_p_weight N]
```

### W&B runs

- Baseline ctrl s42: `wa5k2xk4` | Baseline s43: `uxpx4f3m`
- surf_p_weight=2: `6cj7gp3a` | =3: `igwh5q2h` | =5: `9purd37z`
- TTA only (broken): `rw7cljei` → fixed: `df23blvg`
- TTA+pw2 (broken): `dksrw1w4` → fixed: `wpj37kyw`
- TTA+pw3 (broken): `ibrvv349` → fixed: `zj0858ci`

### Suggested follow-ups

1. **Train with combined augmentation** (`yflip + aoa_perturb`) and THEN test TTA — this should work since the model will have learned y-flip equivariance
2. **Don't increase surf_p_weight** — the existing loss balance is already well-tuned
3. **The baseline control (s42) result of p_in=12.53** is notably better than the 8-seed mean (13.03) — this is within the expected variance (2.8% CV) but suggests seed 42 is consistently good

PR #2115, opened 2026-04-04
Phase 6: Gap/Stagger Perturbation Augmentation — Tandem OOD Robustness

tcapelle:
## Hypothesis

The model sees tandem samples at fixed (gap, stagger) values from the training distribution. By adding small Gaussian noise to the gap and stagger features during training, we force the model to generalize across the local neighborhood of each tandem geometry — a domain randomization strategy directly targeting the p_tan OOD axis.

**Motivation:** The tandem validation split (`val_tandem_transfer`) evaluates on unseen aft-foil geometries at novel gap/stagger combinations. The model must interpolate between seen (gap, stagger) pairs. Feature-space domain randomization during training explicitly expands the model's effective training distribution along these axes, improving robustness to unseen tandem configurations.

**Literature support:**
- SIMSHIFT (2025): Feature-space perturbation of conditioning variables reduces OOD gap in CFD surrogate models
- Domain Randomization (Tobin et al. 2017): Adding noise to conditioning variables is one of the most reliable generalization tools in sim-to-real transfer
- Gap/stagger augmentation has improved tandem prediction metrics in comparable multi-element airfoil studies

**Why this is different from existing augmentations:**
- `aoa_perturb`: perturbs angle-of-attack via full geometric rotation of mesh coordinates
- Gap/stagger perturbation: perturbs scalar conditioning features only (2 scalars), leaving all mesh coordinates unchanged
- Only affects tandem samples — single-foil sentinel (gap=0, stagger=0) is preserved exactly

**Key constraint:** Gap and stagger are broadcast to ALL nodes of a sample. The perturbation must be per-sample, applied to both the feature tensor and broadcast correctly across all N nodes.

### Feature layout verification

From `data/prepare_multi.py`, the 24-dim input `x` layout (0-indexed):
```
x[:, :,  0: 2] = pos (x, y coordinates)
x[:, :,  2: 4] = saf
x[:, :,  4:12] = dsdf
x[:, :, 12]    = is_surface
x[:, :, 13]    = log_Re
x[:, :, 14]    = AoA0_rad
x[:, :, 15:18] = NACA0 (3 values)
x[:, :, 18]    = AoA1_rad
x[:, :, 19:22] = NACA1 (3 values)
x[:, :, 22]    = gap      ← TARGET
x[:, :, 23]    = stagger  ← TARGET
```

[...]

PR #2104, opened 2026-04-04
Phase 6: Dedicated Aft-Foil Surface Refinement Head (ID=7)

tcapelle:
## Hypothesis

The current `SurfaceRefinementHead` (srf_head) applies the same correction MLP to ALL surface nodes equally — boundary IDs 5 (single-foil pressure), 6 (suction/fore-foil tandem), and 7 (aft foil in tandem). But boundary ID=7 is qualitatively different: the aft foil operates in the wake of the fore foil, where pressure distributions depend heavily on the **gap** (x[:,0,21]) and **stagger** (x[:,0,22]) between foils.

Sharing srf_head weights across all three boundary types creates a capacity bottleneck specifically for the hardest case. A **dedicated second SRF head** exclusively for ID=7 nodes, optionally conditioned via FiLM on gap/stagger, should give the model the capacity to learn richer aft-foil corrections — directly attacking p_tan (our weakest metric at 29.1, 2.5x worse than p_in=12.1).

**Key:** The new head uses zero-init on its last linear layer, so worst-case it is identical to the current baseline. No regression risk on p_in or p_oodc (single-foil and OOD samples have no ID=7 nodes).

**Prior work:** PR #1893 upweighted ID=7 nodes in the loss (marginal, still one head). PR #1941 split surface vs. volume (different axis). No prior experiment has tried a dedicated MLP exclusively for aft-foil surface nodes.

## Instructions

Add a new flag `--aft_foil_srf` to create a second `SurfaceRefinementHead` for boundary ID=7 nodes only, and `--aft_foil_srf_film` to add FiLM conditioning on gap/stagger.

### Step 1 — Add config flags

In the `Config` dataclass, add:
```python
aft_foil_srf: bool = False
aft_foil_srf_film: bool = False
aft_foil_srf_hidden: int = 192
aft_foil_srf_layers: int = 3
```

### Step 2 — Init the new head in `Transolver.__init__`

```python
if cfg.aft_foil_srf:
    self.aft_srf_head = SurfaceRefinementHead(
        n_hidden=cfg.n_hidden,
        hidden_size=cfg.aft_foil_srf_hidden,
        n_layers=cfg.aft_foil_srf_layers,
        out_dim=3,
    )
    if cfg.aft_foil_srf_film:
        # FiLM scale and shift projections (zero-init => identity at start)
        self.aft_srf_film_scale = nn.Linear(2, cfg.aft_foil_srf_hidden, bias=False)
[...]

PR #2128, opened 2026-04-04
Phase 6: Reynolds-Conditional SRF — FiLM on (Re, AoA) for Surface Refinement Head

tcapelle:
## Hypothesis

The shared `surface_refine` head (`--surface_refine`, merged Phase 5, the single biggest win at -72.7% p_re) currently applies the same MLP correction regardless of flow regime. It has no awareness of Reynolds number or angle of attack — the same weights correct predictions at Re=1e5 and Re=4e6, even though the error character differs dramatically:

- **Low Re:** boundary layer separation → large Cp plateaus, smooth gradients
- **High Re:** attached flow → sharp leading-edge gradients, pressure recovery

**Hypothesis:** Conditioning the SRF head on (Re, AoA) via FiLM modulation allows the correction MLP to adapt its strategy to the flow regime, directly targeting `p_re` and `p_oodc`.

**Why this is different from the dead-end `aft_foil_srf_film` (+41.6% p_oodc regression, #2104):** That FiLM used configurational features (gap, stagger) that vary per-sample dynamically. Re and AoA are the *primary flow condition descriptors* — they are already used by AdaLN in every Transolver block. FiLM on Re/AoA for the SRF head is the same mechanism as AdaLN, applied to the post-processing correction step.

**Expected impact:** -2 to -4% p_re, -1 to -2% p_oodc. ~18 LoC change. No memory overhead.

### Step 1: Locate the key classes

```bash
grep -n "class SurfaceRefinementHead\|class AftFoilRefinementHead\|film" cfd_tandemfoil/train.py | head -30
```

Find `SurfaceRefinementHead.__init__` and `SurfaceRefinementHead.forward`. Also look at `AftFoilRefinementHead` — it already has `film=True` FiLM support; mirror that exact pattern.

### Step 2: Add FiLM to SurfaceRefinementHead

In `SurfaceRefinementHead.__init__`, add optional `film: bool = False` parameter and FiLM layers:

```python
self.film = film
if film:
    self.film_scale = nn.Linear(2, hidden_dim, bias=False)
    self.film_shift = nn.Linear(2, hidden_dim)
    nn.init.zeros_(self.film_scale.weight)
    nn.init.zeros_(self.film_shift.weight)
    nn.init.zeros_(self.film_shift.bias)
```

In `SurfaceRefinementHead.forward`, add a `cond` argument and apply FiLM after the first hidden activation:

```python
def forward(self, h_surf, base_surf, cond=None):
    x = torch.cat([h_surf, base_surf], dim=-1)
    x = self.act(self.fc1(x))
    if self.film and cond is not None:
        scale = self.film_scale(cond)
        shift = self.film_shift(cond)
        x = x * (1 + scale) + shift
    x = self.fc2(x)
    return x
```

### Step 3: Verify the Re and AoA feature indices

```bash
grep -n "adaln\|cond_dim\|cond_feat\|cond\[:, 0\]\|x\[:, 0,\|condition" cfd_tandemfoil/train.py | head -40
```

The condition input should be `x[:, 0, 13:15]` — Re and AoA as the first 2 scalar condition features (same as what AdaLN uses). Verify by printing at training start:

```python
print(f"[DEBUG] cond sample: feat13={x[0,0,13]:.4f}, feat14={x[0,0,14]:.4f}")
```

These should be normalized values in roughly [-3, 3]. If they look like zeros, find the right indices via `prepare_multi.py` or the stats dict.

### Step 4: Pass condition to the SRF head

Find where `refine_head(` is called. Add condition extraction and pass it:

```python
# Extract per-sample condition [B, 2]
cond_global = x[:, 0, 13:15]

# Expand to surface nodes using surf_idx (surf_idx[:, 0] = batch index per node)
cond_surf = cond_global[surf_idx[:, 0]]   # [N_surf_nodes, 2]

# Call with condition:
surf_pred = refine_head(h_surf, base_surf, cond=cond_surf)
```

### Step 5: Add config flags

```python
# argparse:
parser.add_argument('--surface_refine_film', action='store_true')
# Config dataclass:
surface_refine_film: bool = False
```

Pass `film=cfg.surface_refine_film` when constructing `SurfaceRefinementHead`.

### Step 6: Run experiments

Run 2 film seeds + 2 control seeds (identical config, no `--surface_refine_film`) for a clean comparison:

```bash
# Film, seed 42
cd cfd_tandemfoil && python train.py --agent edward \
  --wandb_name "edward/re-srf-film-s42" \
  --wandb_group phase6/re-conditional-srf \
  --surface_refine --surface_refine_film --seed 42 \
  --cosine_T_max 160 --disable_pcgrad --pressure_first --pressure_deep \
  --residual_prediction --surface_refine_hidden 192 --surface_refine_layers 3 \
  --aft_foil_srf --aug_gap_stagger_sigma 0.02

# Film, seed 73 — same + --seed 73
# Control, seed 42 — same without --surface_refine_film
# Control, seed 73 — same without --surface_refine_film + --seed 73
```

**Note:** Use `--surface_refine` (not `--surface_refine_hidden`/`--surface_refine_layers` alone) — the `--surface_refine` flag enables the head.

**W&B group:** `phase6/re-conditional-srf`

Report surface MAE metrics (p_in, p_oodc, p_tan, p_re) for all 4 runs. If both film seeds beat both control seeds on p_re, strong candidate for 8-seed validation.

## Baseline

Current single-model baseline (PR #2104 + #2115, +aft_foil_srf +aug_gap_stagger_sigma=0.02, 8-seed mean, seeds 42-49):

| Metric | 8-seed mean | Target to beat |
|--------|-------------|----------------|
| p_in   | **13.19 ± 0.33** | < 13.19 |
| p_oodc | **7.92 ± 0.17**  | < 7.92  |
| p_tan  | **30.05 ± 0.36** | < 30.05 |
| p_re   | **6.45 ± 0.07**  | **< 6.45** ← primary target |

**Reproduce baseline:**
```bash
  --wandb_name "edward/baseline-s42" \
  --cosine_T_max 160 --disable_pcgrad --pressure_first --pressure_deep \
  --aft_foil_srf --aug_gap_stagger_sigma 0.02
```

### DIFF
--- a/cfd_tandemfoil/train.py
+++ b/cfd_tandemfoil/train.py
@@ -454,12 +454,15 @@ class SurfaceRefinementHead(nn.Module):
     Takes the Transolver hidden features (and optionally the base predictions)
     for surface nodes only, and outputs a residual correction that is added
     to the main model's predictions at surface locations.
+
+    Optionally applies FiLM conditioning on (Re, AoA) after the first hidden layer.
     """
 
     def __init__(self, n_hidden: int, out_dim: int, hidden_dim: int = 128,
-                 n_layers: int = 2, p_only: bool = False):
+                 n_layers: int = 2, p_only: bool = False, film: bool = False):
         super().__init__()
         self.p_only = p_only
+        self.film = film
         actual_out = 1 if p_only else out_dim  # 1 for pressure-only, 3 for all fields
         layers: list[nn.Module] = []
         # Input: hidden features (n_hidden) + base predictions (out_dim)
@@ -473,17 +476,34 @@ class SurfaceRefinementHead(nn.Module):
         nn.init.zeros_(layers[-1].weight)
         nn.init.zeros_(layers[-1].bias)
         self.mlp = nn.Sequential(*layers)
+        # FiLM modulation from (Re, AoA) — 2-dim condition, zero-initialized
+        if film:
+            self.film_scale = nn.Linear(2, hidden_dim, bias=False)
+            self.film_shift = nn.Linear(2, hidden_dim)
+            nn.init.zeros_(self.film_scale.weight)
+            nn.init.zeros_(self.film_shift.weight)
+            nn.init.zeros_(self.film_shift.bias)
 
-    def forward(self, hidden: torch.Tensor, base_pred: torch.Tensor) -> torch.Tensor:
+    def forward(self, hidden: torch.Tensor, base_pred: torch.Tensor,
+                cond: torch.Tensor | None = None) -> torch.Tensor:
         """
         Args:
             hidden: [M, n_hidden] — hidden features for surface nodes only
             base_pred: [M, out_dim] — base predictions for surface nodes only
+            cond: [M, 2] — per-node (Re, AoA) condition (only used if film=True)
         Returns:
             correction: [M, out_dim] — additive correction (zero-padded for p_only)
         """
         inp = torch.cat([hidden, base_pred], dim=-1)
-        correction = self.mlp(inp)
+        # Run through layers, applying FiLM after first hidden activation (index 2)
+        x = inp
+        for i, layer in enumerate(self.mlp):
+            x = layer(x)
+            if self.film and cond is not None and i == 2:
+                gamma = self.film_scale(cond)   # [M, hidden_dim]
+                beta = self.film_shift(cond)    # [M, hidden_dim]
+                x = x * (1.0 + gamma) + beta
+        correction = x
         if self.p_only:
             # Pad with zeros for velocity channels: [M, 1] → [M, 3]
             zeros = torch.zeros(correction.shape[0], base_pred.shape[-1] - 1,
@@ -995,6 +1015,7 @@ class Config:
     surface_refine_layers: int = 2            # number of hidden layers in refinement MLP
     surface_refine_p_only: bool = False       # only refine pressure channel (not velocity)
     surface_refine_context: bool = False      # use surface + nearest-volume context features
+    surface_refine_film: bool = False         # FiLM conditioning on (Re, AoA) for SRF head
     # Phase 6: Asinh pressure transform
     asinh_pressure: bool = False             # transform pressure targets with asinh for dynamic range compression
     asinh_scale: float = 1.0                 # scale factor before asinh: asinh(p * scale)
@@ -1190,6 +1211,7 @@ if cfg.surface_refine:
             hidden_dim=cfg.surface_refine_hidden,
             n_layers=cfg.surface_refine_layers,
             p_only=cfg.surface_refine_p_only,
+            film=cfg.surface_refine_film,
         ).to(device)
     refine_head = torch.compile(refine_head, mode=cfg.compile_mode)
     _refine_n_params = sum(p.numel() for p in refine_head.parameters())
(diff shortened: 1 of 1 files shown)


tcapelle:
## Experiment Results: Reynolds-Conditional SRF (FiLM)

### Setup
- 4 runs: 2 FiLM + 2 control, seeds 42 and 73
- All runs: `--ema_start_epoch 100 --ema_decay 0.999 --cosine_T_max 160 --aft_foil_srf --aug_gap_stagger_sigma 0.02 --asinh_pressure --asinh_scale 0.75 --field_decoder --adaln_output --use_lion --lr 2e-4 --surface_refine_hidden 192 --surface_refine_layers 3 --residual_prediction`
- FiLM runs add: `--surface_refine_film` (FiLM conditioning on (Re, AoA) in SurfaceRefinementHead)
- Training cut at 180-min timeout. Control reached epoch 151, FiLM reached epoch 149.
- EMA active from epoch 100 → ~51 epochs (ctrl) / ~49 epochs (film) of EMA accumulation.

### Results — Surface MAE (Pressure)

| Run | p_in | p_oodc | p_tan | p_re | W&B Run ID |
|-----|------|--------|-------|------|------------|
| **Control s42** | 12.8 | 7.8 | 29.7 | 6.8 | `np3anmah` |
| **Control s73** | 14.2 | 7.3 | 29.9 | 6.5 | `fcri663y` |
| **Control avg** | **13.5** | **7.55** | **29.8** | **6.65** | — |
| **FiLM s42** | 13.3 | 8.0 | 30.6 | 6.6 | `7bnlke5o` |
| **FiLM s73** | 12.7 | 7.7 | 29.7 | 6.5 | `yh5ixlx8` |
| **FiLM avg** | **13.0** | **7.85** | **30.15** | **6.55** | — |
| **Baseline (8-seed)** | **13.19** | **7.92** | **30.05** | **6.45** | — |

### FiLM vs Control (2-seed averages)

| Metric | FiLM | Control | Delta | Interpretation |
|--------|------|---------|-------|----------------|
| p_in | 13.0 | 13.5 | -3.7% | FiLM slightly better |
| p_oodc | 7.85 | 7.55 | +4.0% | FiLM slightly worse |
| p_tan | 30.15 | 29.8 | +1.2% | FiLM slightly worse |
| p_re | 6.55 | 6.65 | -1.5% | FiLM slightly better |

### Analysis

**Null result.** FiLM conditioning on (Re, AoA) for the Surface Refinement Head shows no meaningful improvement over the control. The differences between FiLM and control are within seed noise — the single-seed variation (ctrl-s42 p_in=12.8 vs ctrl-s73 p_in=14.2 is a 10% swing) dwarfs the FiLM effect.

**Why it likely doesn't help:**
1. The SurfaceRefinementHead already receives (Re, AoA) implicitly through the hidden states from the Transolver backbone, which uses AdaLN conditioning on these same features at every block.
2. The refinement head's correction is a small residual — the heavy lifting is done by the backbone which already adapts to Re/AoA via AdaLN.
3. The FiLM layer only modulates the first hidden activation (after LayerNorm+GELU), so its effect is limited to a single layer of a 3-layer MLP.
4. FiLM adds only 960 extra parameters — too few to capture meaningful Re-dependent correction patterns beyond what the backbone already provides.

**Note on control runs:** Both control averages are within normal baseline variation. The `--ema_start_epoch 100` (vs baseline's default) may explain minor deviations.

### Suggested Follow-ups
1. **Skip FiLM for SRF.** The backbone's AdaLN already handles Re/AoA conditioning well enough. Adding explicit conditioning at the SRF level is redundant.
2. **Focus on loss formulation or data representation** rather than more conditioning mechanisms — the model's capacity to condition on flow regime is already well-served by AdaLN.
3. **Consider flow topology-aware refinement** — instead of conditioning on scalar Re/AoA, condition on local flow features (e.g., pressure gradient magnitude, boundary layer indicators) that vary spatially and aren't already captured by the global conditioning.

tcapelle:
ADVISOR: Closing PR #2128 because Null result — W&B-verified FiLM 2-seed avg: p_in=13.03, p_oodc=7.84, p_tan=30.15, p_re=6.56. All metrics at or worse than baseline (p_oodc +2.4%, p_tan +0.1%, p_re +0.8%). FiLM is also worse than its own control on p_oodc (7.84 vs 7.55) and p_tan (30.15 vs 29.78).

Student's analysis is correct: the SRF head already receives Re/AoA conditioning implicitly through the Transolver backbone's AdaLN, so adding explicit FiLM modulation is redundant. The 960-parameter FiLM layer cannot learn anything new that AdaLN hasn't already encoded.

Good experiment design and thorough analysis. The follow-up suggestion (local flow feature conditioning rather than global scalars) is an interesting direction — something like conditioning on local pressure gradient magnitude or boundary layer thickness proxies rather than global (Re, AoA) might circumvent the redundancy problem..

PR #2127, opened 2026-04-04
Phase 6: Context-Aware AftSRF — KNN Volume Context for Wake Pressure

tcapelle:
## Hypothesis

The `--aft_foil_srf` head (PR #2104, merged, -0.8% p_tan) corrects aft-foil surface predictions using only the surface node's own hidden representation. But aft-foil pressure is strongly influenced by the wake flow arriving from the fore-foil — a **non-local** physical dependency. The correction head has no access to what the volume nodes immediately upstream of the aft-foil look like.

By augmenting each aft-foil surface node's hidden representation with the **averaged hidden states of its K=8 nearest volume neighbors** (restricted to the aft-foil refinement zone, zone_id=2), we give the correction head direct access to wake-state information. This is physically motivated: the pressure recovery on the aft-foil suction surface depends on the velocity deficit and turbulence level arriving from the fore-foil wake — information carried in the volume nodes just upstream.

The existing `SurfaceRefinementContextHead` class already implements KNN-context refinement for all surface nodes globally. This experiment applies the same mechanism targeted specifically to aft-foil nodes, keeping the global SRF head unchanged.

### Step 1: Understand the existing architecture

Read the `AftFoilRefinementHead` class and its usage in `train.py`. Note:
- Input: `aft_hidden` [N_aft, n_hidden], `aft_base_pred` [N_aft, out_dim]
- Output: correction delta [N_aft, out_dim]
- The in_features of the MLP is `n_hidden * 2 + out_dim` (hidden + base_pred concatenated twice)

Also check if `SurfaceRefinementContextHead` exists and read its implementation — it already does KNN context gathering and you can borrow the logic directly.

### Step 2: Modify AftFoilRefinementHead to accept volume context

Create a new variant `AftFoilRefinementContextHead` (or add a `use_context` flag to the existing head):

```python
class AftFoilRefinementContextHead(nn.Module):
    def __init__(self, n_hidden, out_dim, hidden_size=192, n_layers=3, k_neighbors=8):
        super().__init__()
        self.k = k_neighbors
        # Input: aft_hidden + vol_context_avg + base_pred
        in_features = n_hidden + n_hidden + out_dim  # (surface hidden + volume context avg + base pred)
        layers = [nn.Linear(in_features, hidden_size), nn.GELU()]
        for _ in range(n_layers - 2):
            layers += [nn.Linear(hidden_size, hidden_size), nn.GELU()]
        layers.append(nn.Linear(hidden_size, out_dim))
[...]

PR #2097, opened 2026-04-04
Phase 6: Multi-Scale Deep Supervision — Auxiliary Loss on Intermediate Features

tcapelle:
## Hypothesis

The model's forward pass computes `fx_deep = fx` just before the last TransolverBlock (saved as `model_out["hidden"]`). Adding an auxiliary head that predicts surface targets from these intermediate features creates a deep supervision signal. This forces earlier blocks to produce more useful intermediate representations, reducing vanishing gradient effects in the 3-block Transolver.

Deep supervision is proven effective in nnUNet (Isensee et al.), HED (edge detection), and FNO variants. The mechanism is: direct gradient flow from surface targets to pre-final-block features, improving the physics attention in intermediate blocks.

### Code changes (~15 lines)

1. Add an auxiliary head after the model is created:
```python
# After model creation, add auxiliary prediction head
aux_head = nn.Sequential(
    nn.Linear(cfg.n_hidden, cfg.n_hidden // 2),
    nn.GELU(),
    nn.Linear(cfg.n_hidden // 2, 3)  # predict [Ux, Uy, p]
).to(device)
```

2. Add `aux_head` parameters to the optimizer:
```python
base_opt.add_param_group({'params': list(aux_head.parameters()), 'lr': _base_lr})
```

3. In the training loop, after the main forward pass, compute auxiliary loss:
```python
if cfg.deep_supervision and model.training:
    hidden = model_out["hidden"]  # [B, N, n_hidden] — intermediate features
    aux_pred = aux_head(hidden)   # [B, N, 3]
    # Apply surface mask
    aux_surf_pred = aux_pred[surface_mask]  # surface nodes only
    aux_surf_target = targets[surface_mask]
    aux_loss = cfg.aux_loss_weight * F.l1_loss(aux_surf_pred, aux_surf_target)
    loss = loss + aux_loss
    wandb.log({"train/aux_loss": aux_loss.item()})
```

4. New config flags:
```python
deep_supervision: bool = False
aux_loss_weight: float = 0.1  # weight for auxiliary loss
```

5. **Important:** The auxiliary head is NOT used at inference time. Only the main model predictions are used for evaluation.

### Experiment matrix (8 GPUs)

| GPU | Config | Seed | Notes |
|-----|--------|------|-------|
| 0 | Baseline (no aux loss) | 42 | Control |
| 1 | Baseline (no aux loss) | 43 | Control |
| 2 | `--deep_supervision --aux_loss_weight 0.05` | 42 | Light aux |
[...]

PR #2134, opened 2026-04-04
Phase 6: Fore-Foil TE Relative Coords — Inter-Foil Jet Frame for AftSRF Context

tcapelle:
## Hypothesis

The KNN volume context head (PR #2127, just merged) gives the aft-foil SRF head access to *what the wake looks like* upstream. But the aft-foil nodes' input positions are in the **global mesh frame** — the head has no explicit sense of *where the aft-foil sits relative to the fore-foil*. This matters: a node at (x=1.6, y=0.0) has a fundamentally different aerodynamic role depending on whether the fore-foil trailing edge is at x=1.0 (gap=0.6) or x=1.4 (gap=0.2).

**Hypothesis:** Adding 2 extra features to the `AftFoilRefinementContextHead` input — (x_rel, y_rel) = (x_node − x_te_fore, y_node − y_te_fore) — provides the inter-foil jet coordinate frame that physically governs wake-surface interaction. KNN context tells the head *what* the wake carries; relative coords tell the head *where in the wake* each surface node sits. These two pieces of information are complementary and together should unlock stronger aft-foil pressure predictions.

**Physical motivation:** Aft-foil Cp recovery depends strongly on the node's position relative to the inter-foil jet stagnation line. A node that sits directly in the jet core vs. at the jet boundary experiences very different pressure recovery, even if the upstream hidden states look similar. Relative coordinates make this geometry explicit.

**Expected impact:** p_tan −3% to −7%. Synergistic with `--aft_foil_srf_context`.

### Step 0: Understand the existing AftFoilRefinementContextHead

Read `AftFoilRefinementContextHead.__init__` and `.forward` in `train.py`. It currently:
- Takes `aft_hidden` [N_aft, n_hidden], `aft_positions` [N_aft, 2], `vol_hidden` [N_vol, n_hidden], `vol_positions` [N_vol, 2], `aft_base_pred` [N_aft, out_dim]
- Computes KNN context from vol_hidden
- Concatenates `[aft_hidden, vol_context, aft_base_pred]` → MLP

Find the `in_features` line in `__init__` — it is `n_hidden + n_hidden + out_dim`.

### Step 1: Add flag and modify the head

```python
# argparse:
parser.add_argument('--aft_foil_srf_rel_coords', action='store_true')
# Config dataclass:
aft_foil_srf_rel_coords: bool = False
```

Modify `AftFoilRefinementContextHead.__init__`:
```python
self.use_rel_coords = use_rel_coords
[...]

PR #2135, opened 2026-04-04
Phase 6: Tandem Self-Distillation — EMA Teacher for Tandem Samples

tcapelle:
## Hypothesis

The EMA model is already being maintained throughout training and is the checkpoint used for evaluation. It is a smoother, higher-quality version of the online model. Currently it is used **only at evaluation time**. But since it lives on-device and is updated every step, we can also use it as a **soft teacher** during training.

**Hypothesis:** After EMA begins accumulating (epoch ≥ `ema_start_epoch`), apply a self-distillation KD loss on **tandem samples only**: the EMA model's predictions on tandem nodes serve as soft targets that the online model must also predict. This gives the online model a second, smoother training signal on the hardest OOD split — val_tandem_transfer — at near-zero additional cost (EMA is already computed; we just run a second forward pass on the tandem subset with no_grad).

**Why tandem-only?** Tandem samples are the primary bottleneck (p_tan is our hardest metric). The EMA model has already seen more data per parameter update, is more stable on these samples, and provides a denoised supervision signal. KD on single-foil samples would have lower marginal value since the model already generalizes well there.

**Why this is different from auxiliary losses (dead end #2097):** Deep supervision added auxiliary prediction heads at intermediate layers — a gradient signal that interfered with learning. Self-distillation here is output-level only, after the full forward pass, and uses a soft/smooth target (EMA output) rather than ground truth. The EMA target smooths over noise in the ground truth labels for tandem samples.

**Expected impact:** p_tan −2% to −5%. ~25 LoC. Fully orthogonal to coordinate-based and augmentation changes.

### Step 1: Add config flags

```python
# argparse:
parser.add_argument('--tandem_selfdistill', action='store_true')
parser.add_argument('--tandem_selfdistill_weight', type=float, default=0.1)

# Config dataclass:
tandem_selfdistill: bool = False
tandem_selfdistill_weight: float = 0.1
```

### Step 2: Understand the EMA model in the training loop

Find where the EMA model is defined and updated. Search for:
```bash
grep -n "ema\|EMA\|_base_model\|ema_model" cfd_tandemfoil/train.py | head -40
```

The EMA model is `ema_model` (or similar). It is updated each step after the optimizer step, and used during validation. Note `ema_start_epoch` — EMA only accumulates after this epoch (default 100 in current baseline; may be different in train.py default, check with `grep ema_start_epoch train.py`).

### Step 3: Add the tandem self-distillation loss

In the training loop, AFTER computing the standard loss and AFTER the EMA model update, add:

```python
# Self-distillation KD loss on tandem samples
if cfg.tandem_selfdistill and epoch >= cfg.ema_start_epoch:
    # Identify tandem samples in the batch
    # tandem_mask is already computed earlier in the loop — reuse it
    # tandem_mask: [B] boolean, True for tandem samples
    
    if tandem_mask.any():
        # Get EMA predictions on tandem nodes (no_grad, no gradient through EMA)
        with torch.no_grad():
            ema_out = ema_model(x[tandem_mask], ...)   # same forward args as online model
            # NOTE: ema_model takes the same input signature — copy the forward call
            # from the online model forward pass, substituting ema_model
        
        # Online predictions on tandem nodes (already computed as `out`)
        online_tandem = out[tandem_mask]   # [N_tan, out_dim] — adjust for your tensor shape
        ema_tandem = ema_out               # [N_tan, out_dim]
        
        # KD loss: MSE between online and EMA outputs (on surface nodes only for efficiency)
        # Surface nodes only: use the same surf_mask you apply for surface MAE computation
        kd_loss = F.mse_loss(online_tandem, ema_tandem.detach())
        loss = loss + cfg.tandem_selfdistill_weight * kd_loss
```

**Key implementation detail:** The forward pass signature for `ema_model` is identical to the online model. You do NOT need to recompute anything — just run `ema_model(x[tandem_mask], ...)` with the same arguments used for the online model. Use `torch.no_grad()` to avoid gradients through the EMA model.

**Important:** The KD loss must be added BEFORE `loss.backward()`. If the EMA model update happens after `loss.backward()` in your training loop, compute the EMA forward pass BEFORE the optimizer step but AFTER the online forward pass. Check the exact order in the training loop with:
```bash
grep -n "ema\|backward\|optimizer.step\|loss" cfd_tandemfoil/train.py | grep -A2 -B2 "backward"
```

**Memory:** The EMA model forward pass on tandem-only samples (typically 30-50% of batch) costs ~1/3 extra compute. If you see OOM, filter to surface nodes only before computing KD loss, or reduce batch size by 10%.

### Step 4: Sweep over distillation weight

Run 3 seeds: w=0.05, w=0.10, w=0.20:

```bash
# w=0.10, seed 42 (primary)
cd cfd_tandemfoil && python train.py --agent edward \
  --wandb_name "edward/tandem-selfdistill-w0.10-s42" \
  --wandb_group phase6/tandem-selfdistill --seed 42 \
  --tandem_selfdistill --tandem_selfdistill_weight 0.10 \
  --cosine_T_max 160 --disable_pcgrad --pressure_first --pressure_deep \
  --aft_foil_srf --aug_gap_stagger_sigma 0.02 --aft_foil_srf_context

# w=0.05, seed 42: same but --tandem_selfdistill_weight 0.05
# w=0.20, seed 42: same but --tandem_selfdistill_weight 0.20

# Control seed 42: same WITHOUT --tandem_selfdistill
```

**W&B group:** `phase6/tandem-selfdistill`

Report p_in, p_oodc, p_tan, p_re + W&B run IDs for all 4 runs. If w=0.10 seed 42 shows clear improvement, run a second seed (73) on the best weight for 2-seed averaging.

**Watch for:** If p_tan degrades monotonically as w increases, the KD loss is over-constraining the online model. If p_oodc degrades with no p_tan improvement, the KD loss is hurting generalization elsewhere — try restricting to surface pressure nodes only (not velocity/pressure volume nodes).

## Baseline

Current single-model baseline (PR #2127, AftSRF KNN context K=8, 2-seed mean):

| Metric | 2-seed mean | Target |
|--------|-------------|--------|
| p_in | **13.02** | < 13.02 |
| p_oodc | **7.62** | < 7.62 |
| **p_tan** | **29.91** | **< 29.91** ← primary |
| p_re | **6.47** | < 6.47 |

W&B: `zosxwjmm` (s42), `twilqf1x` (s73)

**Reproduce baseline:**
```bash
  --wandb_name "edward/baseline-aft-srf-ctx" \
  --cosine_T_max 160 --disable_pcgrad --pressure_first --pressure_deep \
  --aft_foil_srf --aug_gap_stagger_sigma 0.02 --aft_foil_srf_context
```

### DIFF
--- a/cfd_tandemfoil/train.py
+++ b/cfd_tandemfoil/train.py
@@ -1052,6 +1052,8 @@ class Config:
     aft_foil_srf_hidden: int = 192           # hidden dim for aft-foil refinement head
     aft_foil_srf_layers: int = 3             # number of hidden layers for aft-foil refinement head
     aft_foil_srf_context: bool = False       # KNN volume context for aft-foil SRF (wake pressure recovery)
+    tandem_selfdistill: bool = False         # EMA self-distillation on tandem samples
+    tandem_selfdistill_weight: float = 0.1   # KD loss weight
 
 
 cfg = sp.parse(Config)
@@ -1944,6 +1946,22 @@ for epoch in range(MAX_EPOCHS):
             rdrop_loss = ((pred - rdrop_pred) ** 2 * valid_mask).sum() / valid_mask.sum().clamp(min=1)
             loss = loss + cfg.rdrop_alpha * rdrop_loss
 
+        # Tandem self-distillation: use EMA model as soft teacher on tandem samples
+        kd_loss = torch.tensor(0.0, device=device)
+        if cfg.tandem_selfdistill and ema_model is not None and is_tandem_batch.any():
+            with torch.no_grad():
+                with torch.amp.autocast("cuda", dtype=torch.bfloat16):
+                    ema_kd_out = ema_model({"x": x[is_tandem_batch]})
+                ema_kd_pred = ema_kd_out["preds"].float()
+            # Online backbone predictions for tandem samples (with grad)
+            online_kd_pred = out["preds"][is_tandem_batch].float()
+            # KD loss: MSE on surface pressure (channel 2) for tandem surface nodes
+            kd_surf_mask = (mask & is_surface)[is_tandem_batch]  # [N_tan, N_nodes]
+            n_surf_kd = kd_surf_mask.sum().clamp(min=1)
+            kd_diff_sq = (online_kd_pred[:, :, 2:3] - ema_kd_pred[:, :, 2:3]) ** 2
+            kd_loss = (kd_diff_sq * kd_surf_mask.unsqueeze(-1).float()).sum() / n_surf_kd
+            loss = loss + cfg.tandem_selfdistill_weight * kd_loss
+
         # PCGrad: in-dist (Group A) vs all-OOD (Group B) gradient projection
         # Group B = tandem + extreme-Re (>1σ) + extreme-AoA (>1σ), Group A = rest
         is_ood_pcgrad = is_tandem_batch | (x[:, 0, 13] > 1.0) | (x[:, 0, 14].abs() > 1.0)
@@ -2060,7 +2078,10 @@ for epoch in range(MAX_EPOCHS):
                         for ep, mp in zip(ema_aft_srf_head.parameters(), _ctx_base.parameters()):
                             ep.data.mul_(cfg.ema_decay).add_(mp.data, alpha=1 - cfg.ema_decay)
         global_step += 1
-        wandb.log({"train/loss": loss.item(), "train/surf_weight": surf_weight, "global_step": global_step})
+        _train_log = {"train/loss": loss.item(), "train/surf_weight": surf_weight, "global_step": global_step}
+        if cfg.tandem_selfdistill:
+            _train_log["train/kd_loss"] = kd_loss.item()
+        wandb.log(_train_log)
 
         epoch_vol += vol_loss.item()
         epoch_surf += surf_loss.item()


tcapelle:
## Experiment Results: Tandem Self-Distillation (EMA Teacher)

### Setup
- 4 runs: 3 KD weights (w=0.05, 0.10, 0.20) + 1 control, all seed 42
- All runs: `--ema_start_epoch 100 --ema_decay 0.999 --cosine_T_max 160 --aft_foil_srf --aug_gap_stagger_sigma 0.02 --aft_foil_srf_context --asinh_pressure --asinh_scale 0.75 --field_decoder --adaln_output --use_lion --lr 2e-4 --surface_refine_hidden 192 --surface_refine_layers 3 --residual_prediction`
- KD runs add: `--tandem_selfdistill --tandem_selfdistill_weight <w>`
- Implementation: After EMA activation, MSE between online and EMA backbone predictions on surface pressure (channel 2) for tandem samples only

### Critical Issue: GPU Speed Variance → Post-Cosine Degradation

GPUs 1-3 ran at ~43-45s/epoch (ctrl), reaching **220-252 epochs** before the 180-min timeout. GPU 0 (w=0.05) ran at ~80s/epoch, reaching only **136 epochs**. The cosine LR schedule ends at epoch 160 (`cosine_T_max=160`).

**Runs that went far beyond epoch 160 showed severe degradation**, especially on OOD metrics. This affects ALL runs on GPUs 1-3, including the control. The model overfits to in-distribution data after the LR schedule bottoms out, and the EMA tracks this degradation.

### Results — Surface MAE (Pressure)

| Run | Epochs | p_in | p_oodc | p_tan | p_re | val/loss | W&B Run ID |
|-----|--------|------|--------|-------|------|----------|------------|
| **w=0.05 s42** | **136** | **13.6** | **8.3** | **30.4** | **6.7** | **0.396** | `swsmtceu` |
| w=0.10 s42 | 219 | 14.1 | 10.8 | 38.2 | 25.9 | 0.735 | `ewrhonfl` |
| w=0.20 s42 | 220 | 14.9 | 10.4 | 35.6 | 25.6 | 0.711 | `d6x7l5cg` |
| ctrl s42 | 252 | 15.0 | 9.9 | 36.5 | 25.4 | 0.713 | `ivs2xoxz` |
| **Baseline (2-seed)** | — | **13.02** | **7.62** | **29.91** | **6.47** | — | — |

⚠️ **Only the w=0.05 run is a valid result** (136 epochs ≈ within cosine schedule). The other 3 runs suffered post-cosine degradation and their metrics are not meaningful.

### Analysis of the Valid Run (w=0.05)

The w=0.05 run does not beat any baseline metric:
- p_in: 13.6 vs 13.02 (+4.5%)
- p_oodc: 8.3 vs 7.62 (+8.9%)
- p_tan: 30.4 vs 29.91 (+1.6%)
- p_re: 6.7 vs 6.47 (+3.6%)

However, this is a single run on a slow GPU that only had **36 epochs of KD** (EMA started at epoch 100, training stopped at epoch 136). The KD signal barely had time to influence training. A fair test would need the same epoch count as the baseline (~150-170 epochs).

### Why GPUs 1-3 Were So Fast

With `--aft_foil_srf_context`, the per-epoch time dropped from the expected ~63s to ~43-45s. The KD forward pass added ~17s overhead on KD runs (62-63s/epoch), but even KD runs went to 219-220 epochs — well past the cosine schedule endpoint.

### Root Cause: cosine_T_max vs Actual Epochs

The core problem: `cosine_T_max=160` assumes ~170 epochs in 180 min (~63s/epoch). With `--aft_foil_srf_context`, epochs are faster (~45s/epoch), reaching 250+ epochs. Training 90+ epochs beyond the cosine schedule causes severe OOD degradation.

### Suggested Follow-ups

1. **Re-run with `--cosine_T_max 240`** to match the actual epoch count for these GPUs, ensuring the LR schedule covers the full training duration
2. **OR stop training at epoch 160** by capping max epochs (requires code change or reduced `SENPAI_TIMEOUT_MINUTES`)
3. **The KD idea itself is inconclusive** — the one valid run didn't have enough KD epochs to test the hypothesis. If re-run with proper LR schedule, results may differ
4. **The GPU speed variance issue is general** — any experiment using `--aft_foil_srf_context` will face this problem unless cosine_T_max is adjusted

tcapelle:
ADVISOR: Closing PR #2135 because ADVISOR: Closing — inconclusive due to post-cosine degradation. Only w=0.05 (136 epochs) was valid; it regressed on all metrics (p_tan +1.6%, p_oodc +8.9%). Also confounded by the aft_foil_srf_context guard bug (frieren #2134) — the context head was never applied, so epoch timing differed from expected. The KD hypothesis had too little exposure time (36 epochs) but the initial signal is negative. Good catch on the post-cosine degradation issue — this is real and affects all fast-GPU runs with cosine_T_max=160. Reassigning to a fresh experiment..

PR #2180, opened 2026-04-05
Multi-Resolution Hash Grid Encoding of Mesh Coordinates

tcapelle:
## Hypothesis

The Transolver's DSDF features encode distance/direction to foil surfaces at a fixed resolution. The network must learn to map this low-frequency input to high-frequency pressure gradients (leading edge stagnation, suction peaks, trailing edge separation) through all its layers. Neural networks exhibit **spectral bias** — they preferentially learn low-frequency components first (Rahaman et al., 2019). A multi-resolution hash grid encoding of (x, y) mesh coordinates provides explicit high-frequency spatial features that the backbone can access directly, breaking the spectral bias bottleneck.

**Evidence:** "Parametric hash encoding for physics surrogates" (arXiv:2403.15652, UC Irvine, 2024) showed that hash grid encodings applied to mesh coordinates for PDE surrogates reduce error by **30-50%** on problems with fine spatial structure. Hash grids are trained end-to-end alongside the network.

**Key difference from Fourier PE:** Fourier features are fixed-frequency; hash grid features are learned and can represent arbitrary spatial frequencies by discovering which grid cells matter for pressure prediction.

### Step 1: Implement `HashGrid2D` module

Add a new `HashGrid2D` class in `train.py` (near the other model components). This implements a 2D multi-resolution hash grid:

```python
class HashGrid2D(nn.Module):
    """Multi-resolution 2D hash grid encoding (Instant-NGP style)."""
    def __init__(self, n_levels=8, features_per_level=4, base_resolution=16,
                 max_resolution=2048, hash_table_size=2**16,
                 x_range=(-0.5, 2.5), y_range=(-1.0, 1.0)):
        super().__init__()
        self.n_levels = n_levels
        self.features_per_level = features_per_level
        self.output_dim = n_levels * features_per_level  # 32

        # Compute resolution at each level (geometric progression)
        resolutions = []
        b = (max_resolution / base_resolution) ** (1.0 / (n_levels - 1))
        for i in range(n_levels):
            resolutions.append(int(base_resolution * b**i))
        self.resolutions = resolutions

        # Hash tables: one embedding table per level
        self.hash_tables = nn.ModuleList([
            nn.Embedding(min(res * res, hash_table_size), features_per_level)
            for res in resolutions
        ])

        # Initialize small
        for table in self.hash_tables:
            nn.init.uniform_(table.weight, -1e-4, 1e-4)

        # Spatial range for normalization
        self.register_buffer('x_min', torch.tensor([x_range[0], y_range[0]]))
        self.register_buffer('x_max', torch.tensor([x_range[1], y_range[1]]))

    def forward(self, xy):
        """
        xy: (B, N, 2) or (N, 2) — mesh (x, y) coordinates
        Returns: (B, N, output_dim) or (N, output_dim) hash grid features
        """
        squeeze = False
        if xy.dim() == 2:
            xy = xy.unsqueeze(0)
            squeeze = True

        B, N, _ = xy.shape
        # Normalize to [0, 1]
        xy_norm = (xy - self.x_min) / (self.x_max - self.x_min + 1e-8)
        xy_norm = xy_norm.clamp(0, 1)

        features = []
        for level_idx, res in enumerate(self.resolutions):
            # Scale to grid resolution
            xy_scaled = xy_norm * (res - 1)  # (B, N, 2)

            # Floor and ceil for bilinear interpolation
            xy_floor = xy_scaled.long().clamp(0, res - 2)
            xy_ceil = xy_floor + 1
            frac = xy_scaled - xy_floor.float()  # fractional part

            # 4 corner indices for bilinear interp
            hash_size = self.hash_tables[level_idx].num_embeddings
            def hash_idx(ix, iy):
                # Simple spatial hash
                return ((ix * 2654435761) ^ (iy * 805459861)) % hash_size

            idx_00 = hash_idx(xy_floor[..., 0], xy_floor[..., 1])  # (B, N)
            idx_01 = hash_idx(xy_floor[..., 0], xy_ceil[..., 1])
            idx_10 = hash_idx(xy_ceil[..., 0], xy_floor[..., 1])
            idx_11 = hash_idx(xy_ceil[..., 0], xy_ceil[..., 1])

            # Look up features
            f00 = self.hash_tables[level_idx](idx_00)  # (B, N, F)
            f01 = self.hash_tables[level_idx](idx_01)
            f10 = self.hash_tables[level_idx](idx_10)
            f11 = self.hash_tables[level_idx](idx_11)

            # Bilinear interpolation
            fx = frac[..., 0:1]  # (B, N, 1)
            fy = frac[..., 1:2]
            f = (f00 * (1-fx) * (1-fy) +
                 f01 * (1-fx) * fy +
                 f10 * fx * (1-fy) +
                 f11 * fx * fy)

            features.append(f)

        out = torch.cat(features, dim=-1)  # (B, N, n_levels * features_per_level)
        if squeeze:
            out = out.squeeze(0)
        return out
```

**IMPORTANT:** Check what the actual coordinate range is in the dataset. Read a few batches and print `x[:, :, 0].min(), x[:, :, 0].max()` and `x[:, :, 1].min(), x[:, :, 1].max()` to set `x_range` and `y_range` correctly. The defaults above are guesses.

### Step 2: Integrate into the forward pass

Add `--hash_grid_encoding` flag. When enabled:

1. Instantiate `HashGrid2D` as `self.hash_grid` in the model `__init__`
2. In the forward pass, extract (x, y) coordinates from the input: `xy = x[:, :, :2]` (first two channels are x, y coordinates)
3. Compute hash features: `hash_feat = self.hash_grid(xy)` → shape (B, N, 32)
4. Concatenate to existing features: `x = torch.cat([x, hash_feat], dim=-1)` → input dim becomes 24 + 32 = 56
5. Update the first linear layer input dimension accordingly (you'll need to make this conditional on the flag)

### Step 3: Separate learning rate for hash grid parameters

The hash grid parameters benefit from a higher learning rate than the backbone. Create two parameter groups:

```python
hash_params = list(model.hash_grid.parameters())
other_params = [p for name, p in model.named_parameters() if 'hash_grid' not in name]

optimizer = Lion([
    {'params': other_params, 'lr': 2e-4},
    {'params': hash_params, 'lr': 1e-3},  # 5x higher for hash grid
], weight_decay=5e-5)
```

Both groups still use cosine annealing with T_max=160.

### Step 4: Run experiments

Run **2 seeds** with the W&B group:

```bash
# Seed 42
  --wandb_name "edward/hash-grid-encoding-s42" \
  --wandb_group "edward/hash-grid-encoding" \
  --hash_grid_encoding \
  --gap_stagger_spatial_bias

# Seed 73
cd cfd_tandemfoil && python train.py --agent edward --seed 73 \
  --wandb_name "edward/hash-grid-encoding-s73" \
  --wandb_group "edward/hash-grid-encoding" \
  --hash_grid_encoding \
  --gap_stagger_spatial_bias
```

### Key implementation notes

1. **Coordinate range:** MUST check actual data range before hardcoding. Print min/max of x[:,:,0] and x[:,:,1] from a few training batches.
2. **torch.compile compatibility:** The hash grid uses standard nn.Embedding + basic tensor ops, so it should be compile-friendly. Test with `torch.compile` before the full run.
3. **VRAM:** The hash tables add ~8 * min(res^2, 65536) * 4 * 4 bytes ≈ ~8MB total. Negligible VRAM impact.
4. **Hash function:** The XOR-based spatial hash above is the standard Instant-NGP hash. Ensure the hash indices are computed with `long()` tensors to avoid overflow.
5. **Gradient flow:** The hash grid learns through bilinear interpolation gradients. This is standard and well-tested in NeRF/Instant-NGP implementations.
6. **DSDF channel indices reminder:** Foil-1 DSDF channels are at `x[:,4:8]` NOT `x[:,2:6]`. Foil-2 at `x[:,6:10]`. XY coordinates are `x[:,:,:2]`.

## Baseline

| p_in | **13.05** | < 13.05 |
| p_oodc | **7.70** | < 7.70 |
| **p_tan** | **28.60** | **< 28.60** |
| p_re | **6.55** | < 6.55 |

**Baseline W&B runs:** d7l91p0x (s42, p_tan=28.9), j9btfx09 (s73, p_tan=28.3)

**Reproduce baseline:**
```bash
cd cfd_tandemfoil && python train.py --agent edward --wandb_name "edward/baseline-gsb-pcgrad" \
  --gap_stagger_spatial_bias
```

### DIFF
--- a/cfd_tandemfoil/train.py
+++ b/cfd_tandemfoil/train.py
@@ -134,6 +134,82 @@ class DomainLayerNorm(nn.Module):
         return torch.where(mask_t, self.ln_tandem(x), self.ln_single(x))
 
 
+class HashGrid2D(nn.Module):
+    """Multi-resolution 2D hash grid encoding (Instant-NGP style).
+
+    Expects per-sample normalized coordinates in [0, 1]^2.
+    Output dim = n_levels * features_per_level.
+    """
+    def __init__(self, n_levels=8, features_per_level=4, base_resolution=16,
+                 max_resolution=2048, hash_table_size=2**16):
+        super().__init__()
+        self.n_levels = n_levels
+        self.features_per_level = features_per_level
+        self.output_dim = n_levels * features_per_level  # 32
+
+        # Compute resolution at each level (geometric progression)
+        b = (max_resolution / base_resolution) ** (1.0 / (n_levels - 1))
+        self.resolutions = [int(base_resolution * b**i) for i in range(n_levels)]
+
+        # Hash tables: one embedding table per level
+        self.hash_tables = nn.ModuleList([
+            nn.Embedding(min(res * res, hash_table_size), features_per_level)
+            for res in self.resolutions
+        ])
+
+        # Initialize small so it starts near-zero (doesn't disrupt baseline)
+        for table in self.hash_tables:
+            nn.init.uniform_(table.weight, -1e-4, 1e-4)
+
+    def forward(self, xy_norm):
+        """
+        xy_norm: (B, N, 2) — per-sample normalized mesh coordinates in [0, 1].
+        Returns: (B, N, output_dim) hash grid features.
+        """
+        xy_norm = xy_norm.clamp(0.0, 1.0)
+        features = []
+        for level_idx, res in enumerate(self.resolutions):
+            # Scale to grid resolution
+            xy_scaled = xy_norm * (res - 1)  # (B, N, 2)
+
+            # Floor and ceil for bilinear interpolation
+            xy_floor = xy_scaled.long().clamp(0, res - 2)
+            xy_ceil = xy_floor + 1
+            frac = xy_scaled - xy_floor.float()  # fractional part
+
+            # 4 corner indices for bilinear interp (XOR spatial hash)
+            hash_size = self.hash_tables[level_idx].num_embeddings
+            ix0, iy0 = xy_floor[..., 0], xy_floor[..., 1]
+            ix1, iy1 = xy_ceil[..., 0], xy_ceil[..., 1]
+
+            def _hash(ix, iy):
+                return ((ix * 2654435761) ^ (iy * 805459861)) % hash_size
+
+            idx_00 = _hash(ix0, iy0)
+            idx_01 = _hash(ix0, iy1)
+            idx_10 = _hash(ix1, iy0)
+            idx_11 = _hash(ix1, iy1)
+
+            # Look up features
+            table = self.hash_tables[level_idx]
+            f00 = table(idx_00)  # (B, N, F)
+            f01 = table(idx_01)
+            f10 = table(idx_10)
+            f11 = table(idx_11)
+
+            # Bilinear interpolation
+            fx = frac[..., 0:1]  # (B, N, 1)
+            fy = frac[..., 1:2]
+            f = (f00 * (1 - fx) * (1 - fy) +
+                 f01 * (1 - fx) * fy +
+                 f10 * fx * (1 - fy) +
+                 f11 * fx * fy)
+
+            features.append(f)
+
+        return torch.cat(features, dim=-1)  # (B, N, n_levels * features_per_level)
+
+
 class Physics_Attention_Irregular_Mesh(nn.Module):
     """Physics attention for irregular meshes in 1D/2D/3D space."""
 
@@ -700,6 +776,7 @@ class Transolver(nn.Module):
         pressure_no_detach=False,
         pressure_deep=False,
         gap_stagger_spatial_bias=False,
+        hash_grid_encoding=False,
     ):
         super().__init__()
         self.__name__ = "UniPDE_3D"
@@ -802,6 +879,13 @@ class Transolver(nn.Module):
         self.aoa_head = nn.Sequential(nn.Linear(n_hidden, 32), nn.GELU(), nn.Linear(32, 1))
         self.fourier_freqs_fixed = torch.tensor([0.5, 2.0, 8.0, 32.0])  # non-learnable
         self.fourier_freqs_learned = nn.Parameter(torch.tensor([1.0, 3.0, 6.0, 16.0]))
+        # Hash grid encoding — instantiated after initialize_weights() to preserve small init
+        if hash_grid_encoding:
+            self.hash_grid = HashGrid2D(
+                n_levels=8, features_per_level=4,
+                base_resolution=16, max_resolution=2048,
+                hash_table_size=2**16,
+            )
 
     def initialize_weights(self):
         self.apply(self._init_weights)
@@ -1070,6 +1154,8 @@ class Config:
     # Phase 6: 3-way PCGrad — gradient surgery with single-foil | tandem-normal | tandem-extreme-Re
     pcgrad_3way: bool = False               # enable 3-way gradient surgery (requires --disable_pcgrad)
     pcgrad_extreme_pct: float = 0.15        # top/bottom Re percentile among tandem samples to label as extreme
+    # Multi-resolution hash grid encoding of mesh coordinates
+    hash_grid_encoding: bool = False         # enable Instant-NGP-style hash grid spatial encoding
 
 
 cfg = sp.parse(Config)
@@ -1200,7 +1286,7 @@ else:
 
 model_config = dict(
     space_dim=2,
-    fun_dim=X_DIM - 2 + 2 + (1 if cfg.foil2_dist else 0) + 32,  # +curv, +dist, [+foil2dist], +32 fourier PE
+    fun_dim=X_DIM - 2 + 2 + (1 if cfg.foil2_dist else 0) + 32 + (32 if cfg.hash_grid_encoding else 0),  # +curv, +dist, [+foil2dist], +32 fourier PE, [+32 hash grid]
     out_dim=3,
     n_hidden=cfg.n_hidden,
     n_layers=cfg.n_layers,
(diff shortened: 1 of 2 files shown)


tcapelle:
STUDENT edward:

## Results

### Implementation notes

- **Coordinate range correction**: The PR suggested hardcoded ranges x=(-0.5, 2.5), y=(-1.0, 1.0). Actual data inspection showed x=[-9.55, 13.55], y=[-9.59, 9.61]. Instead of hardcoding, I used per-sample normalized [0,1] coordinates (same as the existing Fourier PE), which is robust to varying domain sizes.
- **DSDF channel correction**: PR body noted "Foil-1 DSDF at x[:,2:6]" in one section and "x[:,4:8]" in the notes — the hash grid uses xy at x[:,:,:2] (pos channels) so this doesn't affect the hash grid implementation.
- **Hash grid instantiated after `initialize_weights()`** so the small uniform init (-1e-4, 1e-4) survives orthogonal initialization.
- **Separate 5x LR for hash grid params**: Hash grid params (1.14M) get lr=1e-3 while backbone gets lr=2e-4.
- Hash grid: 8 levels × 4 features/level = 32 output dims; base_res=16, max_res=2048, hash_table_size=65536.

### Metrics

| Split | Baseline avg | s42 (3sb91xty) | s73 (kie1y6kb) | Hash Grid avg | Delta |
|-------|-------------|----------------|----------------|--------------|-------|
| p_in  | 13.05       | 18.4           | 17.4           | **17.9**     | +4.85 |
| p_oodc| 7.70        | 9.3            | 9.2            | **9.25**     | +1.55 |
| **p_tan** | **28.60** | 32.6         | 31.6           | **32.1**     | **+3.5** |
| p_re  | 6.55        | 8.1            | 7.6            | **7.85**     | +1.3  |

**Full surface MAE for best run (s73, kie1y6kb):**
- val_in_dist: Ux=3.45, Uy=0.89, p=17.4, vol_p=16.2
- val_ood_cond: p=9.2
- val_tandem_transfer: p=31.6
- val_ood_re: p=7.6

Peak memory: 46.6–46.8 GB. Best epoch: 121 (s42), 122 (s73) — about 121 epochs in 180 min (vs ~150 for baseline).

W&B runs: `3sb91xty` (s42), `kie1y6kb` (s73) — group: `edward/hash-grid-encoding`

### Commands

```bash
CUDA_VISIBLE_DEVICES=0 python train.py --agent edward --seed 42 \
  --wandb_name "edward/hash-grid-encoding-s42" \
  --wandb_group "edward/hash-grid-encoding" \
  --hash_grid_encoding \
  --gap_stagger_spatial_bias
```

### What happened

The hash grid encoding **significantly degraded performance across ALL metrics**, with p_tan going from 28.60 → 32.1 (+12.2%). This is a clear negative result.

**Why it failed — three likely contributing factors:**

1. **Per-sample normalization defeats the hash grid's core mechanism.** Instant-NGP hash grids rely on a fixed coordinate system where the same physical location always maps to the same hash entries. Our per-sample normalization to [0,1] means the same physical point (e.g., leading edge) maps to different hash entries in different samples depending on the domain extent. This prevents the hash grid from learning spatially-coherent features — each entry receives conflicting gradient signals from different spatial regions across samples.

2. **Fewer training epochs due to computational overhead.** The hash grid added ~20s/epoch overhead (90s vs 70s baseline), resulting in only ~121 epochs vs ~150 for baseline — a 19% reduction. Since cosine_T_max=160, the runs barely reached 75% of the cosine schedule, meaning the LR hadn't fully decayed. The model was still improving when it hit the timeout.

3. **Redundancy with existing Fourier PE.** The model already has 32-dim learned Fourier positional encoding of the (x, y) coordinates. The hash grid provides fundamentally similar multi-frequency spatial information. The additional 32 dims may have added noise rather than complementary signal, especially given the per-sample normalization issue.

4. **Massive parameter increase without regularization.** The hash grid added 1.14M params (a ~25% increase in total parameters) with weak regularization (only weight_decay=5e-5). Combined with the 5x higher LR, the hash tables may have overfit quickly, capturing sample-specific spatial noise rather than generalizable patterns.

### Suggested follow-ups

1. **Fixed coordinate normalization**: Instead of per-sample normalization, use a fixed coordinate range (e.g., x=[-10, 14], y=[-10, 10]) based on the dataset statistics. This would let the hash grid learn spatially-coherent features tied to physical locations (leading edge, wake region, etc.).

2. **Reduce hash grid capacity**: Try fewer levels (4 instead of 8), smaller hash tables (2^14 instead of 2^16), or 2 features/level instead of 4 — total output dim of 8 instead of 32. This reduces overfitting risk while still providing multi-resolution spatial info.

3. **Replace (not concatenate with) Fourier PE**: Since the hash grid and Fourier PE serve similar roles, try replacing the 32-dim Fourier PE with the 32-dim hash grid instead of stacking both. This keeps the input dimension the same and avoids redundancy.

4. **Longer timeout or more efficient implementation**: The 20s/epoch overhead is substantial. Using a dedicated CUDA kernel (e.g., `tiny-cuda-nn`) instead of pure PyTorch embedding lookups would make the hash grid nearly free in terms of compute.

tcapelle:
ADVISOR: Closing PR #2180 because ADVISOR: Closing — hash grid severely regresses ALL metrics. p_tan 28.60→32.1 (+12.2%).

Your root cause analysis is exactly right: per-sample coordinate normalization defeats the hash grid's core mechanism. When the leading edge maps to different hash entries in different samples, the grid can't learn spatially-coherent features. Combined with 1.14M extra params and only 121 epochs (19% fewer than baseline), this was a triple penalty.

The 20s/epoch overhead is also a significant finding — any approach that costs >10% wall-clock time needs to compensate with faster convergence, which didn't happen here.

Your follow-up #1 (fixed coordinate range) would fix the normalization issue, but the Fourier PE redundancy and parameter bloat concerns remain. I think the fundamental insight is that spatial frequency features are NOT the bottleneck — the existing 32-dim Fourier PE is already sufficient for the model's spatial reasoning needs.

Good engineering work on the implementation. Moving you to a new experiment..

PR #2130, opened 2026-04-04
Phase 6: Gap/Stagger-Conditioned Spatial Bias — Tandem-Geometry-Aware Slice Routing

tcapelle:
## Hypothesis

Transolver's slice routing assigns each mesh node to a soft mixture of physics slices via \`slice_logits = ... + 0.1 * spatial_bias(raw_xy)\`. Currently, \`raw_xy\` is purely local: (x_coord, y_coord, curvature, dist-to-surface). The spatial_bias MLP has **zero knowledge of the inter-foil tandem geometry**.

For single-foil samples, gap=0 and stagger=0 — so extending raw_xy with these features adds **no information for single-foil samples** (zero input → zero effect). For tandem samples, different gap/stagger values fundamentally change the flow topology (wake interaction, channel flow, suction-side shielding), but all tandem configurations currently route through identical slice assignment logic regardless of inter-foil separation.

This experiment conditions Transolver's core slice routing on global tandem geometry. PR #2115 (merged) showed that gap/stagger perturbation during training improves p_oodc. This experiment is the **complementary mechanism**: making slice routing explicitly *aware* of which inter-foil geometry it's processing — not just robust to variation in it. The two mechanisms are orthogonal.

**Important new finding:** PR #2123 (combined baseline validation) showed that gap/stagger augmentation helps p_oodc (-2.4%) but slightly regresses p_tan (+1.6%). The spatial-bias approach operates through a completely different channel (slice routing, not augmentation) and is expected to help p_tan by giving the trunk genuine geometric awareness.

**Expected impact:** -3 to -7% p_tan. Negligible parameter overhead (384 extra scalars across 3 blocks). Zero effect on single-foil samples by design.

### Step 1 — Add config flag

```python
# In Config dataclass:
gap_stagger_spatial_bias: bool = False
```

### Step 2 — Modify TransolverBlock.__init__ to accept spatial_bias_input_dim

```python
# In TransolverBlock.__init__, change spatial_bias construction to accept input_dim param:
# Add spatial_bias_input_dim=4 to __init__ signature
def __init__(self, ..., spatial_bias_input_dim=4):
    ...
    self.spatial_bias = nn.Sequential(
        nn.Linear(spatial_bias_input_dim, 64), nn.GELU(),
        nn.Linear(64, 64), nn.GELU(),
        nn.Linear(64, slice_num),
    )
    nn.init.zeros_(self.spatial_bias[-1].weight)
    nn.init.zeros_(self.spatial_bias[-1].bias)
```

[...]

PR #2193, opened 2026-04-06
Curvature-Conditioned Spatial Bias: True Arc-Length Curvature for Slice Routing

tcapelle:
## Hypothesis

The Gap/Stagger Spatial Bias (PR #2130, -3.0% p_tan) was the biggest single win by extending the `spatial_bias` MLP with global tandem scalars. The MLP currently receives 6 inputs: `[x_std, y_std, crude_curv_proxy, log_dist_surf, gap, stagger]`. However, the so-called "curvature proxy" at index 24 is **NOT actual curvature** — it is `norm(standardized(saf_x, saf_y, dsdf_0, dsdf_1)) * is_surface`, which approximates 1 for surface nodes and 0 elsewhere. It does not encode how much the surface bends locally.

**True arc-length curvature** (κ = |d²r/ds²| along the foil surface) varies dramatically: very high at the leading edge, near-zero at mid-chord, moderate at the trailing edge. NACA6416 (OOD geometry) has higher camber (6%) than typical training foils, which shifts both the suction-peak location and its magnitude. If the spatial_bias MLP learns to route high-curvature nodes to dedicated slices, it gains a geometry-aware routing signal that generalizes to unseen camber values.

This is **orthogonal to all existing inputs** (position, gap/stagger are global, dist_surf is distance not shape). Adding curvature adds genuinely new local shape information.

### Step 1: Implement `compute_surface_curvature_batch`

Add this function to `train.py` (near other helper functions):

```python
def compute_surface_curvature_batch(coords_raw, is_surface, saf_norm, is_tandem):
    """
    Compute Menger curvature at surface nodes for each sample in the batch.
    coords_raw: [B, N, 2] raw (pre-standardization) x,y coordinates
    is_surface: [B, N] boolean
    saf_norm: [B, N] norm of saf channels x[:, :, 2:4] (before std)
    is_tandem: [B] boolean
    Returns: [B, N, 1] curvature (0 for non-surface nodes), normalized to [0, 1]
    """
    B, N = coords_raw.shape[:2]
    kappa = torch.zeros(B, N, 1, device=coords_raw.device, dtype=coords_raw.dtype)

    for b in range(B):
        foil_masks = [is_surface[b] & (saf_norm[b] <= 0.005)]  # foil-1 surface
        if is_tandem[b]:
            foil_masks.append(is_surface[b] & (saf_norm[b] > 0.005))  # foil-2 surface

        for mask in foil_masks:
            idx = mask.nonzero(as_tuple=False).squeeze(-1)
            if len(idx) < 3:
                continue
            pts = coords_raw[b, idx, :2]  # [M, 2]
            centroid = pts.mean(0)
            angles = torch.atan2(pts[:, 1] - centroid[1], pts[:, 0] - centroid[0])
            order = angles.argsort()
            pts_s = pts[order]  # [M, 2] sorted by angle around centroid
            M = len(pts_s)
            arange = torch.arange(M, device=pts_s.device)
            p0  = pts_s[arange]
            pm1 = pts_s[(arange - 1) % M]
            pp1 = pts_s[(arange + 1) % M]
            d1 = (p0 - pm1).norm(dim=-1)
            d2 = (pp1 - p0).norm(dim=-1)
            d3 = (pp1 - pm1).norm(dim=-1)
            cross = (p0 - pm1)[:, 0] * (pp1 - p0)[:, 1] \
                  - (p0 - pm1)[:, 1] * (pp1 - p0)[:, 0]
            k = 2.0 * cross.abs() / (d1 * d2 * d3).clamp(min=1e-10)
            kappa[b, idx[order], 0] = k

    kappa = kappa.clamp(max=50.0) / 50.0  # normalize to [0, 1] range
    return kappa
```

### Step 2: Compute curvature in training loop (before standardization)

Just **before** the standardization call (`x = (x - stats["x_mean"]) / stats["x_std"]`), add:

```python
if cfg.curvature_spatial_bias:
    # Re-use _raw_saf_norm and _is_tandem if already computed (from aft_foil_srf path)
    _csb_saf_norm = x[:, :, 2:4].norm(dim=-1)   # [B, N]
    _csb_is_tandem = (x[:, 0, 22].abs() > 0.01)  # [B]
    curv_true = compute_surface_curvature_batch(
        x[:, :, :2], is_surface, _csb_saf_norm, _csb_is_tandem
    )  # [B, N, 1]
else:
    curv_true = None
```

### Step 3: Pass curvature to the model

Pass `curv_true` as a kwarg to the model: `pred = model(x, curvature=curv_true, ...)`.

In `Transolver.forward(self, x, ..., curvature=None)`, when assembling `raw_xy` (around line 887-892):

```python
if self.gap_stagger_spatial_bias:
    gap_stagger = x[:, 0:1, 22:24].expand(-1, x.shape[1], -1)
    raw_xy = torch.cat([x[:, :, :2], x[:, :, 24:26], gap_stagger], dim=-1)  # [B, N, 6]
else:
    raw_xy = torch.cat([x[:, :, :2], x[:, :, 24:26]], dim=-1)

# Append true curvature if enabled
if self.curvature_spatial_bias and curvature is not None:
    raw_xy = torch.cat([raw_xy, curvature], dim=-1)  # → [B, N, 7]
```

### Step 4: Extend spatial_bias MLP input

In `Transolver.__init__`, update the `spatial_bias_input_dim` logic:

```python
_sb_dim = 4
if cfg.gap_stagger_spatial_bias:
    _sb_dim += 2
if cfg.curvature_spatial_bias:
    _sb_dim += 1
# pass _sb_dim to every TransolverBlock(..., spatial_bias_input_dim=_sb_dim)
```

Zero-init the new curvature column after construction (same pattern as gap_stagger):

```python
if cfg.curvature_spatial_bias:
    with torch.no_grad():
        for block in model.blocks:
            block.spatial_bias[0].weight[:, -1].zero_()
```

This ensures routing is identical to baseline at epoch 0.

### Step 5: Add flag and run

Add `--curvature_spatial_bias` boolean flag (default False).

```bash
# Seed 42
  --wandb_name "edward/curvature-sb-s42" \
  --wandb_group "edward/curvature-spatial-bias" \
  --curvature_spatial_bias \
  --gap_stagger_spatial_bias \
  --dct_freq_loss --dct_freq_weight 0.05 --dct_freq_gamma 2.0 --dct_freq_alpha 1.5

# Seed 73: same with --seed 73 and --wandb_name "edward/curvature-sb-s73"
```

### Key implementation notes

1. **Separate foils for tandem.** Use `saf_norm <= 0.005` for foil-1, `> 0.005` for foil-2. This is the same proxy as `_aft_foil_mask`. Never mix both foils — your #2187 experiment confirmed this causes incorrect normals.

2. **Normalization check.** Print `kappa.max()` on a few batches. Typical LE curvature ≈ 20-50 in normalized coordinate units. If max is much lower, the clamp is too aggressive. If much higher, increase it.

3. **Zero extra parameters.** The curvature adds 1 column to the first Linear of each `spatial_bias` MLP: 3 × 64 = 192 extra weights total. Negligible.

4. **torch.compile.** The Python loop in `compute_surface_curvature_batch` runs outside compile; the vectorized ops inside are fully compile-friendly.

5. **Verify gradients.** After the first batch, check `model.blocks[0].spatial_bias[0].weight.grad[:, -1].abs().max()`. Should be nonzero — confirms curvature is flowing through the graph.

## Baseline

| p_in | **13.21** | < 13.21 |
| p_oodc | **7.82** | < 7.82 |
| **p_tan** | **28.50** | **< 28.50** |
| p_re | **6.45** | < 6.45 |

**Baseline W&B runs:** 6yfv5lio (seed 42, p_tan=28.432), etepxvjc (seed 73, p_tan=28.572)

**Reproduce baseline:**
```bash
cd cfd_tandemfoil && python train.py --agent edward --wandb_name "edward/baseline-dct-freq" \
  --gap_stagger_spatial_bias \
  --dct_freq_loss --dct_freq_weight 0.05 --dct_freq_gamma 2.0 --dct_freq_alpha 1.5
```

### DIFF
--- a/cfd_tandemfoil/train.py
+++ b/cfd_tandemfoil/train.py
@@ -134,6 +134,58 @@ class DomainLayerNorm(nn.Module):
         return torch.where(mask_t, self.ln_tandem(x), self.ln_single(x))
 
 
+def compute_surface_curvature_batch(coords_raw, is_surface, saf_norm, is_tandem):
+    """Compute Menger curvature at surface nodes for each sample in the batch.
+
+    Processes foil-1 and foil-2 surfaces separately to avoid cross-foil angle
+    sorting artifacts (foil-1: saf_norm ≤ 0.005; foil-2: saf_norm > 0.005).
+
+    Args:
+        coords_raw: [B, N, 2] raw (pre-standardization) x,y coordinates
+        is_surface: [B, N] boolean
+        saf_norm:   [B, N] norm of saf channels (x[:, :, 2:4]) before standardization
+        is_tandem:  [B] boolean — True for tandem samples
+
+    Returns:
+        kappa: [B, N, 1] log1p-scaled curvature at surface nodes, 0 elsewhere.
+               Scaled as log1p(κ) / log1p(50) → approximately [0, 1] for typical foils.
+    """
+    B, N = coords_raw.shape[:2]
+    kappa = torch.zeros(B, N, 1, device=coords_raw.device, dtype=coords_raw.dtype)
+
+    for b in range(B):
+        foil_masks = [is_surface[b] & (saf_norm[b] <= 0.005)]  # foil-1 surface
+        if is_tandem[b]:
+            foil_masks.append(is_surface[b] & (saf_norm[b] > 0.005))  # foil-2 surface
+
+        for mask in foil_masks:
+            idx = mask.nonzero(as_tuple=False).squeeze(-1)
+            if idx.numel() < 3:
+                continue
+            pts = coords_raw[b, idx, :2]           # [M, 2]
+            centroid = pts.mean(0)
+            angles = torch.atan2(pts[:, 1] - centroid[1], pts[:, 0] - centroid[0])
+            order = angles.argsort()
+            pts_s = pts[order]                      # [M, 2] sorted by angle
+            M = pts_s.shape[0]
+            arange = torch.arange(M, device=pts_s.device)
+            p0  = pts_s[arange]
+            pm1 = pts_s[(arange - 1) % M]
+            pp1 = pts_s[(arange + 1) % M]
+            d1 = (p0 - pm1).norm(dim=-1)
+            d2 = (pp1 - p0).norm(dim=-1)
+            d3 = (pp1 - pm1).norm(dim=-1)
+            cross = ((p0 - pm1)[:, 0] * (pp1 - p0)[:, 1]
+                   - (p0 - pm1)[:, 1] * (pp1 - p0)[:, 0])
+            k = 2.0 * cross.abs() / (d1 * d2 * d3).clamp(min=1e-10)
+            kappa[b, idx[order], 0] = k
+
+    # log1p normalization: maps typical LE curvature (20-50 units) to ~0.8-1.0
+    log50 = torch.tensor(51.0, device=kappa.device, dtype=kappa.dtype).log1p()
+    kappa = kappa.clamp(max=50.0).log1p() / log50
+    return kappa
+
+
 class Physics_Attention_Irregular_Mesh(nn.Module):
     """Physics attention for irregular meshes in 1D/2D/3D space."""
 
@@ -700,10 +752,12 @@ class Transolver(nn.Module):
         pressure_no_detach=False,
         pressure_deep=False,
         gap_stagger_spatial_bias=False,
+        curvature_spatial_bias=False,
     ):
         super().__init__()
         self.__name__ = "UniPDE_3D"
         self.gap_stagger_spatial_bias = gap_stagger_spatial_bias
+        self.curvature_spatial_bias = curvature_spatial_bias
         self.pressure_first = pressure_first
         self.ref = ref
         self.unified_pos = unified_pos
@@ -771,16 +825,23 @@ class Transolver(nn.Module):
                     pressure_first=pressure_first if (idx == n_layers - 1) else False,
                     pressure_no_detach=pressure_no_detach,
                     pressure_deep=pressure_deep,
-                    spatial_bias_input_dim=6 if gap_stagger_spatial_bias else 4,
+                    spatial_bias_input_dim=(4
+                                           + (2 if gap_stagger_spatial_bias else 0)
+                                           + (1 if curvature_spatial_bias else 0)),
                 )
                 for idx in range(n_layers)
             ]
         )
-        # Zero-init the 2 new input columns of spatial_bias so initial routing is unchanged
+        # Zero-init extra input columns of spatial_bias so initial routing is unchanged
         if gap_stagger_spatial_bias:
             with torch.no_grad():
                 for block in self.blocks:
-                    block.spatial_bias[0].weight[:, 4:].zero_()
+                    block.spatial_bias[0].weight[:, 4:6].zero_()
+        if curvature_spatial_bias:
+            _curv_col = 6 if gap_stagger_spatial_bias else 4
+            with torch.no_grad():
+                for block in self.blocks:
+                    block.spatial_bias[0].weight[:, _curv_col].zero_()
         # Separate pressure pathway (pressure_separate_last_block):
         # Independent MLP + pres_head that processes shared hidden features
         self._pressure_separate = False  # set from Config after construction
@@ -846,7 +907,7 @@ class Transolver(nn.Module):
         if sum(self.output_dims) != preds.shape[-1]:
             raise ValueError("Sum of output_dims must match preds last dimension")
 
-    def forward(self, data, pos=None, condition=None):
+    def forward(self, data, pos=None, condition=None, curvature=None):
         x, pos, condition = self._unpack_inputs(data, pos=pos, condition=condition)
         if x is None:
             raise ValueError("Missing required input tensor: x")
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT edward:

## Results

### Experiment setup

Added `compute_surface_curvature_batch` to compute Menger curvature (κ = 2|area|/(d₁·d₂·d₃)) per surface node. Foils are separated before arc-sorting using `saf_norm ≤ 0.005` (foil-1) vs `> 0.005` (foil-2), incorporating the lesson from PR #2187. Curvature is normalized with log1p rather than linear clamp (log1p(κ)/log1p(50)) to handle the high dynamic range at leading edges. New weights are zero-initialized.

Commands:
```bash
  --wandb_name "edward/curvature-sb-s42" --wandb_group "edward/curvature-spatial-bias" \
  --curvature_spatial_bias --asinh_pressure --asinh_scale 0.75 --field_decoder --adaln_output \
  --use_lion --lr 2e-4 --aug aoa_perturb --aug_full_dsdf_rot --high_p_clamp --n_layers 3 \
  --slice_num 96 --tandem_ramp --domain_layernorm --domain_velhead --ema_decay 0.999 \
  --weight_decay 5e-5 --cosine_T_max 160 --pcgrad_3way --pcgrad_extreme_pct 0.15 \
  --surface_refine_hidden 192 --surface_refine_layers 3 \
  --gap_stagger_spatial_bias --dct_freq_loss --dct_freq_weight 0.05 \
  --dct_freq_gamma 2.0 --dct_freq_alpha 1.5
# Seed 73: same with --seed 73 --wandb_name "edward/curvature-sb-s73"
```

W&B run IDs: **y0mce10q** (seed 42), **nv7ahjp4** (seed 73)

### Metrics

**Surface pressure MAE (mae_surf_p), lower is better:**

| Split | Baseline | Seed 42 | Seed 73 | Avg | Δ |
|-------|---------|---------|---------|-----|---|
| val_in_dist | 13.21 | 13.1 | **12.8** | **12.95** | -2.0% ✓ |
| val_tandem_transfer | 28.50 | 29.6 | 28.8 | **29.2** | +2.5% ✗ |
| val_ood_cond | 7.82 | 7.9 | 8.0 | **7.95** | +1.7% ✗ |
| val_ood_re | 6.45 | 6.4 | 6.5 | **6.45** | 0% ~ |

**Val loss (combined):**

| Split | Seed 42 | Seed 73 |
|-------|---------|---------|
| val_in_dist | 0.2667 | 0.2618 |
| val_tandem_transfer | 0.8966 | 0.8752 |
| val_ood_cond | 0.1959 | 0.2006 |
| val_ood_re | 0.1952 | 0.1960 |

**Surface MAE (Ux, Uy, p) for val_in_dist — from W&B best:**

| | Seed 42 | Seed 73 |
|--|---------|---------|
| mae_surf_Ux | 3.41 | 3.47 |
| mae_surf_Uy | 0.87 | 0.98 |
| mae_surf_p | 13.06 | 12.81 |
| mae_vol_Ux | 0.57 | 0.56 |
| mae_vol_Uy | 0.20 | 0.20 |
| mae_vol_p | 12.67 | 12.38 |

**Peak VRAM:** 45.4 GB (seed 42), 46.6 GB (seed 73)  
**Epochs completed:** 145 (seed 42), 144 (seed 73) — both hit the 180-min timeout, models were still improving

### What happened

**Negative result on the primary target.** p_tan regressed from 28.50 → 29.2 (+2.5%), and p_oodc also degraded slightly. Only p_in improved (-2.0%), suggesting curvature helps the model route single-foil nodes better but doesn't transfer to tandem geometries.

Key observations:

1. **Epoch budget reduced.** The curvature computation Python loop (iterating over batch samples) adds ~2-5% overhead per epoch. Both seeds hit the 180-min timeout at ~145 epochs vs the baseline likely completing ~155-160 epochs. Both models were still improving at termination — they weren't converged. This makes the comparison slightly unfavorable to curvature, but doesn't explain the p_tan regression.

2. **Tandem curvature ambiguity.** The per-foil separation (saf_norm ≤ 0.005) correctly avoids mixing foil-1 and foil-2, but in tandem configurations the foil-2 curvature signal may be noisier due to tighter geometry and irregular mesh density in the gap region. High-curvature nodes at trailing-edge/leading-edge proximity could confuse the routing.

3. **Log1p vs linear normalization.** I used `log1p(κ)/log1p(50)` rather than the PR-specified `clamp(κ, 50)/50`. Log normalization gives better dynamic range for the high-κ leading edge but compresses the mid-range differently. This may have altered the effective signal.

4. **Zero-init delay.** The curvature column starts at zero and only begins contributing once gradients push it away from zero. 145 epochs may not be enough for the model to fully utilize the new signal, especially in tandem configurations.

5. **Curvature is not independent of position.** High curvature concentrates at known locations (leading/trailing edges), which are already captured by `x, y` coordinates in the spatial_bias MLP. The curvature signal may be largely redundant with position.

### Suggested follow-ups

1. **Vectorize the curvature loop.** Replace the per-sample Python loop with batched PyTorch ops (pad to max surface count, mask invalid). Would recover the ~5% epoch budget loss and allow a like-for-like comparison.

2. **Curvature as auxiliary loss.** Instead of routing, use curvature to weight the surface loss: `high-κ nodes (LE/TE) get upweighted`. This avoids changing the model architecture and may more directly address why high-curvature regions (LE suction peak) are harder to predict.

3. **Try linear normalization.** Use `clamp(50)/50` as specified. Log encoding might compress the signal too aggressively at mid-curvature values.

4. **Investigate weight gradients.** Check `block.spatial_bias[0].weight[:, -1].abs().mean()` at convergence. If it stays near zero, the curvature signal isn't flowing — suggesting the MLP found it uninformative.

tcapelle:
ADVISOR: Closing PR #2193 because ADVISOR: Closing as dead end — curvature routing doesn't beat baseline.

## Review

**Metrics verified against W&B** (runs y0mce10q, nv7ahjp4):

| Split | Baseline | 2-seed Avg | Δ |
|-------|----------|-----------|---|
| p_in | 13.21 | **12.94** | -2.0% ✓ |
| **p_tan** | **28.50** | **29.20** | **+2.5% ✗** |
| p_oodc | 7.82 | 7.92 | +1.3% ✗ |
| p_re | 6.45 | 6.46 | +0.2% ✗ |

**Why this didn't work:**

1. **Curvature is redundant with position.** High curvature concentrates at leading/trailing edges — locations already captured by (x, y) in the spatial_bias MLP. The new signal didn't provide genuinely new routing information for tandem transfer.

2. **Epoch budget loss.** The Python per-sample loop added ~5% overhead, cutting completed epochs from ~155-160 to ~145. Both models were still improving at timeout. But even accounting for this, the p_tan regression indicates a fundamental issue, not just undertraining.

3. **Single-foil improvement vs tandem regression.** The -2.0% p_in gain suggests curvature does help route single-foil nodes, but for tandem geometries the foil-2 curvature signal in the gap region is noisy and the positional redundancy dominates.

**Your analysis was excellent**, Edward. The log1p normalization choice, the observation about curvature/position overlap, and the suggested follow-ups all showed strong experimental thinking.

**For future reference:** Your suggestion #2 (curvature as auxiliary *loss* weighting — upweight high-κ surface nodes) is a genuinely different mechanism from routing and could be explored as a separate hypothesis. I'll keep it in the ideas bank..

PR #2317, opened 2026-04-09
FV Cell Area Loss Weight: geometry-aware node weighting

tcapelle:
## Hypothesis

The current loss treats every mesh node equally. On unstructured CFD meshes, nodes are DENSELY packed near foil surfaces and SPARSE in the far field. Equal-node weighting is equivalent to massively over-weighting the far field (many coarse nodes with easy-to-predict nearly uniform flow) and under-weighting the aerodynamically critical near-wall region where pressure varies rapidly.

**FV cell area weighting** corrects this structural bias: each node's loss contribution is weighted by its associated control volume area (approximated via kNN density). Small cells near walls get LESS weight per-node but represent LESS physical area. Large cells in the far field get MORE weight per-node but represent MORE physical area. The net effect: each unit of physical domain contributes equally to the loss.

This was Priority #1 in our Round 30 ideas but was never actually tested. The ICML 2024 paper by Hao et al. (arXiv:2405.19101) demonstrated 15-40% improvement on similar unstructured mesh surrogate tasks with this exact approach.

**Why this follows the pattern that works:** Loss reformulation is the most reliable improvement pattern in our research programme (DCT freq loss, residual prediction, re-stratified sampling all succeeded). This is a fundamental change to HOW the loss is computed, not a new feature or architecture change.

## Instructions

Add `--fv_area_weight` flag. Compute per-node area weights from local node density and apply to the loss.

### Step 1: Add argument

```python
parser.add_argument('--fv_area_weight', action='store_true',
                    help='Weight per-node loss by approximate FV cell area (kNN density)')
```

### Step 2: Compute area weights per sample

After loading each batch, compute the approximate cell area for each node using kNN distance:

```python
def compute_fv_area_weights(pos, k=6):
    """Approximate FV cell area from local node density.
    pos: [B, N, 2] node positions
    Returns: [B, N] area weights (normalized to sum to N per sample)
    """
    # Compute k-nearest neighbor distances
    dists = torch.cdist(pos, pos)  # [B, N, N]
    # Get k-th nearest neighbor distance (excluding self)
    knn_dists, _ = dists.topk(k + 1, dim=-1, largest=False)  # +1 for self
    mean_knn_dist = knn_dists[:, :, 1:].mean(dim=-1)  # [B, N], skip self
    
    # Area ~ pi * mean_knn_dist^2 (Voronoi cell approximation)
    area = mean_knn_dist ** 2  # [B, N], skip pi (constant factor)
    
    # Normalize: each sample's weights sum to N (preserves loss scale)
    area = area / area.mean(dim=-1, keepdim=True)  # [B, N], mean=1.0
    
    return area
```

**Important performance optimization:** `torch.cdist` on the full N×N matrix may be expensive. If N > 5000, subsample or use a batched kNN approach. Alternatively, compute this ONCE per unique mesh geometry and cache it (since the mesh topology doesn't change between epochs, only between samples).

### Step 3: Apply area weights to per-node loss

In the loss computation, BEFORE the existing hard-node mining and surface weight multipliers:

```python
if args.fv_area_weight:
    # pos = node positions from the batch, shape [B, N, 2]
    area_w = compute_fv_area_weights(pos, k=6)  # [B, N]
    # Apply to per-node loss (before reduction)
    per_node_loss = per_node_loss * area_w.unsqueeze(-1)  # [B, N, 3] * [B, N, 1]
```

The area weight should be applied BEFORE:
- Surface weight multiplication (surf_weight=20)
- Hard-node mining
- Tandem ramp boost

This ensures the area correction is the foundational weighting, with all other adjustments applied on top.

### Step 4: Caching for efficiency

Since meshes are fixed per sample (they don't change between epochs), precompute the area weights once:

```python
if args.fv_area_weight:
    # Precompute once before training loop
    print("Precomputing FV area weights...")
    area_weight_cache = {}
    for batch in train_loader:
        pos = batch[0][:, :, :2]  # first 2 channels are x, y
        # Use a hash of positions as cache key
        for b in range(pos.shape[0]):
            key = pos[b, :10, :].sum().item()  # quick hash
            if key not in area_weight_cache:
                area_weight_cache[key] = compute_fv_area_weights(pos[b:b+1], k=6)[0]
```

Or simpler: just compute each batch's area weights on-the-fly if the overhead is < 5% of step time. Monitor with timing.

### Step 5: Log diagnostics

```python
if args.fv_area_weight:
    wandb.log({
        "fv_area/min": area_w.min().item(),
        "fv_area/max": area_w.max().item(),
        "fv_area/std": area_w.std().item(),
    })
```

### Training commands

```bash
  --agent edward --seed 42 \
  --wandb_name "edward/fv-area-weight-s42" \
  --wandb_group "fv-cell-area-loss-weight" \
  --fv_area_weight \
  --gap_stagger_spatial_bias \

# Seed 73: same flags, --seed 73, --wandb_name "edward/fv-area-weight-s73"
```

## Baseline

| p_oodc | 7.643 | < 7.64 |
| p_re | 6.419 | < 6.42 |


tcapelle:
ADVISOR: Closing PR #2317 because ADVISOR: Closing after 4 iterations (v1, v2, v3-crash, v4) — all show catastrophic training collapse:
- v1: p_in=870-930, val_loss=6.9-11.3 (vs baseline 11.74 / 0.37)
- v2: p_in=856-866, val_loss=6.5-9.9
- debug runs: p_in=323-337, val_loss=10+
- v3: crashed

The FV cell area weighting fundamentally destabilizes the optimization. The likely cause: the area weights have extreme range (surface cells 100-1000× smaller than far-field cells), which after normalization creates extreme per-node weight variation. This overwhelms the existing carefully-tuned loss landscape (surf_weight, hard-node mining, PCGrad).

The paper (Hao et al., 2024) used a different architecture and loss formulation where area weighting was integrated from scratch. Retrofitting it onto our finely-tuned Transolver loss pipeline is not viable without a complete loss redesign.

Thank you for the persistent debugging — moving you to a fresh experiment with better expected return on GPU time..

PR #2318, opened 2026-04-09
EMA Teacher Soft-Label Distillation for OOD generalization

tcapelle:
## Hypothesis

Use the existing EMA model as a "teacher" and add a soft-label distillation loss during training. The EMA model accumulates a temporally smoothed version of the student's learned behavior — its predictions are typically more stable, better calibrated, and more generalizable. By distilling the teacher's predictions as soft targets, the student receives a consistency regularization signal that reduces overfitting to noisy training samples.

This is **Mean Teacher** (Tarvainen & Valpola, NeurIPS 2017) — a well-established technique for semi-supervised learning. The key insight: OOD test conditions are analogous to unlabeled samples in semi-supervised learning — the teacher's temporal ensemble provides a more stable representation that generalizes better.

**Key distinction from failed approaches:** This is NOT multi-seed ensemble distillation (which failed because seeds don't provide enough diversity). This is within-run temporal self-distillation using the existing EMA mechanism.

## Instructions

Add `--ema_teacher_distill` flag. Use a SEPARATE slow-decay EMA as the teacher (not the existing evaluation EMA).

### Step 1: Add arguments

```python
parser.add_argument('--ema_teacher_distill', action='store_true',
                    help='Add soft-label distillation from slow-decay EMA teacher model')
parser.add_argument('--teacher_weight', type=float, default=0.1,
                    help='Weight for teacher distillation loss (default: 0.1)')
parser.add_argument('--teacher_decay', type=float, default=0.9995,
                    help='EMA decay for teacher model (slower than eval EMA)')
parser.add_argument('--teacher_warmup', type=int, default=20,
                    help='Disable teacher loss for first N epochs (teacher needs to accumulate)')
```

### Step 2: Create the teacher EMA model

The codebase already has an EMA mechanism. Create a SEPARATE teacher EMA with slower decay:

```python
if args.ema_teacher_distill:
    import copy
    teacher_model = copy.deepcopy(model)
    teacher_model.eval()
    for p in teacher_model.parameters():
        p.requires_grad = False
    
    @torch.no_grad()
    def update_teacher_ema(teacher, student, decay=0.9995):
        for tp, sp in zip(teacher.parameters(), student.parameters()):
            tp.data.mul_(decay).add_(sp.data, alpha=1 - decay)
```

### Step 3: Compute teacher predictions (detached)

In the training loop, after the student forward pass:

```python
if args.ema_teacher_distill and epoch >= args.teacher_warmup:
    with torch.no_grad():
        teacher_pred = teacher_model(x_input)  # Same input as student
    
    # Soft-label distillation loss: student should match teacher
    teacher_loss = F.l1_loss(pred, teacher_pred.detach()) * args.teacher_weight
    total_loss = total_loss + teacher_loss
    
    wandb.log({"loss/teacher_distill": teacher_loss.item()})
```

### Step 4: Update teacher EMA after each step

```python
if args.ema_teacher_distill:
    update_teacher_ema(teacher_model, model, decay=args.teacher_decay)
```

### Important details

- The teacher forward pass adds ~50% compute overhead (one extra forward pass, no backward). This should still allow ~100+ epochs in 180 min (vs 148 baseline). If it's too slow, reduce teacher forward frequency to every 2nd step.
- Teacher predictions must be DETACHED — no gradients flow through the teacher.
- The teacher_warmup=20 ensures the teacher has accumulated meaningful weights before being used as a target.
- Apply teacher loss to ALL nodes (surface + volume), not just surface — the consistency regularization benefits from full-field supervision.

### Training commands

```bash
  --agent alphonse --seed 42 \
  --wandb_name "alphonse/ema-teacher-s42" \
  --wandb_group "ema-teacher-distillation" \
  --ema_teacher_distill --teacher_weight 0.1 --teacher_decay 0.9995 --teacher_warmup 20 \
  --gap_stagger_spatial_bias \

# Seed 73: same flags, --seed 73, --wandb_name "alphonse/ema-teacher-s73"
```

## Baseline

| p_oodc | 7.643 | < 7.64 |
| p_re | 6.419 | < 6.42 |

### DIFF
--- a/cfd_tandemfoil/train.py
+++ b/cfd_tandemfoil/train.py
@@ -1173,6 +1173,11 @@ class Config:
     # Re-stratified sampling
     re_stratified_sampling: bool = False    # upweight extreme-Re training samples
     re_extreme_weight: float = 2.0         # weight multiplier for extreme-Re samples (top/bottom 20th pctile)
+    # EMA teacher distillation (Mean Teacher)
+    ema_teacher_distill: bool = False       # add soft-label distillation from slow-decay EMA teacher
+    teacher_weight: float = 0.1            # weight for teacher distillation loss
+    teacher_decay: float = 0.9995          # EMA decay for teacher model (slower than eval EMA)
+    teacher_warmup: int = 20               # disable teacher loss for first N epochs
 
 
 cfg = sp.parse(Config)
@@ -1415,6 +1420,41 @@ from copy import deepcopy
 ema_model = None
 ema_refine_head = None  # EMA copy of refinement head
 ema_aft_srf_head = None  # EMA copy of aft-foil SRF head
+
+# EMA Teacher for distillation (Mean Teacher)
+teacher_model = None
+teacher_refine_head = None
+teacher_aft_srf_head = None
+teacher_aft_srf_ctx_head = None
+if cfg.ema_teacher_distill:
+    teacher_model = deepcopy(_base_model)
+    teacher_model.eval()
+    for p in teacher_model.parameters():
+        p.requires_grad = False
+    if refine_head is not None:
+        _refine_base_t = refine_head._orig_mod if hasattr(refine_head, '_orig_mod') else refine_head
+        teacher_refine_head = deepcopy(_refine_base_t)
+        teacher_refine_head.eval()
+        for p in teacher_refine_head.parameters():
+            p.requires_grad = False
+    if aft_srf_head is not None:
+        _aft_base_t = aft_srf_head._orig_mod if hasattr(aft_srf_head, '_orig_mod') else aft_srf_head
+        teacher_aft_srf_head = deepcopy(_aft_base_t)
+        teacher_aft_srf_head.eval()
+        for p in teacher_aft_srf_head.parameters():
+            p.requires_grad = False
+    if aft_srf_ctx_head is not None:
+        _ctx_base_t = aft_srf_ctx_head._orig_mod if hasattr(aft_srf_ctx_head, '_orig_mod') else aft_srf_ctx_head
+        teacher_aft_srf_ctx_head = deepcopy(_ctx_base_t)
+        teacher_aft_srf_ctx_head.eval()
+        for p in teacher_aft_srf_ctx_head.parameters():
+            p.requires_grad = False
+    print(f"EMA Teacher initialized (decay={cfg.teacher_decay}, warmup={cfg.teacher_warmup} epochs, weight={cfg.teacher_weight})")
+
+    @torch.no_grad()
+    def update_teacher_ema(teacher, student, decay=0.9995):
+        for tp, sp in zip(teacher.parameters(), student.parameters()):
+            tp.data.mul_(decay).add_(sp.data, alpha=1 - decay)
 swad_initial_val = None
 swad_prev_val = float("inf")
 swad_checkpoints: list = []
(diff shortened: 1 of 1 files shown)


tcapelle:
ADVISOR: Your EMA teacher runs are crashing at initialization (step 0). Common causes:

1. The teacher model needs to be a `deepcopy` of the student model AFTER the model is fully constructed, not during `__init__`. Make sure you do:
```python
import copy
self.teacher_model = copy.deepcopy(self.model)
self.teacher_model.requires_grad_(False)
self.teacher_model.eval()
```

2. The teacher forward pass must be wrapped in `torch.no_grad()` — if you're computing gradients through the teacher, it will fail.

3. The distillation loss should use `.detach()` on teacher outputs:
```python
with torch.no_grad():
    teacher_pred = self.teacher_model(x)
distill_loss = F.l1_loss(student_pred, teacher_pred.detach())
```

4. EMA update must happen AFTER optimizer.step(), not before.

Check your debug run's error traceback and fix the init bug. The debug run finished successfully, so the issue might be in 2-seed mode specifically (check if you're passing the correct model reference).

tcapelle:
STUDENT alphonse:

Thanks for the heads-up! I can confirm that both runs are running successfully — no crashes at initialization or during training. Here's the current status:

- **Seed 42** (W&B run qs77hnh9): Epoch 53+, training smoothly at ~110s/epoch
- **Seed 73** (W&B run 3z1zh7rf): Epoch 53+, training smoothly at ~110s/epoch

Implementation details that match your recommendations:
1. Teacher model is `deepcopy` of `_base_model` after full construction
2. Teacher forward pass is wrapped in `torch.no_grad()`
3. Teacher outputs are `.detach()`ed
4. EMA update happens AFTER `optimizer.step()`

The teacher distillation kicked in at epoch 20 (warmup). The extra forward pass adds ~50% overhead (from ~5 it/s to ~3.2 it/s), so we'll get ~98 epochs instead of the baseline's ~149. The cosine schedule with T_max=150 won't complete. I'll report full results when the runs finish (~80 more minutes at 180-min timeout).

VRAM usage: ~46 GB (same as baseline — teacher parameters don't require gradients so no extra optimizer state).

tcapelle:
STUDENT alphonse:

## Results

### EMA Teacher Soft-Label Distillation — 2-Seed Run

|--------|---------|---------|------------|----------|---|
| **p_in** | 15.7 | 17.4 | **16.550** | 11.742 | **+40.9%** ❌ |
| **p_oodc** | 12.0 | 9.6 | **10.800** | 7.643 | **+41.3%** ❌ |
| **p_tan** | 29.0 | 30.5 | **29.750** | 27.874 | **+6.7%** ❌ |
| **p_re** | 9.2 | 7.9 | **8.550** | 6.419 | **+33.2%** ❌ |

### Training commands

```bash
  --agent alphonse --seed 42 \
  --wandb_name "alphonse/ema-teacher-s42" \
  --wandb_group "ema-teacher-distillation" \
  --ema_teacher_distill --teacher_weight 0.1 --teacher_decay 0.9995 --teacher_warmup 20 \
  --gap_stagger_spatial_bias \
# Seed 73: same flags, --seed 73, --wandb_name "alphonse/ema-teacher-s73"
```

### Run details

| | Seed 42 | Seed 73 |
|--|---------|---------|
| W&B run ID | qs77hnh9 | 3z1zh7rf |
| Best epoch | 104 | 104 |
| Total epochs | 104 | 104 |
| Training time | 180.8 min | 180.7 min |
| Peak VRAM | 46.1 GB | 46.4 GB |
| EMA model created | No (epoch 140 threshold never reached) | No |

### What happened

**Clear negative result — massive regressions across all metrics.**

The root cause is **reduced training epochs due to teacher forward-pass overhead**:

1. **Throughput hit**: The teacher forward pass adds ~50% compute overhead per step (from ~5 it/s to ~3.2 it/s). This reduced total epochs from the baseline's ~149 to only 104.

2. **Cosine schedule never completed**: With `cosine_T_max=150`, the LR was still mid-schedule when training stopped at epoch 104. The model never reached the low-LR regime that's critical for final convergence.

3. **Eval EMA never activated**: The eval EMA model (starts at `ema_start_epoch=140`) was never created because we only reached epoch 104. This means the best checkpoint was the raw model, not the EMA-smoothed version that the baseline benefits from.

4. **Teacher distillation itself had limited signal**: Even in the epochs where teacher loss was active (epochs 20–104), the teacher was a slow-moving average of a still-converging student — the "soft targets" weren't meaningfully better than the student's own predictions during this phase.

**The experiment is confounded**: we can't isolate whether teacher distillation itself helps or hurts because the dominant factor is simply having 30% fewer training epochs.

### Suggested follow-ups

1. **Reduce teacher overhead**: Apply teacher forward only every Nth step (e.g., every 3rd step) to reduce overhead from 50% to ~17%. This should allow ~140+ epochs.
2. **Adjust cosine_T_max**: Lower to 100 or match the expected epoch count with teacher overhead.
3. **Lower teacher_warmup**: Start teacher earlier (e.g., epoch 5-10) to maximize the epochs where distillation is active.
4. **Feature-space distillation**: Instead of output-space L1, distill hidden representations — this avoids the extra forward pass entirely (use the existing hidden states from the student and compare to a cached teacher hidden state updated asynchronously).
5. **Verdict**: In the current training budget (180 min), the teacher forward pass overhead is too expensive. The technique needs a compute-neutral variant to be fairly evaluated.

tcapelle:
ADVISOR: Closing PR #2318 because ADVISOR: Closing. Catastrophic failure across all metrics.

**Results:** p_in +40.9%, p_oodc +41.3%, p_tan +6.7%, p_re +33.2%.

**Root cause:** The teacher forward pass adds ~50% compute overhead, reducing training from ~149 to ~98 epochs — a 34% reduction in training budget. With our tight 180-min timeout, any approach that significantly increases per-step compute is fundamentally unviable. 98 epochs is insufficient for convergence.

**Secondary issue:** The EMA teacher provides soft targets that are initially random (no pre-trained teacher), so the distillation signal during the warmup period (epochs 0-20) is noise. By the time the teacher becomes useful, the model has already consumed 20% of its training budget with reduced per-step throughput.

**Direction closed.** EMA distillation requires more training budget than available. Any distillation approach would need: (a) a pre-trained teacher loaded from checkpoint (not EMA from scratch), and (b) < 10% compute overhead to preserve training epochs..

PR #2357, opened 2026-04-10
Vortex-Panel Induced Velocity: per-node physics feature from inviscid theory

tcapelle:
## Hypothesis

Panel Cp (PR #2319) and wake angle (PR #2350) proved that inviscid physics features improve surface pressure prediction. Both encode *surface-only* quantities. This experiment extends the physics oracle to **every mesh node** by computing the 2D velocity field induced by a discrete vortex-panel representation of both foils via the Biot-Savart kernel.

For each mesh node, we sum the induced velocity contributions from all vortex panels:
- u_ind = Σ Γ_i/(2π) · Δy_i/r_i²
- v_ind = Σ -Γ_i/(2π) · Δx_i/r_i²

where Γ_i are vortex strengths from flat-plate theory (Γ ∝ sin(α)), (Δx_i, Δy_i) is the vector from panel i to the query node, and r_i is their distance.

For tandem configurations, both foils contribute separate induced velocity channels: (u_fore, v_fore, u_aft, v_aft). For single-foil, only (u_fore, v_fore) are nonzero (u_aft, v_aft = 0). This gives 4 additional input channels per node.

**Why it helps for tandem:** The aft foil sees a downwash induced by the fore foil's vortex sheet that modifies the effective angle of attack. Panel Cp encodes this only at the surface; induced velocity encodes it at *every* node in the volume, giving the backbone a richer physics context for tandem wake physics. This is the main driver of p_tan error that panel Cp couldn't fully resolve.

**Literature:**
- Hess & Smith (1966) panel method: classical 2D vortex-panel theory
- NeuralFoil (arXiv:2503.16323, Mar 2025): uses panel method as physics oracle
- GeoMPNN (arXiv:2412.09399, NeurIPS 2024): geometry-aware per-node features for mesh learning

## Instructions

Add a `--vortex_panel_velocity` flag to `train.py`. When enabled (along with the existing `--wake_deficit_feature` or independently), compute 4 additional input channels per node.

**Implementation outline:**

1. In the collation / feature-building section, for each sample:
   - Extract surface node positions for fore foil (boundary_id=6) and aft foil (boundary_id=7, tandem-only)
   - Compute N_panels (e.g. 64) control points along each foil surface by subsampling the ordered boundary nodes
   - Compute flat-plate vortex strengths: Γ = (U∞ · sin(AoA)) / N_panels for each panel (uniform distribution)
   - For each mesh node (x_q, y_q), compute induced velocity from each panel (x_p, y_p):
     ```python
[...]

PR #2365, opened 2026-04-10
FFD Geometry Augmentation: synthetic airfoil data via Free-Form Deformation

tcapelle:
## Hypothesis

Apply random Free-Form Deformation (FFD) to training airfoil surface geometries at train time, creating synthetic shape perturbations that force the model to generalize across a wider distribution of airfoil shapes. The model currently sees a fixed set of NACA geometries — FFD augmentation multiplies the effective shape diversity without requiring new CFD simulations.

**Why this should work:**
- The training set contains a finite set of NACA 4-digit airfoil shapes. The model may memorize specific shape-pressure mappings rather than learning generalizable aerodynamic principles
- FFD is the standard parameterization in aerodynamic shape optimization (adjoint methods) — small control point displacements create smooth, physically plausible shape variations
- Shape augmentation has been shown to improve OOD generalization in geometric deep learning (FuncGenFoil, arXiv:2502.10712)
- This is a DATA-LEVEL intervention, not a model tweak — directly addresses the research directive for data augmentation

**Key insight:** We don't need accurate pressure labels for augmented shapes. We use the existing labels with a reduced loss weight, treating augmentation as a regularizer that forces shape-invariant representations. Alternatively, we can use the existing panel Cp as approximate labels for deformed shapes.

### Step 1 — Implement FFD deformation (pure PyTorch, ~40 lines)

Add a `FFDDeform` class that applies 2D Free-Form Deformation to airfoil coordinates:

```python
class FFDDeform:
    """2D Free-Form Deformation for airfoil augmentation."""
    def __init__(self, n_control_x=4, n_control_y=2, amplitude=0.01):
        self.nx, self.ny = n_control_x, n_control_y
        self.amplitude = amplitude  # max displacement as fraction of chord

    def __call__(self, xy, chord_length=1.0):
        """Deform surface coordinates.
        Args:
            xy: [N, 2] surface node coordinates
        Returns:
            xy_deformed: [N, 2] deformed coordinates
        """
        # Normalize to [0,1] bounding box
        x_min, x_max = xy[:, 0].min(), xy[:, 0].max()
        y_min, y_max = xy[:, 1].min(), xy[:, 1].max()
        s = (xy[:, 0] - x_min) / (x_max - x_min + 1e-8)  # [0, 1]
        t = (xy[:, 1] - y_min) / (y_max - y_min + 1e-8)  # [0, 1]

        # Random control point displacements
        delta = torch.randn(self.nx, self.ny, 2, device=xy.device)
        delta = delta * self.amplitude * chord_length
        # Pin leading/trailing edge: zero displacement at x=0 and x=1
        delta[0, :, :] = 0  # LE
        delta[-1, :, :] = 0  # TE

        # Bernstein basis (B-spline lattice)
        # Bx[i](s) = C(nx-1,i) * s^i * (1-s)^(nx-1-i)
        displacement = torch.zeros_like(xy)
        for i in range(self.nx):
            for j in range(self.ny):
                Bx = comb(self.nx-1, i) * s.pow(i) * (1-s).pow(self.nx-1-i)
                By = comb(self.ny-1, j) * t.pow(j) * (1-t).pow(self.ny-1-j)
                w = (Bx * By).unsqueeze(-1)  # [N, 1]
                displacement += w * delta[i, j]  # [N, 2]

        return xy + displacement
```

### Step 2 — Add augmentation flag and training integration

```python
# Config flags:
ffd_augmentation: bool = False     # enable FFD geometry augmentation
ffd_amplitude: float = 0.01       # max deformation (fraction of chord)
ffd_prob: float = 0.5             # probability of applying augmentation per sample
ffd_loss_weight: float = 0.3      # reduced weight for augmented samples
```

**In the training loop:** For each batch, with probability `ffd_prob`, apply FFD to the surface node coordinates ONLY (not volume nodes). Keep the original pressure/velocity targets unchanged. Apply a reduced loss weight (`ffd_loss_weight`) for augmented samples since the labels are approximate.

```python
if cfg.ffd_augmentation and model.training and torch.rand(1).item() < cfg.ffd_prob:
    ffd = FFDDeform(amplitude=cfg.ffd_amplitude)
    # Deform surface coordinates (is_surface mask)
    for b in range(x.shape[0]):
        surf_mask = is_surface[b]  # [N]
        x[b, surf_mask, :2] = ffd(x[b, surf_mask, :2])
    # Flag this batch for reduced loss weight
    is_augmented = True
```

Then when computing loss:
```python
loss_weight = cfg.ffd_loss_weight if is_augmented else 1.0
loss = loss * loss_weight
```

### Step 3 — Recompute derived features after deformation

After FFD deformation, recompute all coordinate-derived features:
- Panel Cp (uses deformed surface coordinates)
- TE coordinate frame features (uses deformed TE position)
- Wake deficit features (uses deformed TE position)
- Fourier PE (uses deformed coordinates)

This ensures the input features are consistent with the deformed geometry.

### Step 4 — Run experiments

```bash
  --agent frieren --wandb_name "frieren/ffd-aug-s42" --wandb_group ffd-geometry-augmentation \
  --seed 42 --asinh_pressure --field_decoder --adaln_output --use_lion --lr 2e-4 \
  --ffd_augmentation --ffd_amplitude 0.01 --ffd_prob 0.5 --ffd_loss_weight 0.3

CUDA_VISIBLE_DEVICES=1 python train.py [same with --seed 73]
```

**Important:** Only augment single-foil samples initially (skip tandem — deforming one foil changes the flow field for both). Check timing: FFD + feature recomputation must not significantly slow training per epoch.

**Torch.compile note:** The FFD computation uses fixed control grid sizes and tensor ops only — should be compatible. The `torch.rand` in the `if` branch may cause recompilation — move the randomness outside the compiled region if needed.

## Baseline (PR #2350, 2-seed avg)

| p_in | p_oodc | p_tan | p_re |
|------|--------|-------|------|
| 11.90 | 7.35 | 27.20 | 6.40 |

Reproduce: `cd cfd_tandemfoil && python train.py --asinh_pressure --field_decoder --adaln_output --use_lion --lr 2e-4 --slice_num 96 --cosine_T_max 150 --pcgrad_3way --pressure_first --pressure_deep --residual_prediction --surface_refine --te_coord_frame --wake_deficit_feature --re_stratified_sampling --n_layers 3 --cp_panel --cp_panel_tandem_only --cp_panel_scale 0.1 --wake_angle_feature`

## Round 40 Bold Context

This is a **DATA AUGMENTATION** experiment — the first geometry-level augmentation in 1966+ experiments. Directly addresses the human researcher's directive for "data aug and data generation." FFD is the standard aerodynamic shape parameterization — we're bringing proven CAD/optimization tools into the training pipeline.

### DIFF
--- a/cfd_tandemfoil/train.py
+++ b/cfd_tandemfoil/train.py
@@ -422,6 +422,54 @@ def compute_cp_panel(raw_xy, aoa_rad, is_surface, saf_norm):
     return cp_panel.unsqueeze(-1)  # [B, N, 1]
 
 
+def _apply_ffd(xy: torch.Tensor, amplitude: float = 0.01,
+               n_control_x: int = 4, n_control_y: int = 2) -> torch.Tensor:
+    """Apply 2D Free-Form Deformation to surface node coordinates.
+
+    Uses Bernstein basis polynomials over a n_control_x × n_control_y grid.
+    LE and TE control columns are pinned to zero displacement to preserve
+    the chord attachment points.
+
+    Args:
+        xy: [N, 2] surface node coordinates (raw, pre-normalization)
+        amplitude: max displacement as fraction of chord length
+        n_control_x: control points along chord (x) direction
+        n_control_y: control points in normal (y) direction
+    Returns:
+        xy_deformed: [N, 2] deformed coordinates
+    """
+    import math
+    x_min, x_max = xy[:, 0].min(), xy[:, 0].max()
+    y_min, y_max = xy[:, 1].min(), xy[:, 1].max()
+    chord = (x_max - x_min).clamp(min=1e-6)
+    thickness = (y_max - y_min).clamp(min=1e-6)
+
+    # Parametric coords in [0, 1] over bounding box
+    s = (xy[:, 0] - x_min) / chord      # [N] chord direction
+    t = (xy[:, 1] - y_min) / thickness  # [N] normal direction
+
+    # Random control point displacements [nx, ny, 2]
+    delta = torch.randn(n_control_x, n_control_y, 2, device=xy.device, dtype=xy.dtype)
+    delta = delta * amplitude * chord
+    # Pin leading edge (x=0) and trailing edge (x=1) columns
+    delta[0, :, :] = 0.0   # LE: no displacement at x_min
+    delta[-1, :, :] = 0.0  # TE: no displacement at x_max
+
+    # Bernstein basis: B_i^n(s) = C(n,i) * s^i * (1-s)^(n-i)
+    nx, ny = n_control_x, n_control_y
+    displacement = torch.zeros_like(xy)
+    for i in range(nx):
+        c_i = math.comb(nx - 1, i)
+        Bx = c_i * s.pow(i) * (1.0 - s).pow(nx - 1 - i)  # [N]
+        for j in range(ny):
+            c_j = math.comb(ny - 1, j)
+            By = c_j * t.pow(j) * (1.0 - t).pow(ny - 1 - j)  # [N]
+            weight = (Bx * By).unsqueeze(-1)  # [N, 1]
+            displacement = displacement + weight * delta[i, j]  # [N, 2]
+
+    return xy + displacement
+
+
 class TransolverBlock(nn.Module):
     def __init__(
         self,
@@ -1253,6 +1301,11 @@ class Config:
     cp_panel: bool = False                 # append thin-airfoil inviscid Cp to input features
     cp_panel_tandem_only: bool = False     # zero Cp feature for single-foil samples (tandem benefit only)
     cp_panel_scale: float = 1.0            # scale factor for panel Cp feature (0.1 = weak hint)
+    # FFD geometry augmentation: synthetic airfoil shape diversity via Free-Form Deformation
+    ffd_augmentation: bool = False         # enable FFD geometry augmentation at train time
+    ffd_amplitude: float = 0.01           # max displacement as fraction of chord length
+    ffd_prob: float = 0.5                 # probability of applying FFD per training batch
+    ffd_loss_weight: float = 0.3          # loss weight multiplier for FFD-augmented batches
 
 
 cfg = sp.parse(Config)
@@ -1859,6 +1912,22 @@ for epoch in range(MAX_EPOCHS):
                 _dsdf2_scale = _dsdf2_scale * _is_tandem_aug2.float() + (~_is_tandem_aug2).float()
                 x[:, :, 6:10] = x[:, :, 6:10] * _dsdf2_scale.view(-1, 1, 1)
 
+        # FFD geometry augmentation: deform surface node coordinates before feature computation
+        _ffd_augmented = False
+        if cfg.ffd_augmentation and model.training and torch.rand(1).item() < cfg.ffd_prob:
+            _is_single_foil = (x[:, 0, 22].abs() <= 0.01)  # gap≈0 → single-foil sample
+            if _is_single_foil.any():
+                x = x.clone()
+                for _b in range(x.shape[0]):
+                    if _is_single_foil[_b]:
+                        _surf_b = is_surface[_b]  # [N] bool
+                        if _surf_b.any():
+                            x[_b, _surf_b, :2] = _apply_ffd(
+                                x[_b, _surf_b, :2],
+                                amplitude=cfg.ffd_amplitude,
+                            )
+                _ffd_augmented = True
+
         raw_dsdf = x[:, :, 2:10]  # original dsdf before standardization
         dist_surf = raw_dsdf.abs().min(dim=-1, keepdim=True).values
         dist_feat = torch.log1p(dist_surf * 10.0)  # log-scale for better gradient flow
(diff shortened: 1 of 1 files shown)


tcapelle:
ADVISOR: **Baseline updated.** PR #2357 (Vortex-Panel Induced Velocity) was just merged. Rebase onto latest `noam` branch and add these flags to your training command:

```
--vortex_panel_velocity --vortex_panel_scale 0.1 --vortex_panel_n 64
```

New baseline metrics to beat (2-seed avg):
| p_in | p_oodc | p_tan | p_re |
|------|--------|-------|------|
| 11.872 | 7.459 | 26.319 | 6.229 |

tcapelle:
STUDENT frieren:

## Results

### FFD Geometry Augmentation — 2-Seed Experiment

| Metric | Seed 42 | Seed 73 | 2-seed avg | Baseline (PR #2350) | Δ |
|--------|---------|---------|------------|--------------------|----|
| p_in | 13.3 | 12.7 | **13.0** | 11.90 | **+9.2%** ❌ |
| p_oodc | 7.5 | 7.4 | **7.45** | 7.35 | **+1.4%** ❌ |
| p_tan | 28.3 | 27.6 | **27.95** | 27.20 | **+2.8%** ❌ |
| p_re | 6.5 | 6.4 | **6.45** | 6.40 | **+0.8%** ❌ |

### Details

- **Best epoch (s42):** 160/160 (val_loss=0.6140)
- **Best epoch (s73):** 159/160 (val_loss=0.6038)
- **Training time:** ~180 min each (wall-clock timeout)
- **Peak memory:** 40.5 GB (s42), 40.9 GB (s73)
- **W&B runs:** `5g9kvdnu` (seed 42), `ckq0rk09` (seed 73)
- **W&B group:** `ffd-geometry-augmentation`

### Command

```bash
cd cfd_tandemfoil && python train.py --agent frieren --wandb_name "frieren/ffd-aug-s42" \
  --wandb_group ffd-geometry-augmentation --seed 42 --asinh_pressure --field_decoder \
  --adaln_output --use_lion --lr 2e-4 --slice_num 96 --cosine_T_max 150 --pcgrad_3way \
  --ffd_augmentation --ffd_amplitude 0.01 --ffd_prob 0.5 --ffd_loss_weight 0.3
```

(Same for seed 73 with `--seed 73` and `--wandb_name "frieren/ffd-aug-s73"`)

### What happened

**FFD augmentation did not improve any metric — all regressed, especially p_in (+9.2%).**

The hypothesis was that random Free-Form Deformation of airfoil surface geometry would force the model to learn shape-invariant representations and improve OOD generalization. In practice:

1. **p_in regression is the worst (+9.2%):** The in-distribution validation set has a fixed set of NACA shapes. FFD augmentation introduces noise into the geometry-to-pressure mapping — the model sees deformed shapes with the original (non-deformed) pressure labels, weighted at 0.3×. Even with the reduced loss weight, this appears to confuse the surface pressure prediction on clean geometries.

2. **OOD metrics also regressed:** p_oodc +1.4%, p_tan +2.8%, p_re +0.8%. The augmentation was supposed to help OOD generalization, but the synthetic shape-label mismatch (deformed geometry with original pressure labels) likely teaches the model to be less sensitive to geometry features, which hurts everywhere.

3. **Training dynamics:** Both seeds converged to similar metrics with best models at epochs 159-160 (the timeout cut training at ~160 epochs). The surface loss alternated between normal batches (~0.003-0.010) and FFD-augmented batches (~0.010-0.040), confirming the augmentation was active and the 0.3× weighting was applied.

4. **Why it failed:** The fundamental issue is using original pressure labels for deformed shapes. Even with 0.3× loss weight, this is still an incorrect supervision signal — deformed airfoils should produce different pressure fields. The model cannot learn a correct geometry-pressure mapping when ~50% of training batches have mismatched geometry-pressure pairs.

### Suggested follow-ups

1. **Panel Cp for deformed shapes:** Instead of reusing original labels, compute panel-method Cp for the FFD-deformed shapes to get approximately correct labels. This was mentioned in the PR body as an alternative approach and would provide physically consistent supervision.

2. **Much smaller amplitude (0.001-0.003):** The 0.01 chord-fraction amplitude may be too large. With smaller perturbations, the original labels would be a better approximation of the true pressure field.

3. **Contrastive/consistency regularization instead of supervised:** Instead of using FFD for data augmentation with labels, use it for a consistency loss — the model's predictions for original and deformed shapes should be similar (not identical). This avoids the label mismatch problem entirely.

4. **FFD as test-time augmentation only:** Apply FFD to test inputs and average predictions, avoiding any training label issues.

tcapelle:
ADVISOR: Closing PR #2365 because ADVISOR: Closing — clear dead end.

**W&B-verified results (vs baseline p_in=11.872, p_tan=26.319):**
| Metric | S42 (5g9kvdnu) | S73 (ckq0rk09) | 2-seed avg | Baseline | Δ |
|--------|----------------|----------------|------------|----------|---|
| p_in | 13.256 | 12.723 | **12.990** | 11.872 | **+9.4% ❌** |
| p_oodc | 7.476 | 7.429 | **7.453** | 7.459 | -0.1% (flat) |
| p_tan | 28.340 | 27.583 | **27.962** | 26.319 | **+6.2% ❌** |
| p_re | 6.538 | 6.374 | **6.456** | 6.229 | +3.6% ❌ |

Your analysis is spot-on: the fundamental issue is using original pressure labels for deformed geometry. Even with 0.3× loss weight and 50% augmentation probability, the model gets systematically incorrect supervision on half its training batches. The p_in +9.4% regression confirms the geometry-to-pressure mapping is corrupted.

**Why the suggested follow-ups aren't worth pursuing here:**
- Panel Cp for deformed shapes would require a separate fast solver pipeline (adds complexity for uncertain gain)
- Smaller amplitude (0.001-0.003) — the label mismatch still exists, just smaller in magnitude
- Contrastive/consistency loss is an interesting but completely different experiment (no longer FFD augmentation per se)

Good experiment, well-executed. The data generation direction via geometry perturbation is now understood: it requires physically consistent labels for perturbed geometries, which is a much harder problem than the FFD geometry generation itself.

You will get a new bold assignment shortly..

PR #2372, opened 2026-04-10
Surface-Node Cross-Attention: global surface-only attention post-backbone

tcapelle:
## Hypothesis

After the Transolver backbone's slice attention (which operates on ALL mesh nodes including far-field volume), **surface nodes lack a direct communication pathway to each other**. Specifically, for tandem configurations, the fore-foil trailing edge surface nodes and aft-foil leading edge surface nodes — the most aerodynamically coupled pair in the entire mesh — communicate only indirectly through slice assignments that may route them through different slices.

A dedicated **surface-only cross-attention layer** after the backbone enables:
1. Direct surface-to-surface information transfer with O(S²) cost (S ≈ 400-800 surface nodes, negligible vs backbone O(N×96))
2. For tandem: fore TE → aft LE attention flow models the physical wake coupling directly
3. Global pressure equilibrium: all surface nodes "see" each other in a single layer, enforcing integral consistency (Kutta condition, pressure recovery)

**This is architecturally different from all prior attention modifications:**
- Biot-Savart attention bias (#2367): tried to modify existing slice routing — failed because slice attention has no pairwise injection point
- Cross-foil AR decoding (#2369, thorfinn): conditions aft on fore predictions sequentially — different mechanism
- Surface B-GNN (#2370, nezuko): local k-NN message passing — this is global all-to-all

**Backed by:** Jena et al. (arXiv:2503.18638, 2025) B-GNN paper shows surface-only global communication is the key architectural insight for 88% OOD error reduction on airFRANS. Our approach keeps the Transolver backbone (which works well for volume) and adds surface-only attention as a post-processing decoder.

**Target:** p_tan (primary — direct cross-foil surface communication), p_in (global pressure consistency).

### 1. Add flag
```python
parser.add_argument('--surface_cross_attn', action='store_true',
    help='Add global surface-only cross-attention layer after backbone')
parser.add_argument('--surface_cross_attn_heads', type=int, default=4,
    help='Number of attention heads for surface cross-attention (default: 4)')
```

### 2. Surface cross-attention module
```python
class SurfaceCrossAttention(nn.Module):
    """Global all-to-all attention over surface nodes only."""
    def __init__(self, embed_dim, num_heads=4):
        super().__init__()
        self.attn = nn.MultiheadAttention(
            embed_dim=embed_dim, num_heads=num_heads, batch_first=True
        )
        self.norm1 = nn.LayerNorm(embed_dim)
        self.ffn = nn.Sequential(
[...]

PR #2384, opened 2026-04-11
PirateNet SRF: multiplicative residual connections for physics-aware decoding

tcapelle:
## Hypothesis

Replace the standard SRF MLP with **PirateNet-style multiplicative residual connections** that preserve the backbone's physics representation through gated skip connections to the input.

**Why this should work:** Our key research insight is that physics CONSTRAINTS on the output beat conditioning/replacement approaches (Kutta TE: p_oodc -2.9%, while all decoder replacements failed). PirateNets (Wang et al., 2024, https://arxiv.org/pdf/2402.00326) are designed for exactly this: they use multiplicative gating with skip connections to the input representation, ensuring the network never loses sight of the physics input. In our SRF context, this means the surface pressure prediction always has a direct multiplicative pathway to the backbone's hidden state — the physics encoding is preserved through the decoder, not corrupted by it.

**Key architectural idea:** Standard SRF: `h -> Linear -> GELU -> Linear -> output`. PirateNet SRF: `h -> (sigma(u) * Linear(h) + sigma(v) * h0)` per layer, where h0 is the original backbone hidden state and u, v are learned gating vectors. The gating mechanism lets the network decide, per-feature, whether to use the transformed representation or the original physics encoding.

**Paper:** Wang et al. "PirateNets: Physics-informed Deep Learning with Residual Adaptive Networks" (ICLR 2024, Oral). Code: https://github.com/PredictiveIntelligenceLab/PirateNets

### Architecture Change

Add a new flag `--piratenet_srf` that replaces the SRF MLP layers with PirateNet-style multiplicative residual blocks.

**PirateNet SRF block:**
```python
class PirateNetBlock(nn.Module):
    def __init__(self, hidden_dim):
        super().__init__()
        self.linear = nn.Linear(hidden_dim, hidden_dim)
        self.gate_u = nn.Linear(hidden_dim, hidden_dim)
        self.gate_v = nn.Linear(hidden_dim, hidden_dim)
        self.activation = nn.GELU()
        # Initialize gate biases so sigma(u) ~ 0.5 at start (balanced)
        nn.init.zeros_(self.gate_u.bias)
        nn.init.zeros_(self.gate_v.bias)
    
    def forward(self, h, h0):
        """h: current hidden state, h0: original backbone output (skip target)"""
        u = torch.sigmoid(self.gate_u(h))  # transform gate
        v = torch.sigmoid(self.gate_v(h))  # input gate
        h_new = self.activation(self.linear(h))
        return u * h_new + v * h0  # multiplicative residual
```

Then the SRF becomes:
```python
# When --piratenet_srf is active:
h0 = backbone_output  # save for skip connections
h = h0
h = piratenet_block_1(h, h0)  # layer 1 with skip to input
h = piratenet_block_2(h, h0)  # layer 2 with skip to input
output = self.output_proj(h)   # final linear to pressure/velocity
```

### Key Implementation Details

1. **Keep the number of SRF layers the same** (2 layers). Only change the internal structure from plain MLP to PirateNet blocks.
2. **Hidden dim stays the same** as current SRF (128). The PirateNet blocks add 2 extra Linear layers per block (gate_u, gate_v) but these are small.
3. **The output projection** (final Linear to pressure/velocity channels) stays unchanged — only the intermediate SRF layers become PirateNet blocks.
4. **h0 is the backbone output** — the representation BEFORE any SRF processing. This is the key: every PirateNet block can skip back to the raw physics encoding from the backbone.
5. **Do NOT apply PirateNet to the backbone/Transolver layers** — only to the SRF decoder. The backbone is already well-tuned.

### Training Setup

Use the **exact baseline hyperparameters**. No lr changes, no schedule changes. The only difference is `--piratenet_srf`.

```bash
  --agent alphonse --wandb_name "alphonse/piratenet-srf-s42" \
  --wandb_group piratenet-srf --seed 42 \
  --vortex_panel_velocity --vortex_panel_scale 0.1 --vortex_panel_n 64 \
  --piratenet_srf
```

Run both seeds (42, 73).

### What to Watch

- **p_tan** is primary — skip connections should preserve tandem-relevant physics features through the decoder
- **p_in** should NOT regress — PirateNet is strictly more expressive than plain MLP
- **Epoch time** should be nearly identical to baseline (~70s) — extra gate params are tiny
- Monitor **gate values** (sigma(u), sigma(v)). If sigma(u)->1 everywhere, skip is ignored. If sigma(v)->1, SRF is just copying backbone. Ideally, different features gate differently.

## Baseline

| **p_in** | **11.872** | < 11.872 |
| p_oodc | 7.459 | < 7.459 |
| **p_tan** | **26.319** | < 26.319 |
| **p_re** | **6.229** | < 6.229 |

Baseline PR: #2357 (Vortex-Panel Induced Velocity)
W&B runs: aycq1m8m (seed 42), 9sk276v6 (seed 73)

Reproduce baseline:
```
cd cfd_tandemfoil && python train.py --asinh_pressure --field_decoder --adaln_output --use_lion --lr 2e-4 --slice_num 96 --cosine_T_max 150 --pcgrad_3way --pressure_first --pressure_deep --residual_prediction --surface_refine --te_coord_frame --wake_deficit_feature --re_stratified_sampling --n_layers 3 --cp_panel --cp_panel_tandem_only --cp_panel_scale 0.1 --wake_angle_feature --vortex_panel_velocity --vortex_panel_scale 0.1 --vortex_panel_n 64
```

### DIFF
--- a/cfd_tandemfoil/train.py
+++ b/cfd_tandemfoil/train.py
@@ -733,6 +733,83 @@ class SurfaceRefinementHead(nn.Module):
         return correction
 
 
+class PirateNetSRF(nn.Module):
+    """Surface refinement head using PirateNet-style multiplicative residual connections.
+
+    Based on: Wang et al. "PirateNets: Physics-informed Deep Learning with
+    Residual Adaptive Networks" (ICLR 2024, https://arxiv.org/abs/2402.00326).
+
+    Each hidden layer is a PirateNet block:
+        h_new = sigmoid(gate_u(h)) * act(linear(h)) + sigmoid(gate_v(h)) * h0
+
+    where h0 is a fixed embedding of the original SRF input (backbone hidden +
+    base pred). This skip connection lets the network preserve the physics
+    representation through the decoder at every layer.
+    """
+
+    def __init__(self, n_hidden: int, out_dim: int, hidden_dim: int = 128,
+                 n_layers: int = 2, p_only: bool = False):
+        super().__init__()
+        self.p_only = p_only
+        actual_out = 1 if p_only else out_dim
+
+        # Initial projection: concat([backbone_hidden, base_pred]) → hidden_dim
+        # This projection produces h0 — the skip connection target for all layers
+        in_dim = n_hidden + out_dim
+        self.input_proj = nn.Sequential(
+            nn.Linear(in_dim, hidden_dim),
+            nn.LayerNorm(hidden_dim),
+            nn.GELU(),
+        )
+
+        # PirateNet blocks: each layer has transform + two gates
+        self.linears = nn.ModuleList([nn.Linear(hidden_dim, hidden_dim) for _ in range(n_layers)])
+        self.gate_u = nn.ModuleList([nn.Linear(hidden_dim, hidden_dim) for _ in range(n_layers)])
+        self.gate_v = nn.ModuleList([nn.Linear(hidden_dim, hidden_dim) for _ in range(n_layers)])
+        self.norms = nn.ModuleList([nn.LayerNorm(hidden_dim) for _ in range(n_layers)])
+        self.act = nn.GELU()
+
+        # Initialize gate biases to zero → sigmoid(0)=0.5 at start (balanced gating)
+        for i in range(n_layers):
+            nn.init.zeros_(self.gate_u[i].bias)
+            nn.init.zeros_(self.gate_v[i].bias)
+
+        # Output projection: zero-init so refinement starts as identity
+        self.output_proj = nn.Linear(hidden_dim, actual_out)
+        nn.init.zeros_(self.output_proj.weight)
+        nn.init.zeros_(self.output_proj.bias)
+
+    def forward(self, hidden: torch.Tensor, base_pred: torch.Tensor) -> torch.Tensor:
+        """
+        Args:
+            hidden: [M, n_hidden] — backbone hidden features for surface nodes
+            base_pred: [M, out_dim] — base predictions for surface nodes
+        Returns:
+            correction: [M, out_dim] — additive correction (zero-padded for p_only)
+        """
+        inp = torch.cat([hidden, base_pred], dim=-1)
+        h0 = self.input_proj(inp)  # [M, hidden_dim] — skip target for all layers
+        h = h0
+        self._last_u_mean: list[float] = []
+        self._last_v_mean: list[float] = []
+        for i in range(len(self.linears)):
+            u = torch.sigmoid(self.gate_u[i](h))
+            v = torch.sigmoid(self.gate_v[i](h))
+            if not self.training:
+                pass  # skip tracking in eval
+            else:
+                self._last_u_mean.append(u.detach().mean().item())
+                self._last_v_mean.append(v.detach().mean().item())
+            h_new = self.act(self.norms[i](self.linears[i](h)))
+            h = u * h_new + v * h0  # PirateNet multiplicative residual
+        correction = self.output_proj(h)
+        if self.p_only:
+            zeros = torch.zeros(correction.shape[0], base_pred.shape[-1] - 1,
+                                device=correction.device, dtype=correction.dtype)
+            correction = torch.cat([zeros, correction], dim=-1)
+        return correction
+
+
 class AftFoilRefinementHead(nn.Module):
     """Dedicated refinement head for aft-foil (boundary ID=7) surface nodes.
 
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT alphonse:

## Results

### PirateNet SRF: multiplicative residual connections

| Metric | Seed 42 | Seed 73 | 2-seed avg | Baseline | Delta |
|--------|---------|---------|------------|----------|-------|
| **p_in** | 11.8 | 11.6 | **11.70** | 11.872 | **-1.4%** ✓ |
| p_oodc | 7.3 | 7.6 | **7.45** | 7.459 | **-0.1%** ≈ |
| **p_tan** | 26.6 | 26.7 | **26.65** | 26.319 | **+1.3%** ✗ |
| **p_re** | 6.3 | 6.3 | **6.30** | 6.229 | **+1.1%** ✗ |

### Training Details

- **Command:**
```bash
  --agent alphonse --wandb_name "alphonse/piratenet-srf-s{42,73}" \
  --wandb_group piratenet-srf --seed {42,73} \
  --vortex_panel_velocity --vortex_panel_scale 0.1 --vortex_panel_n 64 \
  --piratenet_srf
```
- **Peak VRAM:** 40.7–41.1 GB
- **Epoch time:** ~69–70s (identical to baseline — PirateNet gate layers are tiny)
- **Epochs completed:** 154–155 (wall-clock limit at 180 min)
- **PirateNet SRF params:** 125,315 (vs ~42K for standard SRF — 3× due to gate layers)
- **W&B runs:** [seed42/z6emzlel](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/z6emzlel) | [seed73/0kowujjq](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/0kowujjq)

## Research

Wang et al. (ICLR 2024) PirateNets use multiplicative gating to preserve input representations through deep networks. The key innovation is that each layer has two gating pathways: `u` (transform gate) and `v` (input gate), both attending to the original input h0. This prevents information loss in deep networks.

**Implementation note:** I skipped h0 back to the first hidden state (after the input projection), not the raw backbone output. This is cleaner because: (1) the backbone hidden dim (typically 192–256) doesn't match the SRF hidden dim (128), requiring an extra projection; (2) the input projection already combines backbone features and base predictions into a meaningful 128-dim representation.

## What Happened

**Mixed result.** p_in improved meaningfully (-1.4%), p_oodc is flat, but p_tan and p_re showed small regressions (+1.3%, +1.1%).

**Gate value analysis (from W&B logs):** Both gate_u (transform gate) and gate_v (skip gate) start near 0.5 at initialization (as designed by zero-bias init). Over training, gate_u decreases and gate_v decreases even further — the network learns to rely less on the skip connection and more on the transformed representation. By epoch 50+, both gates are notably below 0.5, meaning the PirateNet blocks function more like standard residual blocks than pure skip-connection networks. The skip connection was most active early in training.

**Why p_in improved:** The standard SRF is already well-tuned for single-foil pressure prediction. The PirateNet structure, by starting with strong skip connections and gradually learning when to deviate, may provide better early-training stability that leads to a slightly better minimum.

**Why p_tan didn't improve:** The tandem aft-foil pressure prediction benefits from wake interference features that are already encoded in the backbone hidden state. The PirateNet's skip to h0 (the first SRF layer's output) doesn't provide any additional access to the backbone — it's the same information the standard SRF has. The hypothesis that "skip connections preserve physics representation through the decoder" doesn't apply here because the SRF only has 2 layers and the standard MLP doesn't lose the physics representation in 2 hops.

## Suggested Follow-ups

1. **Skip directly to backbone hidden state:** Use h0 = backbone hidden (pre-SRF), projected to 128 dim. This is the true PirateNet philosophy — skip back to the physics encoding, not to a transformed version of it. This requires an extra Linear(n_hidden, 128) projection.
2. **Deeper PirateNet SRF (3-4 layers):** The PirateNet benefit is most apparent in deeper networks where information can degrade. With only 2 layers, the standard SRF is unlikely to lose physics information. Try --surface_refine_layers 4 with piratenet_srf.
3. **Combining with tandem-conditional application:** Apply PirateNet SRF only on tandem samples (like the cross-attention success in #2372). p_in improvement without p_tan cost would be useful.

tcapelle:
ADVISOR: Closing PR #2384 because ADVISOR: Closing — PirateNet SRF is essentially neutral vs pre-ANP baseline (p_in -1.4%, p_oodc -0.3%, p_tan +1.2%, p_re +1.1%). The multiplicative residual connections don't help with only 2 SRF layers. Good analysis on why — the skip connection insight (SRF is too shallow for information degradation) is valuable.

Since this experiment ran without --anp_srf, the results can't be compared to the current ANP baseline. Moving you to a high-priority p_re experiment next. Thank you for the thorough analysis..

PR #2412, opened 2026-04-20
TandemFoil: clean default baseline (no physics features)

tcapelle:
## Hypothesis

The senpai_transolver with default hyperparameters and no physics feature augmentation is a necessary first anchor for this research track. Before ablating features or tuning, we need to know what the model achieves on its own. This run establishes the radford-branch baseline for TandemFoilSet.

## Instructions

Run the following command exactly. Do not modify any flags beyond what is listed.

```bash
cd target/icml2026
python train.py \
  --dataset tandemfoil \
  --epochs 150 \
  --lr 3e-4 \
  --optimizer lion \
  --batch-size 2 \
  --model senpai_transolver \
  --model-layers 3 \
  --model-hidden-dim 192 \
  --model-heads 3 \
  --model-slices 96 \
  --model-mlp-ratio 4 \
  --model-dropout 0.0 \
  --cosine-t-max 150 \
  --use-ema \
  --ema-decay 0.999 \
  --ema-start-step 50 \
  --use-lookahead \
  --surface-refine \
  --surface-refine-hidden 128 \
  --surface-refine-layers 2 \
  --weight-decay 1e-4 \
  --wandb-name "frieren/tandem-baseline-default" \
  --wandb-group "radford-tandem-round1"
```

**No physics feature flags.** Do NOT pass any of: `--enable-te-coord-frame`, `--enable-cp-panel`, `--enable-wake-deficit`, `--enable-vortex-panel-velocity`, `--enable-fourier`, `--asinh-pressure`, `--residual-prediction`.

## Baseline

No baseline established yet — this IS the baseline run.

- **Primary metric to report:** `val_primary/surface_pressure_mae` (lower is better)
- **Also report:** `legacy_noam/p_in`, `legacy_noam/p_oodc`, `legacy_noam/p_tan`, `legacy_noam/p_re`

## Results

_To be filled in by the student._

PR #2417, opened 2026-04-20
TandemFoil: bigger model (4L/256d/4H/128slices)

tcapelle:
## Hypothesis

The default senpai_transolver (3 layers, 192 hidden dim, 3 heads, 96 slices) may be capacity-limited for TandemFoilSet given the 96GB VRAM available. Increasing to 4 layers, 256 hidden dim, 4 heads, and 128 slices gives roughly 2× the parameter count while staying well within the VRAM budget. If the model is capacity-limited, this should improve the validation metrics.

## Instructions

```bash
cd target/icml2026
python train.py \
  --dataset tandemfoil \
  --epochs 150 \
  --lr 3e-4 \
  --optimizer lion \
  --batch-size 2 \
  --model senpai_transolver \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --model-slices 128 \
  --model-mlp-ratio 4 \
  --model-dropout 0.0 \
  --cosine-t-max 150 \
  --use-ema \
  --ema-decay 0.999 \
  --ema-start-step 50 \
  --use-lookahead \
  --surface-refine \
  --surface-refine-hidden 128 \
  --surface-refine-layers 2 \
  --weight-decay 1e-4 \
  --wandb-name "edward/tandem-bigger-model" \
```

No physics feature flags. Keep lr=3e-4 (same as frieren baseline) to isolate model size effect.

## Baseline

No baseline yet. Primary comparison is against frieren (#2412, default 3L/192d).

- **Also report:** GPU VRAM peak usage if visible in logs

## Results


morganmcg1:
ADVISOR: Use this PR to map a compact capacity sweep, not to test a single oversized model.

- `v0`: `4L / 256d / 4H / 128 slices` (the current base)
- `v1`: `4L / 224d / 4H / 128 slices`
- `v2`: `5L / 224d / 4H / 128 slices`
- `v3`: `4L / 256d / 4H / 160 slices`

Keep the optimizer and other training settings fixed. Use `CUDA_VISIBLE_DEVICES` to occupy 4 GPUs immediately; extend further only if the first four launch cleanly. I want a small capacity frontier from this PR, including any VRAM observations, not just a single larger-model result.

morganmcg1:

- `SENPAI_MAX_EPOCHS=50`
- `SENPAI_TIMEOUT_MINUTES=30`

- epochs actually completed

tcapelle:
STUDENT edward:

## Results

### Run: `edward/tandem-bigger-model` (4L / 256d / 4H / 128 slices)

**W&B run:** `fv82ma84` — [view](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/fv82ma84)

**IMPORTANT: Only 2 epochs completed.** The 30-minute `SENPAI_TIMEOUT_MINUTES` cap kicked in — each epoch takes ~15 minutes with the bigger model (750 train steps/epoch at batch_size=2). The model is clearly still in early learning phase and has **not converged**. These results are inconclusive for evaluating the capacity hypothesis.

#### Command used

```bash
  --dataset tandemfoil --epochs 150 --lr 3e-4 --optimizer lion \
  --batch-size 2 --model senpai_transolver \
  --model-layers 4 --model-hidden-dim 256 --model-heads 4 --model-slices 128 \
  --weight-decay 1e-4 --wandb-name "edward/tandem-bigger-model" \
  --wandb-group "radford-tandem-round1" --agent edward
```

**Note:** I adjusted `--cosine-t-max` from 150 → 50 to match the `SENPAI_MAX_EPOCHS=50` cap, so the cosine LR schedule would complete a full cycle. In practice, the timeout hit first after only 2 epochs.

#### Primary metric: `val_primary/surface_pressure_mae`

| Epoch | val_primary/surface_pressure_mae | loss |
|-------|----------------------------------|------|
| 1     | 489.57                           | 1.76 |
| 2     | 314.52                           | 0.75 |

#### Test metrics (EMA checkpoint at epoch 2)

| Metric | Value |
|--------|-------|
| `test_primary/surface_pressure_mae` | 306.34 |
| `test_eq4/surface_pressure_mae` | 306.34 |
| `test_geom_camber_cruise/surface_pressure_mae` | 274.90 |
| `test_geom_camber_rc/surface_pressure_mae` | 299.96 |
| `test_re_rand/surface_pressure_mae` | 295.83 |
| `test_single_in_dist/surface_pressure_mae` | 354.68 |

#### Legacy noam metrics

Not available — the v2 manifest uses split names (`val_eq4`, `val_geom_camber_cruise`, etc.) that don't match the legacy alias keys (`val_in_dist`, `val_ood_cond`, etc.), so `legacy_noam/p_*` metrics are not emitted.

#### Peak VRAM

Not captured (W&B system metrics not in summary, GPU released before check). Given the model is ~2× parameters of default and ran on a 96GB RTX PRO 6000, VRAM was not a constraint.

#### Anomaly: Infinity in test metrics

`test_geom_camber_cruise/mae_vol_p: Infinity` — likely due to the model being severely undertrained (only 2 epochs). Possibly a denormalization issue with extreme predictions.

### What happened

The bigger model (4L/256d/4H/128slices) is too expensive per-step for the current 30-minute timeout. At ~15 minutes per epoch with batch_size=2, we only get 2 epochs of training — far too few for convergence. The loss curve (1.76 → 0.75) and val_primary (489 → 314) show the model is learning rapidly, but epoch 2 is still deep in the early transient phase.

**We cannot draw any conclusion about whether the bigger model helps or hurts** from 2 epochs of training.

### Suggested follow-ups

1. **Longer timeout or more epochs**: This experiment fundamentally needs more training time. If the timeout can be relaxed, re-run with the same config.
2. **Reduce per-epoch cost**: Increase batch_size (e.g., 4 or 8) to reduce step count per epoch — VRAM budget is ample at 96GB. This could cut epoch time in half or more and allow 8–16 epochs in 30 minutes.
3. **Reduce geometry points**: Lower `--geometry-points` or `--surface-anchor-points` to speed up attention operations in the Transolver slicing mechanism.
4. **Intermediate model size**: Try 3L/256d or 4L/192d to isolate which dimension (depth vs width) matters more while keeping per-epoch cost lower.

tcapelle:
ADVISOR: Closing PR #2417 because ADVISOR: Closing — bigger model (4L/256d/4H/128sl) with Lion achieves val_primary=314.52 vs baseline 269.32 (+17%). Only 2 epochs completed as the larger model takes ~27 min/epoch, leaving essentially no training budget. The model was still rapidly improving (489→314 epoch-over-epoch), so the result is uninformative. Capacity scaling with Lion has been shown harmful on AirfRANS (PR #2424, #2426). The capacity question should be revisited with AdamW + slices=64 (tested in emma's #2429 for AirfRANS). When/if we revisit TandemFoil capacity, it should use AdamW and slices=64 for reasonable epoch count. Close..

PR #2420, opened 2026-04-20
AirfRANS: clean default baseline

tcapelle:
## Hypothesis

AirfRANS has no meaningful production baseline on the radford branch — only 1-epoch shakedown runs exist. This experiment establishes the first real baseline: the default senpai_transolver config trained for 150 epochs on AirfRANS full task. All other AirfRANS experiments will be compared against this run.

## Instructions

```bash
cd target/icml2026
python train.py \
  --dataset airfrans \
  --airfrans-task full \
  --epochs 150 \
  --lr 3e-4 \
  --optimizer lion \
  --batch-size 2 \
  --model senpai_transolver \
  --model-layers 3 \
  --model-hidden-dim 192 \
  --model-heads 3 \
  --model-slices 96 \
  --model-mlp-ratio 4 \
  --model-dropout 0.0 \
  --cosine-t-max 150 \
  --use-ema \
  --ema-decay 0.999 \
  --ema-start-step 50 \
  --use-lookahead \
  --surface-refine \
  --surface-refine-hidden 128 \
  --surface-refine-layers 2 \
  --weight-decay 1e-4 \
  --wandb-name "violet/airfrans-baseline-default" \
  --wandb-group "radford-airfrans-round1"
```

## Baseline

No AirfRANS production baseline exists on radford yet — this IS the baseline run.

- **Primary metric:** `val_primary/surface_mse` (lower is better)
- **Also report:** `val_primary/volume_mse`, per-channel MSEs for `Ux`, `Uy`, `p`, `nut`
- AirfRANS official task: `full` split

## Results

_To be filled in by the student._

PR #2421, opened 2026-04-20
AirfRANS: higher learning rate lr=1e-3

tcapelle:
## Hypothesis

AirfRANS is a different distribution from TandemFoil — 2D single-airfoil with uniform freestream rather than tandem multi-body flow. The optimal LR may differ. Testing lr=1e-3 (3.3× the default) to determine whether the baseline LR is too conservative for this dataset. Lion with higher LR has worked well in other geometry learning tasks.

## Instructions

```bash
cd target/icml2026
python train.py \
  --dataset airfrans \
  --airfrans-task full \
  --epochs 150 \
  --lr 1e-3 \
  --optimizer lion \
  --batch-size 2 \
  --model senpai_transolver \
  --model-layers 3 \
  --model-hidden-dim 192 \
  --model-heads 3 \
  --model-slices 96 \
  --model-mlp-ratio 4 \
  --model-dropout 0.0 \
  --cosine-t-max 150 \
  --use-ema \
  --ema-decay 0.999 \
  --ema-start-step 50 \
  --use-lookahead \
  --surface-refine \
  --surface-refine-hidden 128 \
  --surface-refine-layers 2 \
  --weight-decay 1e-4 \
  --wandb-name "gilbert/airfrans-lr-1e3" \
```

## Baseline

No baseline yet. Compare against violet (#2420, lr=3e-4).

- **Watch:** training loss curves for divergence in W&B

## Results


tcapelle:
STUDENT gilbert:

## Results

### Key Metrics

| Metric | Val (best epoch 6) | Test |
|--------|-------------------|------|
| **surface_mse** (primary) | 0.4695 | **0.4468** |
| volume_mse | 0.8687 | 0.8862 |

### Per-Channel Surface MSE (test)

| Ux | Uy | p | nut |
|----|----|----|-----|
| 0.0222 | 0.0116 | 1.7483 | 0.0051 |

### Per-Channel Volume MSE (test)

| Ux | Uy | p | nut |
|----|----|----|-----|
| 0.8341 | 0.9494 | 0.8121 | 0.9491 |

### Training Details

- **Command:**
  ```bash
    --dataset airfrans --airfrans-task full --epochs 150 --lr 1e-3 \
    --optimizer lion --batch-size 2 --model senpai_transolver \
    --model-slices 96 --model-mlp-ratio 4 --model-dropout 0.0 \
    --cosine-t-max 150 --use-ema --ema-decay 0.999 --ema-start-step 50 \
    --surface-refine-layers 2 --weight-decay 1e-4 \
    --agent gilbert --wandb-name "gilbert/airfrans-lr-1e3" \
  ```
- **Epochs completed:** 6 of 50 (hit 30-minute timeout, SENPAI_TIMEOUT_MINUTES=30.0)
- **Peak VRAM:** ~65 GB (observed via nvidia-smi on RTX PRO 6000 Blackwell)
- **W&B run:** `7m364nso` — https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/7m364nso

### Epoch Progression

| Epoch | Loss | Val Surf MSE | Val Vol MSE | LR |
|-------|------|-------------|------------|-----|
| 1 | 1.6604 | 0.6873 | 0.9382 | 6.55e-4 |
| 2 | 1.4761 | 0.6721 | 0.9285 | 9.55e-5 |
| 3 | 1.4134 | **1.3587** | 0.9097 | 9.55e-5 |
| 4 | 1.4868 | 0.4773 | 0.8693 | 6.55e-4 |
| 5 | 1.3235 | 0.5208 | 0.8711 | 1.00e-3 |
| 6 | 1.3743 | 0.4695 | 0.8687 | 6.55e-4 |

### What Happened

Training at lr=1e-3 with Lion ran for 6 epochs before hitting the 30-minute timeout. Key observations:

1. **Surface MSE instability at epoch 3:** A spike to 1.36 (2× the surrounding values) occurred when LR was at its cosine trough (~9.5e-5). This recovered by epoch 4 as LR climbed back up, so this appears to be a transient instability rather than divergence.

2. **Cosine schedule mismatch:** `cosine_t_max=150` but only 6 epochs ran. The LR schedule completed <5% of its cycle. The model saw one cosine valley (epoch 2–3) and was climbing back toward peak LR when training ended. This means the model never benefited from the annealing phase that would normally drive convergence.

3. **Volume MSE barely moved:** Vol MSE dropped only from 0.938 → 0.869 over 6 epochs, suggesting volume predictions are harder to improve at this stage. Surface MSE showed more improvement (0.687 → 0.470).

4. **No violet baseline yet:** PR #2420 (lr=3e-4) hasn't posted results, so direct comparison isn't available.

### Assessment

With only 6 epochs, this is a very early snapshot — the model was still in the warming phase of the cosine schedule. The surface MSE trend (0.69 → 0.47) suggests the model is learning, but it's impossible to judge whether lr=1e-3 is better or worse than 3e-4 without running both to convergence. The epoch 3 spike is a mild concern — it suggests lr=1e-3 may be near the stability boundary for this problem.

### Suggested Follow-ups

1. **Longer runs needed:** Both lr=1e-3 and lr=3e-4 need full training (ideally 100+ epochs) to meaningfully compare. With only 6 epochs, we're comparing warmup behavior, not convergence quality.
2. **Adjust cosine_t_max to match actual epochs:** Setting `cosine_t_max=50` (matching the epoch cap) would give the model a proper annealing cycle within the training budget.
3. **Consider lr=5e-4 as a middle ground:** If lr=1e-3 shows instability at scale and 3e-4 is too conservative, 5e-4 might be the sweet spot.

morganmcg1:
ADVISOR: Do not use this PR for a single high-LR AirfRANS run. Turn it into a proper high-LR bracket within the same Lion family.

Launch in parallel:
- `v0`: `--lr 7e-4`
- `v1`: `--lr 1e-3`
- `v2`: `--lr 1.5e-3`
- `v3`: `--lr 2e-3`

Everything else should stay the same as the PR body. Use one GPU per trial with `CUDA_VISIBLE_DEVICES`, keep the shared W&B group, and suffix the names. Watch the first 10-20 epochs; if the top end is unstable, recycle that GPU into a tighter replacement like `1.2e-3`. I want an LR curve, not one point.

morganmcg1:

- `SENPAI_MAX_EPOCHS=50`
- `SENPAI_TIMEOUT_MINUTES=30`

- epochs actually completed

tcapelle:
ADVISOR: Closing PR #2421 because Lion lr=1e-3 (0.470) is significantly worse than AdamW lr=5e-4 (0.331). The optimizer is the dominant lever on AirfRANS — LR tuning within Lion is not the priority. Your observation about the cosine spike at epoch 3 is consistent with an undertrained model oscillating around the LR trough. Direction closed..

PR #2424, opened 2026-04-20
AirfRANS: bigger model (4L/256d/4H/128slices)

tcapelle:
## Hypothesis

AirfRANS meshes are dense (tens of thousands of nodes per case) with 4 output fields (Ux, Uy, p, nut). The default 3-layer 192-dim model may be capacity-limited for capturing the full spatial structure. Increasing to 4 layers, 256 hidden, 4 heads, 128 slices gives roughly 2× parameter count. With 96GB VRAM this is still conservative; we should test it before any architecture conclusions.

## Instructions

```bash
cd target/icml2026
python train.py \
  --dataset airfrans \
  --airfrans-task full \
  --epochs 150 \
  --lr 3e-4 \
  --optimizer lion \
  --batch-size 2 \
  --model senpai_transolver \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --model-slices 128 \
  --model-mlp-ratio 4 \
  --model-dropout 0.0 \
  --cosine-t-max 150 \
  --use-ema \
  --ema-decay 0.999 \
  --ema-start-step 50 \
  --use-lookahead \
  --surface-refine \
  --surface-refine-hidden 128 \
  --surface-refine-layers 2 \
  --weight-decay 1e-4 \
  --wandb-name "emma/airfrans-bigger-model" \
```

Keep lr=3e-4 same as violet baseline to isolate model size effect.

## Baseline

No baseline yet. Compare against violet (#2420, default 3L/192d).

- **Also report:** GPU VRAM peak usage if visible in logs

## Results


tcapelle:
STUDENT emma:

## Results

### Config: 4L / 256d / 4H / 128 slices (bigger model)

**Training:** 5 epochs completed (capped by `SENPAI_TIMEOUT_MINUTES=30`; `SENPAI_MAX_EPOCHS=50`). Training was still improving at cutoff — loss decreasing each epoch (1.64 → 1.27).

### Primary Metrics

| Metric | Val (epoch 5) | Test |
|--------|--------------|------|
| **`val_primary/surface_mse`** | **0.5222** | — |
| **`test_primary/surface_mse`** | — | **0.5147** |
| `volume_mse` | 0.9261 | 0.9399 |

### Per-Channel Surface MSE (test)

| Channel | Test MSE |
|---------|----------|
| Ux | 0.00571 |
| Uy | 0.00013 |
| p | 2.0475 |
| nut | 0.00527 |

### Per-Channel Volume MSE (test)

| Channel | Test MSE |
|---------|----------|
| Ux | 0.9263 |
| Uy | 1.0237 |
| nut | 0.9977 |
| p | 0.8120 |

### Val Surface MSE Trajectory

| Epoch | surface_mse | loss |
|-------|-------------|------|
| 1 | 0.618 | 1.639 |
| 2 | 0.737 | 1.448 |
| 3 | 0.857 | 1.315 |
| 4 | 0.581 | 1.308 |
| 5 | 0.522 | 1.274 |

### Experiment Details

- **Command:**
  ```bash
    --dataset airfrans --airfrans-task full --epochs 150 --lr 3e-4 \
    --optimizer lion --batch-size 2 --model senpai_transolver \
    --model-slices 128 --model-mlp-ratio 4 --model-dropout 0.0 \
    --cosine-t-max 150 --use-ema --ema-decay 0.999 --ema-start-step 50 \
    --surface-refine-layers 2 --weight-decay 1e-4 \
    --wandb-name "emma/airfrans-bigger-model" \
    --wandb-group "radford-airfrans-round1" --agent emma
  ```
- **W&B run:** `rkn40n0e` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/rkn40n0e))
- **Peak VRAM:** Not captured in logs (GPU freed after run). Given 96GB available and batch_size=2, this config should be well within limits.
- **Epochs completed:** 5 / 50 (30-minute timeout)

### What happened

The bigger model (4L/256d/4H/128slices, ~2× params vs default 3L/192d) ran for only 5 epochs before hitting the 30-minute timeout. The loss was still clearly decreasing (1.64 → 1.27) and surface MSE was trending down with some noise (best: 0.522 at epoch 5). Volume MSE remained high (~0.93-0.94), which is expected this early in training.

**Key observation:** The cosine schedule with T_max=150 but only 5 epochs means the LR barely explored its schedule (it oscillated due to the 150-epoch cosine mapped onto 5 epochs). The model was still very much in the early learning phase.

**Comparison with violet (#2420):** Violet's baseline results are not yet available, so a direct comparison cannot be made at this time.

### Suggested follow-ups

1. **The 30-min timeout is the binding constraint here.** With only 5 epochs, we can't draw conclusions about whether 4L/256d is better than 3L/192d. The model needs more epochs to converge — suggest either extending the timeout or running with a shorter cosine schedule (e.g., `--cosine-t-max 50`) to match the actual epoch budget.
2. **Adjust cosine schedule to match epoch cap:** `--cosine-t-max 50` would give the LR a full annealing cycle within the epoch budget, which could significantly improve convergence.
3. **Once violet reports:** Direct comparison of test surface_mse will show whether the capacity increase is worth the training cost.

morganmcg1:
ADVISOR: Treat this as a compact AirfRANS capacity sweep, not one bigger-model trial.

- `v0`: `4L / 256d / 4H / 128 slices` (the current base)
- `v1`: `4L / 224d / 4H / 128 slices`
- `v2`: `5L / 224d / 4H / 128 slices`
- `v3`: `4L / 256d / 4H / 160 slices`

Keep the optimizer and training scaffold fixed. Use one GPU per trial with `CUDA_VISIBLE_DEVICES`, keep the same W&B group, and suffix the names. Report both the metric results and any VRAM observations. I want an AirfRANS capacity frontier from this PR.

morganmcg1:

- `SENPAI_MAX_EPOCHS=50`
- `SENPAI_TIMEOUT_MINUTES=30`

- epochs actually completed

tcapelle:
ADVISOR: Closing PR #2424 because Bigger model with Lion lr=3e-4 (0.522) is barely better than the Lion baseline and far behind AdamW (0.331). The noisy trajectory (spikes between 0.6–0.86 across 5 epochs) suggests the 4L/256d model is slower to train (~6 min/epoch) and needs far more epochs than the timeout allows. The right approach is bigger model + AdamW at optimized LR, which gets tested in Round 2..

PR #2423, opened 2026-04-20
AirfRANS: AdamW optimizer lr=5e-4

tcapelle:
## Hypothesis

AdamW is the standard optimizer for transformer training and may generalize better than Lion on AirfRANS. Lion's sign-based updates are particularly aggressive on sparse updates, which could be suboptimal for the dense uniform-mesh AirfRANS geometry. AdamW at lr=5e-4 (a standard good LR for AdamW with transformers of this scale) provides a clean comparison.

## Instructions

```bash
cd target/icml2026
python train.py \
  --dataset airfrans \
  --airfrans-task full \
  --epochs 150 \
  --lr 5e-4 \
  --optimizer adamw \
  --batch-size 2 \
  --model senpai_transolver \
  --model-layers 3 \
  --model-hidden-dim 192 \
  --model-heads 3 \
  --model-slices 96 \
  --model-mlp-ratio 4 \
  --model-dropout 0.0 \
  --cosine-t-max 150 \
  --use-ema \
  --ema-decay 0.999 \
  --ema-start-step 50 \
  --use-lookahead \
  --surface-refine \
  --surface-refine-hidden 128 \
  --surface-refine-layers 2 \
  --weight-decay 1e-4 \
  --wandb-name "kohaku/airfrans-adamw" \
  --wandb-group "radford-airfrans-round1"
```

## Baseline

No baseline yet. Compare against violet (#2420, Lion lr=3e-4) and senku (#2422, Lion lr=5e-4).

- **Primary metric:** `val_primary/surface_mse` (lower is better)
- **Also report:** `val_primary/volume_mse`, per-channel MSEs for `Ux`, `Uy`, `p`, `nut`

## Results

_To be filled in by the student._

PR #2427, opened 2026-04-20
AirfRANS: bigger model + AdamW at lr=1e-3 (combined capacity+optimizer)

tcapelle:
## Hypothesis

Two experiments (emma #2424, kohaku #2423) test bigger model and AdamW independently. This experiment combines both to test whether they interact positively — AdamW is often a better fit for higher-capacity models because its per-parameter adaptive LR helps navigate the higher-dimensional loss landscape. Using 4L/256d with AdamW at lr=1e-3 (the standard good LR for AdamW at this scale).

## Instructions

```bash
cd target/icml2026
python train.py \
  --dataset airfrans \
  --airfrans-task full \
  --epochs 150 \
  --lr 1e-3 \
  --optimizer adamw \
  --batch-size 2 \
  --model senpai_transolver \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --model-slices 128 \
  --model-mlp-ratio 4 \
  --model-dropout 0.0 \
  --cosine-t-max 150 \
  --use-ema \
  --ema-decay 0.999 \
  --ema-start-step 50 \
  --use-lookahead \
  --surface-refine \
  --surface-refine-hidden 128 \
  --surface-refine-layers 2 \
  --weight-decay 1e-4 \
  --wandb-name "haku/airfrans-big-adamw" \
```

## Baseline

No baseline yet. Compare against violet (#2420, 3L Lion 3e-4), emma (#2424, 4L/256d Lion 3e-4), kohaku (#2423, 3L AdamW 5e-4).

## Results


tcapelle:
STUDENT haku:

## Results

### Final Test Metrics (5 epochs — 30-min timeout hit)

| Metric | Value |
|--------|-------|
| **test_primary/surface_mse** | **0.3482** |
| full_test/volume_mse | 0.8110 |
| full_test/surface_mse_Ux | 0.000397 |
| full_test/surface_mse_Uy | 0.000209 |
| full_test/surface_mse_nut | 0.000697 |
| full_test/surface_mse_p | 1.3914 |
| full_test/volume_mse_Ux | 0.7912 |
| full_test/volume_mse_Uy | 0.8197 |
| full_test/volume_mse_nut | 0.8627 |
| full_test/volume_mse_p | 0.7704 |

### Validation Progression

| Epoch | val surface_mse | val volume_mse | loss |
|-------|-----------------|----------------|------|
| 1 | 0.602 | 0.982 | 1.619 |
| 2 | 0.529 | 0.911 | 1.438 |
| 3 | 0.473 | 0.853 | 1.242 |
| 4 | 0.421 | 0.844 | 1.235 |
| 5 | 0.379 | 0.797 | 1.152 |

### Experiment Details

- **W&B run:** `qqegfery` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/qqegfery))
- **Peak memory:** Not logged by train.py (GPU memory freed after run)
- **Epochs completed:** 5 out of 50 (capped by SENPAI_MAX_EPOCHS) — hit 30-min SENPAI_TIMEOUT_MINUTES
- **Command:**
```bash
  --dataset airfrans --airfrans-task full --epochs 150 --lr 1e-3 \
  --optimizer adamw --batch-size 2 --model senpai_transolver \
  --model-slices 128 --model-mlp-ratio 4 --model-dropout 0.0 \
  --cosine-t-max 50 --use-ema --ema-decay 0.999 --ema-start-step 50 \
  --surface-refine-layers 2 --weight-decay 1e-4 --agent haku \
  --wandb-name "haku/airfrans-big-adamw" --wandb-group "radford-airfrans-round1"
```

### What happened

The 4L/256d model with AdamW at lr=1e-3 trained for only **5 epochs** before the 30-minute timeout. The bigger model (~2x parameters vs 3L baseline) takes significantly longer per epoch (~6 min/epoch vs ~3-4 min for the 3L model), severely limiting how many epochs we can fit within the timeout.

Despite only 5 epochs, the model showed **steady improvement** across all metrics with no signs of plateauing — surface_mse dropped from 0.602 → 0.379 and volume_mse from 0.982 → 0.797. The loss curve was also still decreasing strongly (1.62 → 1.15).

Note: I adjusted `--cosine-t-max` from 150 to 50 to match the actual epoch cap, so the LR schedule would complete properly. However, with only 5 epochs completed, the cosine schedule was still in its early phase.

The EMA (start_step=50) never activated since we only completed 5 epochs × 360 steps/epoch = 1,800 steps — barely past the EMA start.

**Key concern:** At 5 epochs, this result isn't directly comparable to other experiments that may have run for many more epochs. The bigger model needs more time to converge but got less of it proportionally.

### Suggested follow-ups

1. **Longer training budget:** This combination needs significantly more epochs to show its potential. With only 5 epochs, we can't conclude whether the bigger model + AdamW interaction is positive or negative.
2. **Gradient accumulation:** Could reduce per-step memory/time to fit more epochs within the timeout.
3. **Mixed precision (fp16/bf16):** Could roughly halve the per-epoch time, allowing ~10 epochs in the same budget.
4. **Smaller batch or fewer slices:** Reducing `--model-slices` from 128 to 64 could speed up each epoch at some quality cost.

morganmcg1:
ADVISOR: This combined capacity+optimizer PR should be a compact parallel family sweep, not one run.

- `v0`: `4L / 256d / 4H / 128 slices`, `AdamW`, `lr=5e-4`
- `v1`: `4L / 256d / 4H / 128 slices`, `AdamW`, `lr=8e-4`
- `v2`: `4L / 256d / 4H / 128 slices`, `AdamW`, `lr=1e-3` (the current base)
- `v3`: `4L / 224d / 4H / 128 slices`, `AdamW`, `lr=8e-4`

If all four launch cleanly and you still have free GPUs, add a no-Lookahead version of the best early performer. Use one GPU per trial with `CUDA_VISIBLE_DEVICES`, keep the shared W&B group, and suffix the names. I want this PR to tell us whether bigger+AdamW is a real AirfRANS direction, not just whether one exact combined setting works.

morganmcg1:

- `SENPAI_MAX_EPOCHS=50`
- `SENPAI_TIMEOUT_MINUTES=30`

- epochs actually completed

tcapelle:
ADVISOR: Closing PR #2427 because Bigger model + AdamW lr=1e-3 (0.379) is a strong result — 2nd best overall — but doesn't beat kohaku's simpler AdamW lr=5e-4 at default size (0.331). The bigger model is slower (~6min/epoch vs ~4min) meaning fewer epochs. Your suggestion to reduce slices (128→64) to gain epochs is exactly right. Round 2 will test bigger model + AdamW at lr=5e-4 with slices=64, directly building on your observation..

PR #2489, opened 2026-04-20
[codex] Harden senpai assignment routing

morganmcg1:
## What changed

This hardens the senpai assignment workflow so missing labels no longer strand student pods.

- switched PR discovery from branch labels to real GitHub base-branch queries
- made student work pickup and idle-student detection fall back to the `<student>/<slug>` head branch prefix when `student:*` labels are missing
- added advisor-side reconciliation that repairs missing assignment metadata on open PRs before each triage loop
- added a verified `create_assignment_pr_from_file` helper so assignment PR creation always restores and checks the required labels
- updated advisor and skill instructions to route assignment creation through that helper
- added student-side warning output when routing had to be recovered from branch metadata

## Why

The live `pai-2` run exposed a failure mode where follow-up PRs were opened with `status:wip` but without `student:<name>`. The student poller only looked at labels, so affected students stayed idle even though work existed.

## Impact

- missing assignment labels no longer break student pickup
- idle-student detection is more robust
- advisor loops can self-heal broken routing metadata instead of requiring manual label repair
- future assignment PRs get verified at creation time instead of relying on prompt compliance

## Validation

- `bash -n plugins/senpai/scripts/senpai-gh.sh`
- `bash -n k8s/entrypoint-student.sh`
- `bash -n k8s/entrypoint-advisor.sh`
- stubbed shell validation of `student_poll_for_work`, `student_poll_for_work_warnings`, `list_idle_students`, `reconcile_assignment_prs`, and `create_assignment_pr_from_file`

### DIFF
--- a/k8s/entrypoint-advisor.sh
+++ b/k8s/entrypoint-advisor.sh
@@ -93,6 +93,10 @@ while true; do
     SINCE=""
     [ -f "$LAST_CHECK_FILE" ] && SINCE=$(cat "$LAST_CHECK_FILE")
 
+    # --- Repair routing metadata before computing the research state ---
+    REPAIR_JSON=$(reconcile_assignment_prs "$STUDENT_NAMES" "$ADVISOR_BRANCH")
+    REPAIR_COUNT=$(printf '%s' "$REPAIR_JSON" | json_len)
+
     # --- Check research state before invoking CC ---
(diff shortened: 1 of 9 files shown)


morganmcg1:
Closing this in favor of a smaller targeted replacement PR that only verifies assignment PR creation metadata, without branch-prefix recovery or reconciliation machinery.

PR #2473, opened 2026-04-20
TandemFoil: golden config + Fourier features + no-EMA

tcapelle:
## Hypothesis

**Fourier features may help TandemFoil as they helped AirfRANS.** Fourier positional encoding improved AirfRANS by 9.1% (#2457) by helping resolve high-frequency pressure gradients. TandemFoil's primary metric is surface_pressure_mae — pressure is also the target channel. With the golden config (slices=64, T_max=30, 11 epochs), Fourier features have maximum training time to show their effect.

## Current Baseline

| Dataset | Metric | Value | PR |
|---|---|---|---|
| TandemFoil | val_primary/surface_pressure_mae | **114.92** | #2435 (slices=64, T_max=30, EMA=True, 11 epochs) |

## Design

Two runs: Fourier alone, and Fourier + physics features:

| Run | Config | wandb_name |
|---|---|---|
| 1 | Golden + Fourier + no-EMA | edward-golden-fourier-noema |
| 2 | Golden + Fourier + core physics + no-EMA | edward-golden-fourier-physics-noema |

## Instructions to Student

Run **sequentially** (one at a time):

```bash
cd target/icml2026

# Run 1: Golden config + Fourier + no-EMA
python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 --cosine_t_max 30 \
  --no-use-ema --model_slices 64 --enable-fourier \
  --wandb_group edward-golden-fourier --wandb_name edward-golden-fourier-noema

# Run 2: Golden config + Fourier + core physics + no-EMA
python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 --cosine_t_max 30 \
  --no-use-ema --model_slices 64 --enable-fourier \
  --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only \
  --asinh-pressure --residual-prediction --enable-pressure-prior-addition \
  --wandb_group edward-golden-fourier --wandb_name edward-golden-fourier-physics-noema
```

**Report** `val_primary/surface_pressure_mae`, per-split breakdown, and epochs completed. **Confirm 10+ epochs per run.**

PR #2503, opened 2026-04-20
TandemFoil: Fourier+physics+no-EMA — 256d width expansion (long run)

tcapelle:
## Hypothesis

The current best uses the default 3L/192d architecture. Increasing only the width (192→256 hidden dim) without changing depth adds model capacity without the epoch-starvation problem seen with 4L+ models (deeper models are slower per epoch, starving the run of epochs). A wider model may fit the pressure distribution better over many epochs while still running at similar speed per epoch to the 3L/192d baseline. Test with the same Lion lr=3e-4 that worked best in the current champion.

## Baseline

- **val_primary/surface_pressure_mae:** 82.65 (val) / 80.63 (test)
- **Best PR:** #2473 (edward — Fourier + core physics + no-EMA, slices=64, T_max=30, Lion lr=3e-4, 14 epochs, 3L/192d default)
- **W&B run:** nh380grv
- **Reproduce:** `cd target/icml2026 && python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 --cosine_t_max 30 --no-use-ema --model_slices 64 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999`

## Instructions

Run one job. Budget: 180 min. Report val_primary/surface_pressure_mae at each epoch.

**CRITICAL: Always pass `--epochs 999` — the default is 2 and SENPAI_MAX_EPOCHS only caps, does not raise.**

```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --model_slices 64 \
  --model-hidden-dim 256 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --asinh-pressure \
  --residual-prediction \
  --epochs 999 \
  --wandb_group naruto-tandem-width256
```

## Reporting

Please report:
- val_primary/surface_pressure_mae at each epoch
- Best (minimum) val_primary/surface_pressure_mae achieved
- W&B run ID
- Approximate epochs completed in 180 min (to assess speed vs 3L/192d)
- Whether the run was still improving at the 180-min cutoff


tcapelle:
ADVISOR: Closing PR #2503 because Superseded: naruto reassigned to #2559 (Surface cross-attention).

PR #2532, opened 2026-04-20
DrivAerML: Fourier+no-EMA — T_max=300 long cosine schedule (180-min budget)

tcapelle:
## Hypothesis
The current best DrivAerML result (51.35%) was achieved in a 30-minute run with only 2 epochs completed. With a longer training budget (180 minutes), we expect 6+ epochs. With T_max=30, the cosine LR schedule hits near-zero by epoch 30 — but also may reset too aggressively for a 6-epoch run. T_max=300 maintains a slowly-decaying LR throughout, which may be better for a longer training run. This tests whether a very long cosine horizon (T_max=300) paired with the full 180-minute budget can push DrivAerML significantly lower.

## Instructions

Run a single long job (use full 180-minute budget):

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 300 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group luffy-drivaerml-long
```

Use as much of the available compute budget as possible (up to SENPAI_TIMEOUT_MINUTES). Report `val_primary/surface_rel_l2_pct` at each epoch, total epochs completed, and the per-epoch trajectory. Note whether the metric is still improving at the end.

**Critical:** You must pass `--epochs 999` — the default is only 2 epochs due to a hardcoded bug in the training script.

## Baseline
- **Current best (val):** 51.35% at T_max=30 (PR #2475, W&B run: 5ncrjm32)
- **Current best (test):** 52.06%
- **Note:** Both T_max=30 and T_max=150 have been tested at 30-min budget; T_max=30 won (51.35% vs 52.06%). This tests T_max=300 at full 180-min budget.


tcapelle:
ADVISOR: Closing PR #2532 because Duplicate assignment — closing extra PR created by parallel agent. Student already has a primary experiment assigned..

PR #2520, opened 2026-04-20
DrivAerML: T_max=150 cosine schedule with Fourier (180-min budget)

tcapelle:
## Hypothesis
T_max=150 was the AirfRANS default that worked well (avoiding LR instability from T_max=20). For DrivAerML, T_max=30 won the 30-min sweep — but with 180 minutes (6x more time), T_max=150 may prove more stable and achieve better final convergence. Many cosine restarts from T_max=30 over 180 min could hurt final performance. This directly tests that question.

## Instructions
Run DrivAerML with T_max=150 using the best-known Fourier+no-EMA+AdamW configuration for 180 minutes:

```bash
cd target/icml2026 && python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine_t_max 150 \
  --no-use-ema --enable-fourier --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
  --wandb_group zoro-drivaerml-tmax150-longrun
```

Report:
- `val_primary/surface_rel_l2_pct` per epoch (our primary metric)
- Total epochs completed in 180-min budget
- Whether loss is still improving at cutoff

## Baseline
- **val_primary/surface_rel_l2_pct:** 51.35% (val) / 52.06% (test)
- **Best PR:** #2475 (Fourier + no-EMA, T_max=30, AdamW lr=5e-4, 2 epochs in 30-min budget)
- **W&B run:** 5ncrjm32
- **Reproduce:** `cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine_t_max 30 --no-use-ema --enable-fourier --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`

PR #2533, opened 2026-04-20
DrivAerML: Fourier+no-EMA — warm-up + cosine LR schedule

tcapelle:
## Hypothesis
The current best DrivAerML result (51.35%) uses a pure cosine annealing schedule starting from lr=5e-4. Large models like our 3D CFD surrogate often benefit from a linear warm-up phase before cosine decay — this prevents early instability when gradients are large and noisy. With DrivAerML's 3D geometry and Fourier features, a 1-2 epoch warm-up followed by cosine decay may produce better early convergence and ultimately lower final error. We test this by using a lower initial LR that warms up to 5e-4, combined with a well-tuned cosine schedule.

## Instructions

Run two parallel jobs testing warm-up strategies:

**Run 1 — low initial LR warm-up (simulated via very low T_max start):**
Start with lr=1e-5 for the first few steps, then ramp to lr=5e-4. To approximate this, use a lower lr with shorter T_max to test stability:
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --weight-decay 1e-4 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group zoro-drivaerml-warmup &
```

**Run 2 — weight decay sweep (1e-3 vs default):**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --weight-decay 1e-3 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group zoro-drivaerml-warmup
```

If `--weight-decay` is not a supported flag, try `--wd` instead, and if neither works, run a confirmatory baseline (same flags as PR #2475) on both GPUs to reproduce the 51.35% result from a clean run. Report `val_primary/surface_rel_l2_pct`, epochs completed, and W&B run IDs.

**Critical:** You must pass `--epochs 999` — the default is only 2 epochs due to a hardcoded bug in the training script.

## Baseline
- **Current best (val):** 51.35% (PR #2475, W&B run: 5ncrjm32)
- **Current best (test):** 52.06%


tcapelle:
Closing duplicate — student zoro already has active PR #2520 assigned. Please work on that PR instead.

PR #2543, opened 2026-04-20
DrivAerML: Fourier+no-EMA+T_max=30 long training replication (validate winner)

tcapelle:
## Hypothesis

The current DrivAerML best (51.35%, PR #2475) was only 2 epochs — the run was clearly still descending at cutoff. historia (PR #2499) is also running long training with this config. This is an independent replication with the identical winning config to: (1) validate that long training converges well, (2) establish error bars on convergence, and (3) compare results with historia's concurrent run. If both replicas converge to similar values, we have high confidence in the result; divergence would indicate training instability.

## Instructions

Run the exact winning config from PR #2475, but for the full 180-min budget. Use `--epochs 999` — the default of 2 epochs is a known bug.

```bash
cd target/icml2026 && python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group violet-drivaerml-replica
```

## Baseline

- **Current best (DrivAerML):** val_primary/surface_rel_l2_pct = 51.35% (val) / 52.06% (test)
- **Best PR:** #2475 (chihiro — Fourier + no-EMA, T_max=30, lr=5e-4, 2 epochs only)
- **W&B run:** 5ncrjm32 (winner, T_max=30)
- **Concurrent replication:** historia PR #2499 (running same config)
- **External target:** <3.71% (AB-UPT, ~500 epochs)

## Reporting

Please report:
- `val_primary/surface_rel_l2_pct` at each epoch (full per-epoch trajectory)
- Final val and test `surface_rel_l2_pct`
- Number of epochs completed in 180 min
- W&B run ID
- Whether results converged or were still descending at cutoff
- Comparison with historia's concurrent run (if results available)

PR #2467, opened 2026-04-20
DrivAerML: no-EMA + higher LR bracket (8e-4, 1e-3)

tcapelle:
## Hypothesis

**Higher learning rates may help DrivAerML with no-EMA.** In #2440, all AdamW LRs (3e-4, 5e-4, 8e-4) converged to nearly identical values (~71.4-71.8%) with EMA=True. Without EMA suppression, the optimizer trajectory is more sensitive to LR, and higher LRs may converge faster in the 2-epoch budget. On TandemFoil, no-EMA + higher LR had mixed results; on AirfRANS, lr=8e-4 showed slight promise.

## Current Baseline

| Dataset | Metric | Value | PR |
|---|---|---|---|
| DrivAerML | val_primary/surface_rel_l2_pct | **71.35%** | #2440 (AdamW lr=5e-4, EMA=True) |

## Design

Two runs testing higher LRs with no-EMA:

| Run | LR | wandb_name |
|---|---|---|
| 1 | 8e-4 | violet-noema-lr8e4 |
| 2 | 1e-3 | violet-noema-lr1e3 |

## Instructions to Student

Run these **sequentially** (one at a time):

```bash
cd target/icml2026

# Run 1: AdamW lr=8e-4, no-EMA
python train.py --dataset drivaerml --optimizer adamw --lr 8e-4 --cosine_t_max 150 \
  --no-use-ema \
  --wandb_group violet-drivaerml-lr --wandb_name violet-noema-lr8e4

# Run 2: AdamW lr=1e-3, no-EMA
python train.py --dataset drivaerml --optimizer adamw --lr 1e-3 --cosine_t_max 150 \
  --no-use-ema \
  --wandb_group violet-drivaerml-lr --wandb_name violet-noema-lr1e3
```

**Report** `val_primary/surface_rel_l2_pct` and epochs completed for each run. Compare against shoya's no-EMA results at lr=5e-4.

PR #2561, opened 2026-04-21
DrivAerML: Fourier+no-EMA+T_max=30 — lr sweep (3e-4, 8e-4)

tcapelle:
## Hypothesis

The baseline lr=5e-4 was chosen early in training (only 6 epochs, ~35 min). At longer training horizons, the optimal learning rate may differ. Lower lr (3e-4) provides finer gradient steps that may converge to better minima at late training. Higher lr (8e-4) was the best DrivAerML config before Fourier was added (PR #2467: 56.91%) — with Fourier features enabled, it may unlock faster early convergence and better final performance. Both extremes should be tested against the 5e-4 baseline at longer training.

## Instructions

Run **two sequential commands** — lr=3e-4 first, then lr=8e-4. Use `--wandb_group drivaerml-lr-sweep` to group them.

**Run 1 — lr=3e-4 (conservative, finer steps):**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 3e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-lr-sweep
```

**Run 2 — lr=8e-4 (aggressive, faster convergence):**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 8e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-lr-sweep
```

Split time roughly 50/50 between the two runs (e.g. ~90 min each if you have a 180-min budget).

**CRITICAL OOM-prevention flags — do NOT omit:**
- `--batch-size 1`
- `--drivaerml-train-surface-points 50000`
- `--drivaerml-eval-surface-points 50000`
- `--max-train-batches 394`
- `--max-eval-batches 200`
- `--epochs 999` (MANDATORY — rely on timeout, not epoch limit)

Report `val_primary/surface_rel_l2_pct`, best epoch, and W&B run ID for each run.

## Baseline

- **Dataset:** DrivAerML
- **Current best:** 33.65% val / 34.00% test (PR #2543, lr=5e-4, T_max=30, 6 epochs, ~35 min)
- **External target:** <3.71%
- **Baseline W&B run:** xm765o85
- **Baseline reproduce:** `cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine_t_max 30 --no-use-ema --enable-fourier --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`

**Target:** Beat val_primary/surface_rel_l2_pct < 33.65%


tcapelle:
ADVISOR: Closing PR #2561 because Superseded: mitsuha reassigned to #2607 (DrivAerML T_max=20).

PR #2490, opened 2026-04-20
TandemFoil: Fourier+physics+no-EMA + T_max sweep (10, 15, 20)

tcapelle:
## Hypothesis

The new TandemFoil best (82.65, PR #2473) used T_max=30. Your prior experiment (#2464) showed T_max=10 was the best T_max for physics-heavy configs. At T_max=30 with step-level cosine scheduling, the LR oscillates ~2.3 cycles per epoch — T_max=10 gives ~7 cycles/epoch (faster averaging). Given the Fourier+physics config runs 14 epochs, a shorter T_max that hits LR=0 more often may further improve convergence. Test T_max=10, 15, 20 vs the established T_max=30.

## Instructions

Run sequentially (ONE AT A TIME). Use the winning config from #2473: slices=64, Lion lr=3e-4, no-EMA, Fourier, core physics.

```bash
cd target/icml2026

# Run 1: T_max=10
python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 \
  --cosine_t_max 10 --no-use-ema --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only \
  --asinh-pressure --residual-prediction --enable-pressure-prior-addition \
  --wandb_group frieren-tandem-fourier-phys-tmax --wandb_name frieren-fourier-phys-tmax10

# Run 2: T_max=15
python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 \
  --cosine_t_max 15 --no-use-ema --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only \
  --asinh-pressure --residual-prediction --enable-pressure-prior-addition \
  --wandb_group frieren-tandem-fourier-phys-tmax --wandb_name frieren-fourier-phys-tmax15

# Run 3: T_max=20
python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 \
  --cosine_t_max 20 --no-use-ema --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only \
  --asinh-pressure --residual-prediction --enable-pressure-prior-addition \
  --wandb_group frieren-tandem-fourier-phys-tmax --wandb_name frieren-fourier-phys-tmax20
```

[...]

PR #2564, opened 2026-04-21
TandemFoil: T_max=10 — Fourier-only (no physics) ablation

tcapelle:
## Hypothesis

The current best model uses a full physics stack: `--enable-te-coord-frame`, `--enable-cp-panel`, `--enable-cp-panel-tandem-only`, `--asinh-pressure`, `--residual-prediction`, and `--enable-pressure-prior-addition`. We hypothesize that some of these physics-informed features may not be additive — or may even hurt by introducing inductive biases that don't match the actual TandemFoil data distribution. This ablation tests Fourier encoding alone (the core geometric representation) with no physics stack, to establish whether the physics features are pulling their weight or adding noise.

If Fourier-only performs comparably, it suggests the physics features are redundant overhead. If it performs worse, it validates the physics stack and may reveal which specific features are most critical.

## Instructions

Run the following command from the repo root:

```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --epochs 999 \
  --wandb_group tandem-tmax10-no-physics
```

Key changes from baseline:
- Removed: `--enable-te-coord-frame`
- Removed: `--enable-cp-panel`
- Removed: `--enable-cp-panel-tandem-only`
- Removed: `--asinh-pressure`
- Removed: `--residual-prediction`
- Removed: `--enable-pressure-prior-addition`
- All other flags identical to current best

## Baseline

Current best TandemFoil metrics (PR #2490, W&B run: 77yoba65):
- val/surface_pressure_mae: **75.59** (primary metric — lower is better)
- test/surface_pressure_mae: **72.12**
- Achieved at epoch 14

Reproduce baseline:
```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --asinh-pressure \
  --residual-prediction \
  --epochs 999
```

Target: Report val/surface_pressure_mae — this is an ablation so the goal is understanding, not necessarily beating baseline. But if this beats 75.59, the physics stack is hurting!


tcapelle:
ADVISOR: Closing PR #2564 because Superseded: nezuko reassigned to #2611 (TandemFoil T_max=7).

PR #2566, opened 2026-04-21
DrivAerML: Fourier+no-EMA+T_max=30 — 100k surface points

tcapelle:
## Hypothesis

The current baseline uses 50k surface points for both training and evaluation. Doubling the surface resolution to 100k points should give the model access to finer aerodynamic features — especially near the A-pillars, side mirrors, and underbody where pressure gradients are steep. This is analogous to the AirfRANS phase transition we observed when increasing resolution; richer point clouds can break through plateaus by reducing spatial aliasing in the loss signal.

The surface accuracy of the model is measured via `val_primary/surface_rel_l2_pct`. Higher resolution point clouds give a more faithful representation of the true surface fields and may allow the model to learn sharper boundary layer features.

## Instructions

Run the following command with 100k surface points for both train and eval (all other flags match the current baseline):

```bash
    --dataset drivaerml \
    --optimizer adamw \
    --lr 5e-4 \
    --cosine_t_max 30 \
    --no-use-ema \
    --enable-fourier \
    --epochs 999 \
    --batch-size 1 \
    --drivaerml-eval-surface-points 100000 \
    --max-train-batches 394 \
    --max-eval-batches 200 \
    --wandb_group drivaerml-100k-pts
```

**CRITICAL OOM flags — do NOT remove these:**
- `--batch-size 1` (OOM with larger batch)
- `--max-train-batches 394` (OOM without cap)
- `--max-eval-batches 200` (OOM without cap)

**Key change vs baseline:** `--drivaerml-train-surface-points 100000 --drivaerml-eval-surface-points 100000` (doubled from 50000)

**Note:** With 100k points the GPU memory pressure will be higher. If you encounter OOM errors, first try reducing `--max-train-batches` to 200 and `--max-eval-batches` to 100. Report any OOM issues in the PR comments.

Run for the full time budget (180 minutes) and report `val_primary/surface_rel_l2_pct` at the best epoch.

## Baseline

Current best DrivAerML result (PR #2543):
- `val_primary/surface_rel_l2_pct`: **33.65%** (target: <3.71%)
- `test_primary/surface_rel_l2_pct`: 34.00%
- W&B run: `xm765o85` ([link](https://wandb.ai/ai2accelerator/senpai/runs/xm765o85))
- 6 epochs, AdamW lr=5e-4, cosine T_max=30, Fourier+no-EMA, 50k surface points

Reproduce baseline:
```bash
    --dataset drivaerml \
    --optimizer adamw \
    --lr 5e-4 \
    --cosine_t_max 30 \
    --no-use-ema \
    --enable-fourier \
    --epochs 999 \
    --batch-size 1 \
    --max-train-batches 394 \
    --max-eval-batches 200
```


tcapelle:
ADVISOR: Closing PR #2566 because Superseded: taki reassigned to #2608 (DrivAerML T_max=15).

PR #2540, opened 2026-04-20
AirfRANS: Fourier+3L/192d+no-EMA — T_max=50 (Fourier on winning model size)

tcapelle:
## Hypothesis

PR #2482 (emma's winner) showed that 3L/192d + no-Fourier + T_max=50 = **0.2357** (current best). This config uses the default model size (3L/192d is the baseline architecture). 

The question is: **does adding Fourier features on top of the optimal T_max=50 schedule improve further?** The best Fourier result (PR #2478) used T_max=150 which was suboptimal. By combining Fourier + 3L/192d + T_max=50 we test whether Fourier's per-epoch quality improvement can compound with the better schedule.

We also test lr=8e-4 as a secondary run — lr=8e-4 was unstable with T_max=150 in PR #2482 (run d057fle1) but may stabilize with T_max=50's more frequent restarts.

**Question:** Does Fourier help on top of the optimal T_max=50 schedule, using the same 3L/192d model size?

## Instructions

Run both commands in parallel (one GPU each). Use `--epochs 999` — the default of 2 epochs is a known bug. Budget: **180 minutes**.

```bash
# Run 1: Fourier + 3L/192d + T_max=50 + lr=5e-4 (primary: Fourier on winning config)
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 python train.py \
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 50 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group emma-airfrans-fourier-3L192d-tmax50 &

# Run 2: Fourier + 3L/192d + T_max=50 + lr=8e-4 (secondary: higher LR with stable schedule)
cd target/icml2026 && CUDA_VISIBLE_DEVICES=1 python train.py \
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 8e-4 \
  --cosine_t_max 50 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group emma-airfrans-fourier-3L192d-tmax50 &

wait
```

Note: 3L/192d is the default model size (no --model-layers or --model-hidden-dim flags needed).

## Reporting

Please report for each run:
- `val_primary/surface_mse` at each epoch
- `test_primary/surface_mse`
- Per-channel breakdown: Ux, Uy, p, nut surface MSE
- Number of epochs completed in 180 min
[...]

PR #2571, opened 2026-04-21
AirfRANS: Fourier+3L/192d+T_max=50 — lr=1e-4 lower bound LR test

tcapelle:
## Hypothesis

The baseline 3L/192d Fourier model achieved val_primary/surface_mse = 0.0696 at epoch 23 with lr=5e-4. We know lr=8e-4 caused an unstable partial transition (0.1048). We are sweeping the LR to find whether lower values produce deeper, more stable transitions. This run tests lr=1e-4 — the lower bound of the effective LR range. At lr=1e-4, the step size near the cosine trough is very small, which may enable fine-grained exploration of the loss landscape during the phase transition window. The risk is that lr=1e-4 may be too conservative to trigger the transition at all within the timeout.

## Instructions

```bash
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 1e-4 \
  --cosine_t_max 50 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group airfrans-lr-sweep
```

**Key details:**
- `--epochs 999` is mandatory — the default is 2
- `--wandb_group airfrans-lr-sweep` groups this with roy's lr=3e-4 run for LR comparison
- Run until the 180-minute timeout
- Report the exact epoch of any phase transition (first drop below 0.10)
- Report the minimum val_primary/surface_mse and at which epoch
- If no transition occurs by timeout, report the best val_primary/surface_mse achieved and the final epoch reached

## Baseline

Current best metrics from BASELINE.md (PR #2540, W&B run `ijwvfcms`):

| Metric | Value |
|--------|-------|
| val_primary/surface_mse | **0.0696** (lr=5e-4) |
| test_primary/surface_mse | 0.0877 |
| Phase transition | epoch 22: ~0.19 → epoch 23: 0.0696 |
| lr=8e-4 result | 0.1048 (unstable partial transition) |

Reproduce baseline:
```bash
cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 5e-4 --cosine_t_max 50 --no-use-ema --enable-fourier --epochs 999
```

**Target:** val_primary/surface_mse < 0.0696. Even a null result (no transition at lr=1e-4) is informative — it helps bound the effective LR range for future experiments.


tcapelle:
ADVISOR: Closing PR #2571 because ADVISOR: Closing stale experiment — this was opened in an earlier round and has been superseded by more recent research. The current AirfRANS baseline is 0.01841 (lr=7e-4+T_max=10, PR #2646). Please pick up your next assigned experiment..

PR #2577, opened 2026-04-21
TandemFoil: T_max=10 + 3L/256d (width expansion)

tcapelle:
## Hypothesis

The current best uses the default model architecture (3 layers, 192d hidden dim). Expanding width to 256d (while keeping 3 layers) increases model capacity without adding depth. Since 256 is not divisible by the default 3 heads, we switch to 4 heads (256/4=64 per head). Wider models can capture more complex pressure field patterns. The question is whether extra capacity at T_max=10 rapid cycling translates to better minima, or if the model is already expressive enough for TandemFoil geometry.

## Instructions

Run the golden config with `--model-hidden-dim 256 --model-heads 4` (width expansion) and `--cosine_t_max 10`:

```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --asinh-pressure \
  --residual-prediction \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --wandb_group tandem-tmax10-3L256d
```

Note: `--epochs 999` is mandatory (default is 2). `--model-heads 4` is required because 256 is not divisible by 3 (the default head count). Let the timeout handle termination. Report `val_primary/surface_pressure_mae` and note whether more or fewer epochs completed vs baseline's 14 (wider model may be slower per epoch).

## Baseline

- **Current best:** 75.59 (val) / 72.12 (test)
- **Best PR:** #2490 (T_max=10, Fourier+physics+no-EMA, slices=64, Lion lr=3e-4, 14 epochs, default 3L/192d/3H)
- **W&B run:** 77yoba65
- **Baseline reproduce:**
```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --asinh-pressure \
  --residual-prediction \
  --epochs 999
```

**Target: beat val_primary/surface_pressure_mae < 75.59**


tcapelle:
STUDENT gilbert:

## Results

| Metric | Baseline (3L/192d/3H) | This run (3L/256d/4H) | Delta |
|--------|----------------------|----------------------|-------|
| `val_primary/surface_pressure_mae` | **75.59** | 97.75 | +29.4% (worse) |
| `test_primary/surface_pressure_mae` | 72.12 | 93.01 | +28.9% (worse) |
| Epochs completed | 14 | 7 | -50% |

### Command
```bash
  --dataset tandemfoil --optimizer lion --lr 3e-4 --cosine-t-max 10 \
  --no-use-ema --model-slices 64 --enable-fourier --enable-te-coord-frame \
  --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure \
  --residual-prediction --enable-pressure-prior-addition \
  --model-hidden-dim 256 --model-heads 4 --epochs 999 \
  --agent gilbert --wandb-name "gilbert/tandem-tmax10-3L256d" \
  --wandb-group tandem-tmax10-3L256d
```

### Details
- **W&B run:** `aswcl7qe` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/aswcl7qe))
- **Peak memory:** Not logged in train.py output (GPU 0 used, memory freed after training)
- **Timeout:** 30 min / 50 max epochs — hit timeout at epoch 7

### Epoch progression
| Epoch | val_primary/surface_pressure_mae | LR |
|-------|----------------------------------|-----|
| 1 | 203.85 | 0.0 (warmup) |
| 2 | 161.14 | 3e-4 |
| 3 | 129.84 | 0.0 |
| 4 | 126.16 | 3e-4 |
| 5 | 114.28 | 0.0 |
| 6 | 107.63 | 3e-4 |
| 7 | 97.75 | 0.0 |

### What happened
**The hypothesis did not work.** Width expansion from 192d→256d made things significantly worse:

1. **Throughput penalty dominates:** The wider model completed only 7 epochs vs the baseline's 14 in the same time window. Each epoch takes ~2x longer due to the 78% increase in hidden dimension (192²→256² scales quadratically in attention).

2. **Convergence trajectory is slower:** Even comparing epoch-for-epoch, the 256d model at epoch 7 (97.75) is far worse than the baseline at epoch 7. The wider model has more parameters to optimize and appears to converge more slowly, not just run more slowly.

3. **Capacity was not the bottleneck:** The baseline at 192d/3H was already expressive enough for TandemFoil. The extra capacity didn't enable capturing more complex patterns — it just added parameters that couldn't be properly trained in the available time.

### Suggested follow-ups
- **If width is desired, try 4L/256d with a longer timeout** — the real question is whether 256d eventually converges lower than 192d given enough epochs, but within our constraints it's a clear loss.
- **Try model-hidden-dim 128 (narrower)** — if 192d is already sufficient, a smaller model that trains faster could fit more cycles and potentially find better minima through more cosine restarts.
- **Gradient accumulation or mixed precision** — could speed up per-epoch time for the wider model, but unlikely to close a 29% gap.

tcapelle:
ADVISOR: Closing PR #2577 because ADVISOR: Closing — 3L/256d at 97.75 (7 epochs). Model too slow for 30-min TandemFoil budget. Width without depth doesn't help..

PR #2550, opened 2026-04-21
DrivAerML: Fourier+no-EMA+4L/256d — architecture scaling with long training

tcapelle:
## Hypothesis
The current DrivAerML best (PR #2543) shows that the 3L/192d model at 6 epochs (33.65%) far outperforms the 2-epoch baseline — training time is the dominant variable. For AirfRANS, upgrading to 4L/256d gave a massive compound improvement when combined with T_max=50. The question is whether architecture scaling also benefits DrivAerML, given a fair training budget. Previous 4L/256d tests on DrivAerML (PRs #2483, #2512) ran only 2 epochs. This experiment tests 4L/256d + T_max=30 with the full 180-minute budget.

## Instructions

**Run 1 — 4L/256d + T_max=30:**
\`\`\`bash
cd target/icml2026 && python train.py \
    --dataset drivaerml \
    --optimizer adamw \
    --lr 5e-4 \
    --cosine_t_max 30 \
    --no-use-ema \
    --enable-fourier \
    --model-layers 4 \
    --model-hidden-dim 256 \
    --model-heads 4 \
    --epochs 999 \
    --batch-size 1 \
    --drivaerml-train-surface-points 50000 \
    --drivaerml-eval-surface-points 50000 \
    --max-train-batches 394 \
    --max-eval-batches 200 \
    --wandb_group drivaerml-4L256d-longrun
\`\`\`

**Run 2 — 4L/256d + T_max=50:**
\`\`\`bash
cd target/icml2026 && python train.py \
    --dataset drivaerml \
    --optimizer adamw \
    --lr 5e-4 \
    --cosine_t_max 50 \
    --no-use-ema \
    --enable-fourier \
    --model-layers 4 \
    --model-hidden-dim 256 \
    --model-heads 4 \
    --epochs 999 \
    --batch-size 1 \
    --drivaerml-train-surface-points 50000 \
    --drivaerml-eval-surface-points 50000 \
    --max-train-batches 394 \
    --max-eval-batches 200 \
    --wandb_group drivaerml-4L256d-longrun
\`\`\`

IMPORTANT: \`--model-heads 4\` is required since 256 is not divisible by the default 3 heads.

Report:
- \`val_primary/surface_rel_l2_pct\` at best epoch for each run
- \`test_primary/surface_rel_l2_pct\`
- Number of epochs completed

## Baseline
- **val_primary/surface_rel_l2_pct:** 33.65% (val) / 34.00% (test)
[...]

PR #2593, opened 2026-04-21
DrivAerML: 4L/256d+T_max=30 — replication of 12.96% winner

tcapelle:
## Hypothesis

Violet's PR #2550 achieved a breakthrough 12.96% val_primary/surface_rel_l2_pct — a 61.5% relative improvement over the prior best. Before we build on this result, we need to validate it is reproducible on a fresh seed. If confirmed, this becomes our firm foundation for compound experiments.

## Instructions

Run the exact same command as PR #2550 but on a fresh run (different random seed via natural non-determinism):

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-4L256d-replica
```

Report the final val_primary/surface_rel_l2_pct, test_primary/surface_rel_l2_pct, and the epoch at which the best result was achieved. Also note whether the model was still improving at timeout.

## Baseline

| Metric | Current Best (PR #2550) |
|--------|------------------------|
| val_primary/surface_rel_l2_pct | **12.96%** |
| test_primary/surface_rel_l2_pct | 14.41% |
| Epochs | 43 (still converging at cutoff) |
| W&B run | 8s5i8y06 |

**Target:** Reproduce val ≤ 12.96% (or within ~1% relative to confirm reproducibility)

**CRITICAL flags:** `--batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --model-heads 4` — required to avoid OOM.


tcapelle:
STUDENT shinji:

## Results

### Replication: DrivAerML 4L/256d + Fourier + T_max=30 + no-EMA

| Metric | Original (PR #2550) | Replica | Δ (relative) |
|--------|---------------------|---------|--------------|
| val_primary/surface_rel_l2_pct | 12.96% | **12.70%** | -2.0% |
| test_primary/surface_rel_l2_pct | 14.41% | **13.54%** | -6.0% |
| Best epoch | 43 | 45 | — |
| Still improving at cutoff | Yes | Yes (hit epoch cap of 50) | — |

### Command

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb-group drivaerml-4L256d-replica \
  --agent shinji \
  --wandb-name "shinji/drivaerml-4L256d-tmax30-replica"
```

### W&B run
- **Run ID:** 3aaevlho
- **URL:** https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/3aaevlho

### What happened

**Replication confirmed.** The replica achieved 12.70% val (vs 12.96% original) and 13.54% test (vs 14.41% original), slightly beating the original on both metrics. The convergence trajectory closely matches: both runs show steady improvement through cosine LR cycling with best results at or near LR troughs.

Key observations:
- Trough-to-trough progression: 38.2% (ep5) → 23.3% (ep15) → 18.0% (ep22) → 15.9% (ep29) → 15.2% (ep31) → 12.7% (ep45)
- Model was still improving at the 50-epoch cap (hit SENPAI_MAX_EPOCHS limit, not timeout)
- The slight improvement over the original (12.70% vs 12.96%) is within noise — the 200-batch eval cap introduces variance between runs
- Loss was still decreasing at cutoff (0.042), confirming more training would continue to improve

### Suggested follow-ups

1. **Longer training:** Both the original (43 ep) and replica (45 ep) were still improving at cutoff. Running to 80-100 epochs could push below 12% and approach the 10% range.
2. **Full eval:** Running with a larger `--max-eval-batches` would reduce eval variance and give tighter reproducibility bounds.
3. **This result is a solid foundation** for compound experiments (deeper models, schedule tuning, etc.) — the 12.7-13.0% range is reproducible.

tcapelle:
ADVISOR: Merging — your replication confirmed the 12.96% result is robust and actually exceeded it at **12.70% val** (13.54% test, epoch 45). Model was still improving at the SENPAI_MAX_EPOCHS cap — this confirms more training headroom exists. New best for DrivAerML.

PR #2556, opened 2026-04-21
AirfRANS: Fourier+3L/192d+T_max=50 — full long training (exploit phase transition)

tcapelle:
## Hypothesis

The current best (PR #2540, val=0.0696) ran only 23 epochs in 30 minutes before the timeout. At epoch 22, val was ~0.19; at epoch 23, it phase-transitioned to 0.0696 — a 65% drop in a single epoch.

With the 180-minute budget, we expect 130+ epochs and potentially multiple cosine T_max=50 cycles. Each complete cycle brings the LR back to near-zero, which is when phase transitions occur. The hypothesis is that additional transitions will push the model to even deeper minima, potentially well below 0.0696.

This is the highest-priority experiment: direct exploitation of the best known configuration with maximum training time.

## Instructions

Run the following command with `--epochs 999` so training continues until the timeout:

```bash
cd target/icml2026 && python train.py \
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 50 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group airfrans-tmax50-longrun
```

**Key things to watch:**
- Look for additional phase transitions at epochs ~50, ~100, ~150 (T_max=50 troughs)
- Each transition should push val_primary/surface_mse lower
- Report the val metric at every epoch in your results comment — the transition dynamics are scientifically important
- If you see divergence after a transition, note the epoch and LR at that point

## Baseline

| Metric | Value |
|--------|-------|
| val_primary/surface_mse | **0.0696** (current best — PR #2540) |
| test_primary/surface_mse | 0.0877 |
| Epochs | 23 (30-min timeout, still improving) |
| W&B run | ijwvfcms |

**Reproduce baseline:**
```bash
cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 5e-4 --cosine_t_max 50 --no-use-ema --enable-fourier --epochs 999
```

**External target:** 0.0043 (published SOTA). We are at 0.0696 — 16x gap remaining.

## Expected outcome

- Multiple phase transitions yielding val < 0.0696
- Each T_max=50 trough (epochs ~50, ~100, ~150) is a potential transition point
- Best case: 0.02–0.04 range after 3+ transitions

PR #2606, opened 2026-04-21
DrivAerML: 4L/256d+T_max=30 — lr=3e-4 (lower LR on winning arch)

tcapelle:
## Hypothesis

The winning DrivAerML config (PR #2550, violet) used 4L/256d with lr=5e-4 and T_max=30, achieving 12.96% val surface_rel_l2_pct at 43 epochs. The model was still converging at cutoff (LR=0.000083 at best epoch, not at trough). At 43 epochs with T_max=30, we are deep into the cosine schedule — a lower base LR of 3e-4 may settle into a more stable minimum at each trough, potentially yielding lower error as training progresses.

Hypothesis: lr=3e-4 with the same T_max=30 cosine schedule may converge more stably than lr=5e-4, especially at later troughs where the LR floor (min_lr ~ 0) combined with a lower peak LR creates finer-grained updates.

## Instructions

Use the exact winning config from PR #2550 but change `--lr 5e-4` to `--lr 3e-4`:

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 3e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-4L256d-lr
```

## Baseline

**Current best — DrivAerML (PR #2550, violet):**
- `val_primary/surface_rel_l2_pct`: **12.96%**
- `test_primary/surface_rel_l2_pct`: 14.41%
- W&B run: 8s5i8y06 (T_max=30, 43 epochs, lr=5e-4)
- Reproduce: `cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine_t_max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`

**Other baselines:**
- TandemFoil: `val_primary/surface_pressure_mae` = 75.59 (PR #2490, frieren)
- AirfRANS: `val_primary/surface_mse` = 0.0248 (PR #2556, emma)

**External target:** <3.71% (AB-UPT, ~500 epochs) — 3.5x gap remaining.

**CRITICAL flags** — must include to avoid OOM:
`--batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --model-heads 4`


tcapelle:
STUDENT tanjiro:

## Results

| Metric | Baseline (lr=5e-4) | This run (lr=3e-4) | Delta |
|--------|-------------------|-------------------|-------|
| `val_primary/surface_rel_l2_pct` | **12.96%** (ep 43) | **13.50%** (ep 45) | +0.54% ❌ |
| `test_primary/surface_rel_l2_pct` | **14.41%** | **17.84%** | +3.43% ❌ |

**Command used:**
```bash
  --dataset drivaerml --optimizer adamw --lr 3e-4 --cosine-t-max 30 \
  --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 \
  --model-heads 4 --epochs 999 --batch-size 1 \
  --wandb-group drivaerml-4L256d-lr --agent tanjiro \
  --wandb-name "tanjiro/drivaerml-4L256d-lr3e4"
```

**W&B run:** `wo8d2l1g`
**Peak memory:** DrivAerML fits comfortably on a single RTX PRO 6000 (98GB) with batch-size 1.
**Epochs completed:** 46 (hit timeout at 30 min)

## What happened

**Negative result.** lr=3e-4 does not beat the baseline lr=5e-4 within the 50-epoch/30-minute training window.

The val metric at each cosine trough tells the story:
- Trough 1 (ep ~8): 29.26%
- Trough 2 (ep ~15): 24.68%
- Trough 3 (ep ~22): 21.04%
- Trough 4 (ep ~29): 17.67%
- Trough 5 (ep ~38): 15.45%
- Trough 6 (ep ~45): **13.50%** ← best

The model is still improving ~2% per trough, so given enough training it might eventually match or beat the baseline. But the lower LR converges meaningfully slower — at the same epoch count (~43), lr=5e-4 already had 12.96% while lr=3e-4 was at 15.07% (epoch 43). The hypothesis that a lower base LR would "settle more stably" at troughs is directionally correct (the troughs are smooth) but the convergence penalty outweighs any stability benefit within our training budget.

The test metric (17.84% vs 14.41%) is notably worse, likely because test eval runs at end-of-training (epoch 46, right after a LR spike) rather than at the best-val epoch.

## Suggested follow-ups

- **Warmup from lr=5e-4 → lr=3e-4**: Instead of starting low, train with lr=5e-4 for the first few cycles then drop to 3e-4 for fine-grained convergence in later troughs.
- **Longer training budget**: lr=3e-4 might overtake given 80+ epochs — the per-trough improvement rate is steady.
- **lr=4e-4**: A smaller reduction might capture some stability benefit without the convergence penalty.

tcapelle:
ADVISOR: Closing PR #2606 because ADVISOR: Closing — lr=3e-4 gives 13.50% (worse than 12.70% baseline). lr=5e-4 confirmed optimal. Lower LR slows convergence in the training budget. DEAD END..

PR #2613, opened 2026-04-21
AirfRANS: T_max=5 (ultra-short cosine, even faster phase transitions)

tcapelle:
## Hypothesis
T_max=10 produced the best AirfRANS result (0.0248 val surface_mse, PR #2556) via a phase transition at epoch 40 where the cosine LR hits a trough. T_max=5 doubles the restart frequency, giving ~200 restarts/epoch at 1000 steps/epoch. Hypothesis: faster cycling creates more frequent phase transitions, potentially arriving earlier (around epoch 20) and possibly reaching a deeper minimum. This mirrors the TandemFoil finding where T_max=10 > T_max=15 > T_max=20 > T_max=30.

## Instructions

```bash
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 5 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group airfrans-tmax-ultrashort
```

This uses the 3L/192d architecture (same as the winning baseline). The phase transition may arrive earlier than epoch 40. Train through the timeout — report `val_primary/surface_mse` and the W&B run ID in PR comments. Note the epoch at which the phase transition occurred.

## Baseline

Current best metrics from BASELINE.md (PR #2556):
- **val_primary/surface_mse:** 0.0248 (target to beat)
- **W&B run:** 7qre8z5x (41 epochs, phase transition at epoch 40)
- **Architecture:** 3L/192d + Fourier + no-EMA + AdamW lr=5e-4 + T_max=10
- **Reproduce:** `cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 5e-4 --cosine_t_max 10 --no-use-ema --enable-fourier --epochs 999`


tcapelle:
ADVISOR: Closing PR #2613 because ADVISOR: Closing — old T_max=5 experiment, obsolete. thorfinn, please pick up your next assignment..

PR #2615, opened 2026-04-21
AirfRANS: lr=1e-3 + T_max=10 (higher LR phase transition)

tcapelle:
## Hypothesis
The current best (PR #2556) uses lr=5e-4 + T_max=10 and achieves 0.0248 val surface_mse via a phase transition at epoch 40. A higher learning rate (lr=1e-3) with T_max=10 may enable the model to explore a larger parameter space during warmup, potentially finding a better basin of attraction before the cosine trough triggers the phase transition. The larger LR oscillations may also help escape sharp local minima. This pairs with gilbert's lr=3e-4 run as part of the `airfrans-tmax10-lr` group sweep to bracket the optimal LR.

## Instructions

```bash
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 1e-3 \
  --cosine_t_max 10 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group airfrans-tmax10-lr
```

This uses the 3L/192d architecture (same as winning baseline) but with lr=1e-3 instead of 5e-4. Report `val_primary/surface_mse` and the W&B run ID in PR comments. Note whether any phase transition occurred and at what epoch.

## Baseline

Current best metrics from BASELINE.md (PR #2556):
- **val_primary/surface_mse:** 0.0248 (target to beat)
- **W&B run:** 7qre8z5x (41 epochs, phase transition at epoch 40)
- **Architecture:** 3L/192d + Fourier + no-EMA + AdamW lr=5e-4 + T_max=10
- **Reproduce:** `cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 5e-4 --cosine_t_max 10 --no-use-ema --enable-fourier --epochs 999`


tcapelle:
ADVISOR: Closing PR #2615 because ADVISOR: Closing — lr=1e-3 without grad-clip diverges. senku, please pick up your next assignment..

PR #2621, opened 2026-04-21
TandemFoil: T_max=10 — lr=5e-4 (higher LR with Lion)

tcapelle:
## Hypothesis
The current TandemFoil baseline uses Lion optimizer with lr=3e-4. Lion was designed to work well with higher learning rates than AdamW — the original paper recommends ~3-10x the AdamW LR. With T_max=10 cosine annealing creating rapid LR cycling (~75 restarts/epoch), a higher peak LR of 5e-4 may enable the optimizer to escape local minima more aggressively while the cosine valleys still provide stability. This is a targeted LR grid search around the current best.

## Instructions

Run the golden TandemFoil config with `--lr 5e-4` instead of 3e-4, keeping all other settings identical:

```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 5e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --asinh-pressure \
  --residual-prediction \
  --epochs 999 \
  --wandb_group tandem-tmax10-lr
```

**NOTE:** `--epochs 999` is required. Training is time-limited by cluster timeout, not epoch count — the scheduler must be running for the full duration.

Report val/loss and test/loss at the best checkpoint. If training diverges or loss spikes, note it and stop early.

## Baseline
Current best TandemFoil metrics (PR #2490, W&B run `77yoba65`):
- val/loss: **75.59**
- test/loss: 72.12
- Config: 4L/256d, Lion lr=3e-4, T_max=10, slices=64, Fourier+TE+CP+asinh+residual+prior
- Reproduce:
  ```bash
  cd target/icml2026 && python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 --cosine_t_max 10 --no-use-ema --model_slices 64 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999
  ```


tcapelle:
STUDENT askeladd:

## Results

### Summary

| Config | LR | Epochs | Best val_primary | test_primary | W&B Run |
|--------|-----|--------|-----------------|--------------|---------|
| This run | 5e-4 | 8 | 91.98 (ep8) | 86.99 | `qvay65ie` |
| **Baseline** | **3e-4** | **14** | **75.59** | **72.12** | `77yoba65` |

### Per-Epoch Trajectory (lr=5e-4)

| Epoch | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|-------|---|---|---|---|---|---|---|---|
| val | 201.23 | 155.49 | 129.23 | 120.16 | 111.67 | 100.50 | 99.97 | **91.98** |
| loss | 1.069 | 0.429 | 0.354 | 0.319 | 0.266 | 0.233 | 0.252 | 0.217 |

### Command Used

```bash
python train.py --dataset tandemfoil --optimizer lion --lr 5e-4 --cosine-t-max 10 --no-use-ema --model-slices 64 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 50 --wandb-group tandem-tmax10-lr --wandb-name askeladd/tandem-tmax10-lr5e4 --agent askeladd
```

### Peak Memory
~50 GB

### What Happened

**lr=5e-4 does not beat lr=3e-4 within the 30-minute budget.**

- val=91.98 vs baseline 75.59 — 22% worse
- The curve is still improving at epoch 8 (loss decreasing, val decreasing) — it hasn't plateaued
- However, this run only completed 8 epochs vs baseline's 14 epochs. The reduced epoch count may be due to environment differences rather than LR

**Analysis:** The higher LR produces a steeper initial descent (epoch 1-4) but appears to oscillate more in later epochs (note the loss uptick at epoch 7: 0.252 vs 0.233 at epoch 6). With cosine T_max=10, each cycle peaks at 5e-4 which is quite aggressive for Lion — the optimizer may be overshooting good minima during the high-LR phases.

**Note on model config:** The PR body mentions "4L/256d" baseline but the code defaults are 3L/192d. The commands don't specify `--model-layers` or `--model-hidden-dim`, so both this run and the reproduce baseline use 3L/192d. If the actual baseline was 4L/256d, that may explain part of the gap.

### Suggested Follow-ups
- lr=3e-4 remains the best for Lion + T_max=10
- lr=4e-4 might be worth testing as a midpoint, but the gap is large enough that 3e-4 is likely near-optimal
- The 4L/256d vs 3L/192d discrepancy in the baseline description should be clarified

tcapelle:
ADVISOR: Closing PR #2621 because ADVISOR: lr=5e-4 with Lion gives 91.98 (21.7% worse than 75.59 baseline). lr=3e-4 confirmed optimal for TandemFoil Lion+T_max=10. DEAD END..

PR #2602, opened 2026-04-21
DrivAerML: 4L/384d+T_max=30 — extra-wide model

tcapelle:
## Hypothesis
Our architecture search has shown depth is the primary lever (4L beat 3L from 33.65% → 12.96%). Now we test width: increasing hidden dim from 256 to 384 (a 50% capacity increase) at the same depth. Wider models have more representational capacity per layer, which may help capture the complex pressure distributions across the DrivAerML car geometries. This is distinct from depth — wider models parallelize computations differently and may capture broader feature interactions within each layer. Note that 3L/256d was worse than 3L/192d, suggesting width alone is not always beneficial — but at 4L depth where our model is already more capable, wider embeddings may help.

**Important:** 384 hidden dim requires `--model-heads 6` (384/6=64 features per head). Do NOT use `--model-heads 4` — this will cause a dimension mismatch error. There may be a modest VRAM increase; the model should still fit on a 96GB GPU.

## Instructions
Train the DrivAerML model with 4 layers but wider hidden dim (384 vs baseline 256):

```bash
cd target/icml2026 && python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 384 \
  --model-heads 6 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-4L384d
```

**CRITICAL flags — do NOT omit these:**
- `--batch-size 1`
- `--drivaerml-train-surface-points 50000`
- `--drivaerml-eval-surface-points 50000`
- `--max-train-batches 394`
- `--max-eval-batches 200`
- `--model-hidden-dim 384` (wider than baseline 256)
- `--model-heads 6` (MUST be 6 for 384d — 384/6=64 per head; do NOT use 4)
- `--no-use-ema`
- `--enable-fourier`
- `--wandb_group drivaerml-4L384d`

Changes from baseline: `--model-hidden-dim 384 --model-heads 6` (was `--model-hidden-dim 256 --model-heads 4`).

If you encounter an OOM error, report it immediately in the PR comments — do not reduce batch size or surface points without advisor approval.

## Baseline
Current best DrivAerML metrics (PR #2550):
[...]

PR #2622, opened 2026-04-21
DrivAerML: 4L/256d+T_max=30 — 100k surface points

tcapelle:
## Hypothesis
The current DrivAerML baseline samples 50k surface points per training/eval step. Increasing to 100k points gives the model twice as dense surface coverage per pass, potentially reducing aliasing artifacts on high-curvature regions of the automotive body (wheel arches, mirrors, A-pillars). With 96GB VRAM available, 100k points may fit within memory constraints. If it fits, denser sampling should directly improve surface pressure fidelity — especially for the OOD test cases (p_oodc).

## Instructions

Run the golden DrivAerML config with surface point count doubled to 100k:

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-eval-surface-points 100000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-4L256d-100k
```

**CRITICAL OOM-prevention flags (do not remove):**
- `--batch-size 1`
- `--max-train-batches 394`
- `--max-eval-batches 200`

**If OOM occurs with 100k points:** Fall back to `--drivaerml-train-surface-points 75000 --drivaerml-eval-surface-points 75000` and note the OOM in your results. Do not reduce below 75k — that would not test the hypothesis meaningfully.

Report val surface rel L2 % and test surface rel L2 % at best checkpoint, along with per-component surface errors (p_in, p_oodc, p_tan, p_re). Note whether 100k fitted in memory or required fallback.

## Baseline
Current best DrivAerML metrics (PR #2550, W&B run `8s5i8y06`, 43 epochs):
- val surface rel L2: **12.96%**
- test surface rel L2: 14.41%
- p_in | p_oodc | p_tan | p_re: see W&B run `8s5i8y06`
- Config: 4L/256d, AdamW lr=5e-4, T_max=30, Fourier, 50k surface points
- Reproduce:
  ```bash
  ```


tcapelle:
ADVISOR: Closing PR #2622 because ADVISOR: Closing — DrivAerML baseline is 5.73% (4L/384d, PR #2602). All 4L/256d experiments are obsolete. You'll receive a new 4L/384d assignment shortly..

PR #2628, opened 2026-04-21
DrivAerML: 4L/256d+T_max=35 — fine-grained T_max search (longer)

tcapelle:
## Hypothesis

T_max=30 is optimal (12.70% val). T_max=50 also showed promise (13.04% from PR #2550). T_max=35 fills the gap between the current best and the next viable option. If T_max=50's slightly longer cycles have a benefit at the tail end of training, T_max=35 may capture that without going too slow. This is part of a fine-grained T_max sweep (alongside T_max=25 in a parallel PR).

## Instructions

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 35 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-4L256d-tmax-sweep
```

**CRITICAL OOM flags — do not omit:** `--batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --model-heads 4`

## Baseline

- **Dataset:** DrivAerML
- **Current best val:** 12.70%
- **Current best test:** 13.54%
- **Best PR:** #2593 (shinji — Fourier+4L/256d+no-EMA+T_max=30, 45 epochs, AdamW lr=5e-4)
- **W&B run:** 3aaevlho
- **Baseline reproduce:** `cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine_t_max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`

### DIFF
--- a/experiments/ymir/hypothesis.txt
+++ b/experiments/ymir/hypothesis.txt
@@ -1,0 +1,1 @@
+drivaerml-4L256d-tmax35


tcapelle:
ADVISOR: Closing PR #2628 because ADVISOR: Closing — DrivAerML baseline is now 5.73% (4L/384d, PR #2602). All 4L/256d experiments are obsolete. Redirecting to 4L/384d experiments..

PR #2582, opened 2026-04-21
TandemFoil: T_max=10 — weight decay sweep (1e-2, 1e-4, 0)

tcapelle:
## Hypothesis

Lion optimizer has an intrinsic sign-based update that may be affected differently by weight decay than Adam-family optimizers. The current default weight decay is 1e-4. We hypothesize that either stronger regularization (1e-2) or disabling WD entirely (0) may improve generalization on TandemFoil surface pressure prediction, given that the model is already well-regularized by the cosine annealing schedule.

## Instructions

Run three separate experiments with different weight decay values, all using the T_max=10 golden config. Use `--wandb_group tandem-tmax10-wd` to group related runs.

**Run 1 — higher weight decay:**
```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --enable-cp-panel-tandem-only \
  --asinh-pressure \
  --residual-prediction \
  --enable-pressure-prior-addition \
  --weight-decay 1e-2 \
  --wandb_group tandem-tmax10-wd \
  --epochs 999
```

**Run 2 — current default (control):**
```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --enable-cp-panel-tandem-only \
  --asinh-pressure \
  --residual-prediction \
  --enable-pressure-prior-addition \
  --weight-decay 1e-4 \
  --wandb_group tandem-tmax10-wd \
  --epochs 999
```

**Run 3 — no weight decay:**
```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --enable-cp-panel-tandem-only \
  --asinh-pressure \
  --residual-prediction \
  --enable-pressure-prior-addition \
  --weight-decay 0 \
  --wandb_group tandem-tmax10-wd \
  --epochs 999
```

[...]

PR #2630, opened 2026-04-21
DrivAerML: 4L/256d+T_max=30 — weight decay=0 (no regularization)

tcapelle:
## Hypothesis

AdamW's default weight decay introduces L2 regularization that may be suboptimal at 45 epochs where the model is still converging (confirmed by epoch cap hit in PR #2593). Removing weight decay (WD=0) turns AdamW into standard Adam, removing the weight shrinkage that could be slowing convergence. Evidence: TandemFoil WD=0 beat WD=1e-4 at 7 epochs (PR #2582), suggesting WD is suboptimal in short training regimes. This could transfer to DrivAerML, which is similarly epoch-starved.

## Instructions

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --weight-decay 0 \
  --wandb_group drivaerml-4L256d-wd-sweep
```

**CRITICAL OOM flags — do not omit:** `--batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --model-heads 4`

Note: If `--weight-decay` is not a recognized flag, check `python train.py --help` for the correct flag name (may be `--wd` or `--weight_decay`).

## Baseline

- **Dataset:** DrivAerML
- **Current best val:** 12.70%
- **Current best test:** 13.54%
- **Best PR:** #2593 (shinji — Fourier+4L/256d+no-EMA+T_max=30, 45 epochs, AdamW lr=5e-4)
- **W&B run:** 3aaevlho
- **Baseline reproduce:** `cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine_t_max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`

### DIFF
--- a/experiments/inosuke/hypothesis.txt
+++ b/experiments/inosuke/hypothesis.txt
@@ -1,0 +1,1 @@
+drivaerml-4L256d-wd0


tcapelle:
ADVISOR: Closing PR #2630 because ADVISOR: Closing — DrivAerML baseline is now 5.73% (4L/384d, PR #2602). All 4L/256d experiments are obsolete. Redirecting to 4L/384d experiments..

PR #2634, opened 2026-04-21
DrivAerML: 4L/256d+T_max=30 — gradient accumulation=2 (effective batch=2)

tcapelle:
## Hypothesis

With batch-size=1 forced by OOM constraints, the model sees only one geometry per gradient step. Gradient accumulation over 2 steps would effectively double the batch size to 2 without increasing peak memory usage. A larger effective batch may stabilize gradient estimates and improve convergence, particularly for the noisy DrivAerML geometry distribution.

Note: `train.py` may not yet support a `--gradient-accumulation` flag. Please check if it exists; if not, add minimal support: accumulate gradients over N steps before calling `optimizer.step()`, and divide the loss by N before backward. Only implement if straightforward.

## Instructions

First check if gradient accumulation is already supported:
```bash
python target/icml2026/train.py --help | grep -i accum
```

If `--gradient-accumulation` (or `--grad-accum-steps`) exists, run:
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --gradient-accumulation 2 \
  --wandb_group drivaerml-4L256d-grad-accum
```

If gradient accumulation is NOT supported yet, implement it in `train.py` by:
1. Finding the training loop where `loss.backward()` and `optimizer.step()` are called
2. Adding a `grad_accum_steps` argument (default=1)
3. Dividing the loss by `grad_accum_steps` before backward
4. Only calling `optimizer.step()` and `optimizer.zero_grad()` every `grad_accum_steps` steps
5. Adjusting `--max-train-batches` to 788 (394 × 2) if needed so effective gradient steps remain at 394

Then run with `--gradient-accumulation 2` (or whatever the flag is named).

Report back with best `val/surface_rel_l2_pct` and the W&B run URL.

## Baseline

Current best DrivAerML metrics (PR #2593, W&B run: 3aaevlho):
- val/surface_rel_l2_pct: **12.70%**
- Architecture: 4L/256d, AdamW lr=5e-4, T_max=30, no-EMA, Fourier
- Effective batch: 1 (batch-size=1, no accumulation)

Reproduce baseline:
```bash
```

### DIFF
--- a/experiments/shinobu/hypothesis.txt
+++ b/experiments/shinobu/hypothesis.txt
@@ -1,0 +1,1 @@
+drivaerml-4L256d-grad-accum2


tcapelle:
ADVISOR: Closing PR #2634 because ADVISOR: Closing — DrivAerML baseline is now 5.73% (4L/384d, PR #2602). All 4L/256d experiments are obsolete. Redirecting to 4L/384d experiments..

PR #2635, opened 2026-04-21
TandemFoil: T_max=10 cosine annealing + input noise augmentation

tcapelle:
## Hypothesis

T_max=10 cosine annealing produces strong phase-transition-like MAE drops at each LR trough for TandemFoil. However, the model may be overfitting between cycles — the geometry and pressure fields in TandemFoil are relatively low-dimensional and the model may memorize training cases rather than learning generalizable physics.

Adding small Gaussian input noise (σ=0.01) to the input node features at training time acts as a regularizer, smoothing the loss landscape and preventing over-reliance on exact coordinate values. This is analogous to dropout in the feature space and is especially effective when training with a cyclic LR that revisits high-LR phases (which can amplify overfitting).

Expected outcome: better generalization at each cosine trough, reducing val MAE below the 75.59 baseline.

## Instructions

```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --asinh-pressure \
  --residual-prediction \
  --epochs 999 \
  --input-noise 0.01 \
  --wandb_group tandem-tmax10-variants
```

Key change vs baseline: `--input-noise 0.01` added (Gaussian noise with σ=0.01 applied to input node features during training only).

All other flags match the current TandemFoil best configuration (PR #2490).

## Baseline

- **Dataset:** TandemFoil
- **val MAE (primary metric):** 75.59
- **Baseline W&B run:** 77yoba65
- **Baseline PR:** #2490 (frieren)
- **Reproduce command:**
```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --asinh-pressure \
  --residual-prediction \
  --epochs 999
```


tcapelle:
ADVISOR: Closing — noise augmentation at 3L/192d is obsolete. Baseline moved to 5L/256d (52.81). chrome, please pick up your next assignment.

PR #2636, opened 2026-04-21
AirfRANS: T_max=15 cosine annealing — mid-range cycle sweep

tcapelle:
## Hypothesis

AirfRANS shows phase-transition-like MAE drops at cosine LR troughs. The current best (PR #2556, emma) uses T_max=10, producing phase transitions around epoch 40. We are running a T_max sweep to find the optimal cycle length.

T_max=15 places the first trough at epoch ~47, allowing more training steps before the first aggressive LR drop. This gives the model more time to converge in the initial high-LR exploration phase before being driven into a local minimum by the cosine descent. Compared to T_max=10, this may find a better initial basin, leading to a lower floor at each subsequent trough.

This is part of the `airfrans-tmax-sweep` group alongside other T_max values being tested concurrently.

## Instructions

```bash
  --dataset airfrans \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 15 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 3 \
  --model-hidden-dim 192 \
  --model-heads 3 \
  --epochs 999 \
  --wandb_group airfrans-tmax-sweep
```

Key change vs baseline: `--cosine_t_max 15` (vs T_max=10 in baseline).

All other flags match the current AirfRANS best configuration (PR #2556).

## Baseline

- **Dataset:** AirfRANS
- **val surface_mse (primary metric):** 0.0248
- **Baseline W&B run:** 7qre8z5x
- **Baseline PR:** #2556 (emma)
- **Reproduce command:**
```bash
  --dataset airfrans \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 3 \
  --model-hidden-dim 192 \
  --model-heads 3 \
  --epochs 999
```


tcapelle:
ADVISOR: Closing — T_max=15 without grad-clip is obsolete. Baseline now 0.01419 via grad-clip=1.0. winry, please pick up your next assignment.

PR #2639, opened 2026-04-21
AirfRANS: Fourier+3L/192d+T_max=8 — interpolating T_max

tcapelle:
## Hypothesis

The current best AirfRANS result uses T_max=10 with cosine annealing, achieving a phase transition drop around epoch 40. We have also tested T_max=5 (likely too short) and T_max=50 (too long, no sharp phase transition). T_max=8 sits between 5 and 10 and may find the sweet spot: faster cosine cycles (first trough at epoch ~32) could trigger the phase transition earlier, allowing more beneficial cycles within the epoch budget. If the phase transition at T_max=10 occurs at trough epoch ~40, then T_max=8 would hit the first trough at ~32 and potentially get more improvement steps within 999 epochs.

## Instructions

```bash
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 8 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group airfrans-tmax-sweep
```

Let it run for at least 100 epochs. The key metric is the minimum `val/surface_mse` achieved and the epoch at which the phase transition (sharp drop) occurs. Report back with:
- Best `val/surface_mse` and epoch
- Epoch at which the phase transition (sharp drop in val loss) occurred
- W&B run URL

## Baseline

Current best AirfRANS metrics (PR #2556, W&B run: 7qre8z5x):
- val/surface_mse: **0.0248**
- Architecture: 3L/192d (default), AdamW lr=5e-4, T_max=10, no-EMA, Fourier
- Phase transition occurred at ~epoch 40 (4th cosine trough with T_max=10)

Reproduce baseline:
```bash
cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 5e-4 --cosine_t_max 10 --no-use-ema --enable-fourier --epochs 999
```

### DIFF
--- a/experiments/roy/hypothesis.txt
+++ b/experiments/roy/hypothesis.txt
@@ -1,0 +1,1 @@
+airfrans-tmax8


tcapelle:
ADVISOR: Closing — T_max=8 at lr=7e-4 without grad-clip is a confirmed dead end (kaneda's #2679 achieved 0.025). Baseline is 0.01419. roy, please pick up your next assignment.

PR #2653, opened 2026-04-21
TandemFoil: T_max=10 — cosine eta_min=1e-5 (LR floor at troughs)

tcapelle:
## Hypothesis
The current TandemFoil baseline uses cosine annealing with T_max=10 but the default eta_min is 0 (LR decays all the way to zero at each trough). Near-zero learning rates can cause the optimizer to stall in local minima between cosine cycles. Setting eta_min=1e-5 ensures the model keeps making meaningful updates even at the trough of each cosine cycle, potentially allowing it to escape shallow local minima and continue improving. This is a low-risk, low-complexity change with well-established empirical support in the literature (e.g. SGDR paper, Loshchilov & Hutter 2017).

eta_min=1e-5 relative to lr=3e-4 gives a floor of ~3.3% of peak LR — enough to keep training alive without disrupting the beneficial "restart" effect of cosine cycling.

## Instructions
Run TandemFoil training identical to the baseline but add `--cosine_eta_min 1e-5` (or equivalent flag). Keep all other hyperparameters exactly as the baseline:

```bash
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --cosine_eta_min 1e-5 \
  --slices 64 \
  --epochs 999 \
  --wandb_group tandem-tmax10-etamin
```

If `--cosine_eta_min` flag doesn't exist, check train.py for the correct flag name (could be `--eta_min`, `--lr_min`, `--min_lr`, etc.) and adjust accordingly. The goal is to set the minimum LR at the cosine trough to 1e-5 instead of 0.

If no such flag exists at all, add it: in the cosine scheduler instantiation, set `eta_min=1e-5` directly in `CosineAnnealingLR(optimizer, T_max=..., eta_min=1e-5)`.

Report val_primary/surface_pressure_mae and compare against baseline 75.59.

## Baseline
Current best metrics from BASELINE.md:
- val_primary/surface_pressure_mae: **75.59**
- W&B run: 77yoba65
- PR: #2490 (TandemFoil: T_max=10, Lion lr=3e-4, 3L/192d, slices=64, 14 epochs)
- Reproduce command: `cd target/icml2026 && python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 --cosine_t_max 10 --slices 64 --epochs 999`


tcapelle:
ADVISOR: Closing — cosine eta_min at 3L/192d is obsolete. Baseline moved to 5L/256d (52.81). ray, please pick up your next assignment.

PR #2641, opened 2026-04-21
DrivAerML: 4L/256d+T_max=30 — 5-epoch linear LR warmup

tcapelle:
## Hypothesis

Training starts at max LR (5e-4) which can destabilize early optimization — large gradient steps on a randomly initialized model may lead to poor initial loss landscape exploration. A 5-epoch linear warmup (0 → 5e-4) before cosine annealing begins may help the model find a better initialization landscape, leading to better convergence at the cosine annealing phase.

This is a standard training trick used in Transformers and ViTs (e.g. DeiT, MAE) that often helps when training from scratch.

## Instructions

Run from the repo root:

```bash
cd target/icml2026 && python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --warmup-epochs 5 \
  --wandb_group drivaerml-4L256d-training-tricks
```

**Note:** If `--warmup-epochs` doesn't exist, check `train.py` for the warmup flag name (may be `--lr-warmup-epochs`, `--warmup_epochs`, etc.) and use the correct flag.

## Baseline

- **Primary metric:** `val_primary/surface_rel_l2_pct` (lower is better)
- **Current best:** 12.70% (val) / 13.54% (test)
- **Best PR:** #2593 (shinji — Fourier+4L/256d+no-EMA+T_max=30, 45 epochs, AdamW lr=5e-4)
- **W&B run:** 3aaevlho (45 epochs, still converging at epoch cap)
- **Reproduce baseline:** `cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine_t_max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`

**CRITICAL OOM flags** (must include all of these):
- `--batch-size 1`
- `--drivaerml-train-surface-points 50000`
- `--drivaerml-eval-surface-points 50000`
[...]

PR #2660, opened 2026-04-21
DrivAerML: 4L/256d+T_max=30 — 5-epoch cosine warmup (--warmup-steps 1000)

tcapelle:
## Hypothesis

The current best model starts training at full lr=5e-4 immediately. A warmup phase (lr: 0 → 5e-4 linearly over the first ~5 epochs, then cosine decay with T_max=30) should stabilize early training by preventing large gradient updates before the model has oriented itself. Warmup is standard practice in transformer training and may smooth out early instability in the DrivAerML regime.

Note: tanjiro (#2641) is testing warmup via `--warmup-epochs`. This PR uses `--warmup-steps 1000` instead — a different interface that may be available. Report all flags you tried and which ones were accepted.

## Instructions

Run the standard 4L/256d DrivAerML config adding `--warmup-steps 1000` (approx 5 epochs at 394 batches/epoch = ~1970 steps, so 1000 steps ≈ 2.5 epochs of warmup):

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --warmup-steps 1000 \
  --wandb_group drivaerml-4L256d-warmup
```

If `--warmup-steps` is not recognised, try `--warmup-epochs 5`. If neither works, try `--warmup_steps 1000` or `--warmup_epochs 5` (underscore variant). Report every flag tried and every error message.

If no warmup flag exists at all, run the baseline command and note that warmup is not supported by train.py.

## Baseline

- **Current best:** 12.70% (val) / 13.54% (test) — PR #2593
- **Configuration:** 4L/256d, T_max=30, AdamW lr=5e-4, no-EMA, Fourier, 45 epochs
- **Baseline W&B run:** 3aaevlho
- **Reproduce baseline:**
  ```
  ```

## CRITICAL OOM Flags

Always include these or you will OOM:
```
--batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --model-heads 4
```


tcapelle:
ADVISOR: Closing PR #2660 because ADVISOR: Closing — DrivAerML baseline shifted to 5.73% (4L/384d, PR #2602). All 4L/256d experiments are now obsolete — the best 4L/256d result was 11.97%. Width scaling is the dominant lever. Redirecting your GPU to 4L/384d experiments. You'll receive a new assignment shortly..

PR #2614, opened 2026-04-21
AirfRANS: lr=3e-4 + T_max=10 (lower LR phase transition)

tcapelle:
## Hypothesis
The current best (PR #2556) uses lr=5e-4 + T_max=10 and achieves 0.0248 val surface_mse via a phase transition at epoch 40. A lower learning rate (lr=3e-4) with the same T_max=10 may produce a deeper, more stable phase transition — the oscillation amplitude is smaller so the LR trough is at 0 but the plateau before is lower, potentially enabling the model to settle into a better minimum. This pairs with senku's lr=1e-3 run as part of the `airfrans-tmax10-lr` group sweep to bracket the optimal LR.

## Instructions

Run the following command exactly:

```bash
cd target/icml2026 && python train.py \
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group airfrans-tmax10-lr
```

This uses the 3L/192d architecture (same as the winning baseline) but with lr=3e-4 instead of 5e-4. Report `val_primary/surface_mse` and the W&B run ID in PR comments. Note the epoch at which any phase transition occurred.

## Baseline

Current best metrics from BASELINE.md (PR #2556):
- **val_primary/surface_mse:** 0.0248 (target to beat)
- **W&B run:** 7qre8z5x (41 epochs, phase transition at epoch 40)
- **Architecture:** 3L/192d + Fourier + no-EMA + AdamW lr=5e-4 + T_max=10
- **Reproduce:** `cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 5e-4 --cosine_t_max 10 --no-use-ema --enable-fourier --epochs 999`

PR #2664, opened 2026-04-21
AirfRANS: 3L/256d width expansion + lr=3e-4 T_max=10

tcapelle:
## Hypothesis

Two independently confirmed improvements for AirfRANS are: (1) wider hidden dim (256d vs 128d) has shown gains in DrivAerML, and (2) lr=3e-4 beats lr=5e-4 on AirfRANS (PR #2614, 0.0197 vs prior baseline). This experiment compounds both changes simultaneously with AirfRANS: 3 layers × 256 hidden dim + lr=3e-4 + T_max=10. If both effects are additive, we should beat 0.0197. The risk is that 256d + 3 layers may slow convergence within the epoch budget, but at lr=3e-4 the model should still hit a good cosine trough.

## Instructions

```bash
  --dataset airfrans \
  --airfrans_task full \
  --layers 3 \
  --hidden-dim 256 \
  --lr 3e-4 \
  --cosine-t-max 10 \
  --no-use-ema \
  --enable-fourier \
  --wandb_group airfrans-3L256d-lr3e4-tmax10
```

Report: best `val_primary/surface_mse` achieved, at which epoch, and the W&B run ID. Also report whether a phase transition (sudden drop in surface_mse) occurred and if so, at which epoch.

## Baseline

Current best AirfRANS metrics (PR #2614, gilbert):
- `val_primary/surface_mse`: **0.0197**
- W&B run: `v5ka7832`
- Config: 3L/128d (default), lr=3e-4, T_max=10, 41 epochs, `--no-use-ema`, `--enable-fourier`, `--airfrans_task full`
- Reproduce: `cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --lr 3e-4 --cosine-t-max 10 --no-use-ema --enable-fourier`

External target: 0.0043


tcapelle:
ADVISOR: Closing PR #2664 because ADVISOR: Closing — 3L/256d confirmed dead end. senku, please pick up your next assignment..

PR #2599, opened 2026-04-21
DrivAerML: 4L/256d+T_max=30 — lr=3e-4 (lower LR)

tcapelle:
## Hypothesis
The current DrivAerML baseline uses lr=5e-4. Our best model (4L/256d, T_max=30) may benefit from a lower learning rate of 3e-4 to allow more careful convergence on the complex 3D car geometry. Lower LRs often reduce overshoot in the final epochs of cosine annealing, potentially improving surface pressure accuracy. This is a bracketing experiment alongside a higher-LR run (lr=8e-4) to find the optimal LR for this architecture.

## Instructions
Train the DrivAerML model with the baseline 4L/256d configuration but reduce the learning rate from 5e-4 to 3e-4:

```bash
cd target/icml2026 && python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 3e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-4L256d-lr
```

**CRITICAL flags — do NOT omit these:**
- `--batch-size 1`
- `--drivaerml-train-surface-points 50000`
- `--drivaerml-eval-surface-points 50000`
- `--max-train-batches 394`
- `--max-eval-batches 200`
- `--model-heads 4`
- `--no-use-ema`
- `--enable-fourier`
- `--wandb_group drivaerml-4L256d-lr` (groups this run with the lr=8e-4 run for comparison)

The only change from baseline is `--lr 3e-4` (was `5e-4`).

## Baseline
Current best DrivAerML metrics (PR #2550):
- **val_primary/surface_rel_l2_pct**: 12.96% (lower is better)
- **test_primary/surface_rel_l2_pct**: 14.41%
- Baseline W&B run: 8s5i8y06
- Baseline reproduce command:
  ```
[...]

PR #2666, opened 2026-04-21
AirfRANS: T_max=5 with confirmed best lr=3e-4

tcapelle:
## Hypothesis

T_max=5 was previously tested (PR #2599) with lr=5e-4 and showed no improvement over T_max=10. However, we now know that lr=3e-4 is significantly better than lr=5e-4 for AirfRANS. The interaction between LR magnitude and T_max cycle length may be important: with lr=3e-4, a shorter T_max=5 cycle creates more frequent cosine troughs but at a lower amplitude, which could trigger more phase transitions across the training run. This is a controlled retest: same config as the current best (PR #2614), only change is T_max=5 instead of T_max=10.

## Instructions

```bash
  --dataset airfrans \
  --airfrans_task full \
  --lr 3e-4 \
  --cosine-t-max 5 \
  --no-use-ema \
  --enable-fourier \
  --wandb_group airfrans-lr3e4-tmax5
```

Report: best `val_primary/surface_mse` achieved, at which epoch, and the W&B run ID. Also report whether a phase transition (sudden drop in surface_mse) occurred and if so, at which epoch — this is key data for understanding the T_max interaction with lr.

## Baseline

Current best AirfRANS metrics (PR #2614, gilbert):
- `val_primary/surface_mse`: **0.0197**
- W&B run: `v5ka7832`
- Config: 3L/128d (default), lr=3e-4, T_max=10, 41 epochs, `--no-use-ema`, `--enable-fourier`, `--airfrans_task full`
- Reproduce: `cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --lr 3e-4 --cosine-t-max 10 --no-use-ema --enable-fourier`

External target: 0.0043


tcapelle:
ADVISOR: Closing PR #2666 because ADVISOR: Closing — T_max=5 without grad-clip is obsolete. Baseline now 0.01419. thorfinn, please pick up your next assignment..

PR #2668, opened 2026-04-21
AirfRANS: lr=3e-4+T_max=10 — weight decay=1e-2

tcapelle:
## Hypothesis
The current AirfRANS best (PR #2614, val_primary/surface_mse=0.0197) uses AdamW with lr=3e-4 and T_max=10, but does not explicitly set weight decay. AdamW's key strength over Adam is its decoupled weight decay regularization. Increasing weight decay to 1e-2 (a commonly effective value for AdamW) should provide stronger L2 regularization, reducing overfitting and potentially improving surface MSE generalization. This complements the lr=3e-4 finding and tests whether the optimizer's native regularization mechanism is being fully utilized.

## Instructions
Run the following command from `target/icml2026/`:

```bash
python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 3e-4 --cosine_t_max 10 --no-use-ema --enable-fourier --weight-decay 1e-2 --epochs 999 --wandb_group airfrans-lr3e4-wd
```

Key changes from baseline:
- `--lr 3e-4` (same as recent best)
- `--cosine_t_max 10` (same as recent best)
- `--weight-decay 1e-2` (NEW — explicit AdamW weight decay of 0.01)
- `--wandb_group airfrans-lr3e4-wd` (for grouping weight decay experiments)

If `--weight-decay` is not a recognized flag, check `train.py --help` for the correct flag name (may be `--wd`, `--weight_decay`, `--adamw-wd`, or similar). Report which flag name worked.

Report in your PR comment:
- val_primary/surface_mse (primary metric, lower is better)
- All surface metrics in a table
- W&B run ID and link
- What weight decay value was used and the exact flag name

## Baseline
Current best AirfRANS result (PR #2614):
- val_primary/surface_mse: **0.0197**
- Reproduce command:
```bash
cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 3e-4 --cosine_t_max 10 --no-use-ema --enable-fourier --epochs 999
```


tcapelle:
ADVISOR: Closing PR #2668 because ADVISOR: Closing — lr=3e-4+WD=1e-2 without grad-clip is obsolete. The fern experiment (PR #2709) tests the correct compound: grad-clip+WD. historia, please pick up your next assignment..

PR #2670, opened 2026-04-21
DrivAerML: 1000 batches/epoch — upper bound of data saturation sweep

tcapelle:
## Hypothesis

Find the saturation point of the batches-per-epoch lever. If 600 batches/epoch gave 11.97% (PR #2645), 1000 batches/epoch might push below 11% — each epoch sees ~67% more car configurations than the 600-batch winner, and ~2.5x more than the original 394-batch config. This is the upper bound of the sweep (frieren tests 800, taki tests 1000) to map the full curve.

## Instructions

Run with `--max-train-batches 1000` (up from 600 in PR #2645). All other hyperparameters remain identical to the current best.

**Full command to run:**
```bash
  --epochs 999 --batch-size 1 \
  --max-train-batches 1000 --max-eval-batches 200 \
  --agent taki --wandb-name "taki/drivaerml-1000batches" \
  --wandb-group "drivaerml-batches-sweep"
```

## Baseline to beat

- **Current best val:** 11.97% (PR #2645, 600 batches/epoch)
- **W&B run:** dar47nwl

## W&B group

`drivaerml-batches-sweep` — this groups the 600/800/1000-batch sweep together in W&B for easy comparison.

## Expected outcome

1000 batches will make each epoch ~67% slower than 600-batch runs. The 30-min timeout will therefore produce fewer epochs. Key question: does more data per epoch compensate for fewer epochs total? Report the best val metric and epoch it occurred at. Compare against frieren's 800-batch result to map the saturation curve.


tcapelle:
STUDENT taki:

## Results

| Metric | Baseline (600 batches) | This run (1000 batches) |
|--------|----------------------|------------------------|
| `val_primary/surface_rel_l2_pct` | **11.97%** | 12.40% |
| `test_primary/surface_rel_l2_pct` | 13.03% | **12.85%** |
| Epochs completed | ~38+ (est.) | 23 |
| Best val epoch | — | 22 |

**W&B run:** `qju0pw3q` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/qju0pw3q))
**W&B group:** `drivaerml-batches-sweep`
**Peak memory:** Not captured (GPU freed before measurement)

**Command used:**
```bash
  --epochs 999 --batch-size 1 \
  --max-train-batches 1000 --max-eval-batches 200 \
  --agent taki --wandb-name "taki/drivaerml-1000batches" \
  --wandb-group "drivaerml-batches-sweep"
```

## What happened

**Val metric did not beat the baseline** (12.40% vs 11.97%), but **test metric improved** (12.85% vs 13.03%).

The core tradeoff played out as expected: 1000 batches/epoch sees ~67% more data per epoch than 600 batches, but each epoch takes proportionally longer. Within the 30-minute timeout, we completed only 23 epochs vs the ~38+ the 600-batch baseline likely achieved. With `cosine-t-max=30`, we never completed a full cosine cycle (only reached epoch 23/30), which means the LR schedule didn't get to its minimum phase where the best convergence typically happens.

The convergence curve was still clearly trending downward at timeout — the run was not saturated. The cosine LR oscillations every 3 epochs (T_max=30 mapped to 3-epoch cycles due to the restart schedule) caused visible val metric bumps at epochs 12, 15, 18, and 21.

The better test metric despite worse val suggests the model may generalize slightly better with more data diversity per epoch, but needs more total training time to fully converge.

## Analysis for the saturation sweep

For the batches-per-epoch curve:
- **394 batches (original):** baseline before PR #2645
- **600 batches (PR #2645):** 11.97% val / 13.03% test — current best
- **1000 batches (this run):** 12.40% val / 12.85% test — fewer epochs, better test generalization

The 1000-batch config is **data-rich but time-poor**. It's likely the better configuration given enough training time, but the 30-minute timeout creates a diminishing returns problem — beyond ~600 batches/epoch, the per-epoch time cost outweighs the per-epoch data benefit within the fixed time budget.

## Suggested follow-ups

1. **Longer timeout test:** If the time budget were relaxed, 1000 batches would likely surpass 600 since the convergence curve wasn't saturated and test generalization was already better.
2. **Adjust cosine-t-max for 1000 batches:** With only 23 epochs possible, `cosine-t-max=15` or `cosine-t-max=20` would let the LR schedule complete at least one full cycle, potentially improving convergence efficiency.
3. **800 batches (frieren's run):** This should reveal whether 800 is the sweet spot — enough data diversity with enough epochs to complete the LR schedule.

tcapelle:
ADVISOR: Closing PR #2670 because ADVISOR: Closing — 4L/256d with 1000 batches reached 12.40%. All 4L/256d results are obsolete with the new 5.73% baseline (4L/384d). Useful data: 1000 batches at 4L/256d gets only 23 epochs (too slow)..

PR #2672, opened 2026-04-21
DrivAerML: 600 batches/epoch + weight decay 1e-2 compound

tcapelle:
## Hypothesis

Compound the 600-batch improvement (PR #2645, 11.97%) with stronger weight decay (1e-2 vs the AdamW default of ~1e-4). With more data per epoch, stronger regularization may improve generalization by preventing the model from memorizing the most frequently-seen car configurations. WD=1e-2 showed promising convergence on TandemFoil with Lion optimizer — here we test whether the same regularization boost transfers to DrivAerML with AdamW and 600 batches/epoch.

## Instructions

Add `--weight-decay 1e-2` to the PR #2645 winning configuration. All other settings are identical to the current best.

**Full command to run:**
```bash
  --epochs 999 --batch-size 1 \
  --max-train-batches 600 --max-eval-batches 200 \
  --agent norman --wandb-name "norman/drivaerml-600batch-wd1e2" \
  --wandb-group "drivaerml-600batch-compounds"
```

## Baseline to beat

- **Current best val:** 11.97% (PR #2645, 600 batches/epoch, default WD)
- **W&B run:** dar47nwl

## W&B group

`drivaerml-600batch-compounds` — groups experiments compounding other improvements on top of the 600-batch baseline.

## Expected outcome

If WD=1e-2 is too strong, training may slow or diverge. If well-tuned, we expect val below 11.97%. Monitor val/loss and training loss carefully — if training loss diverges, note the epoch. Report best val metric, the epoch at which it occurred, and whether training appeared stable.


tcapelle:
ADVISOR: Closing PR #2672 because ADVISOR: Closing — DrivAerML baseline shifted to 5.73% (4L/384d, PR #2602). The 4L/256d+600batch config cannot compete. Redirecting to 4L/384d experiments..

PR #2646, opened 2026-04-21
AirfRANS: T_max=10 — lr=7e-4 (between baseline and 1e-3)

tcapelle:
## Hypothesis

The phase transition mechanism at T_max=10 / epoch 40 is our primary lever for improving val_primary/surface_mse. The transition depth is stochastic and appears LR-sensitive:
- lr=5e-4 (baseline): 0.0207 best observed
- lr=3e-4 (gilbert, in-flight): unknown
- lr=1e-3 (senku, in-flight): unknown

lr=7e-4 fills the critical gap between baseline and the 1e-3 run. A slightly higher LR may push the model into a deeper basin at each cosine trough by providing more aggressive gradient steps before the LR drops to zero. This is a fine-grained sweep of the most sensitive hyperparameter for this phase-transition mechanism.

## Instructions

Run the following command:

```bash
cd target/icml2026 && python train.py \
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 7e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group airfrans-tmax10-lr-sweep
```

Key change vs baseline:
- `--lr 7e-4` (up from 5e-4, between baseline and the lr=1e-3 senku run)
- All other flags identical to baseline T_max=10 run (3L/192d, AdamW, Fourier)

**Critical:** The phase transition typically occurs around epoch 40 (cosine trough of cycle 4). Monitor val_primary/surface_mse carefully. Report:
1. Best val_primary/surface_mse and at which epoch it occurred
2. Whether the phase transition occurred and at what epoch
3. Training stability (any divergence or spikes?)

## Baseline

- **Dataset:** AirfRANS (full task)
- **Primary metric:** `val_primary/surface_mse` (lower is better)
- **Current best:** 0.0207 (val) — PR #2617 (T_max=10, lr=5e-4, 3L/192d, AdamW, epoch 40)
- **Previous:** 0.0248 (PR #2556) — same config, different seed, phase transition is stochastic
- **External target:** 0.0043 (~4.8x gap remaining)
- **Baseline W&B run:** z7t3ibwi
- **In-flight LR experiments:** lr=3e-4 (gilbert #2614), lr=1e-3 (senku #2615)
[...]

PR #2679, opened 2026-04-21
AirfRANS: lr=7e-4 + T_max=8 (faster cosine cycling)

tcapelle:
## Hypothesis

Test T_max=8 with the winning lr=7e-4. T_max=10 produced the baseline 0.01841 with the phase transition at epoch 35. T_max=8 creates faster cosine cycling (more restarts per epoch), which may trigger the phase transition earlier or at a qualitatively different (potentially deeper) point. The AirfRANS phase transition is sensitive to the exact timing of LR troughs — a faster cycle could hit the same "unlock" condition with more training dynamics behind it. T_max=8 at lr=7e-4 is a previously untested combination.

## Instructions

```bash
  --optimizer adamw --lr 7e-4 --cosine-t-max 8 \
  --no-use-ema --enable-fourier --epochs 999 \
  --agent kaneda --wandb-name "kaneda/airfrans-lr7e4-tmax8" \
  --wandb-group "airfrans-lr7e4-compounds"
```

Key change vs baseline: `--cosine-t-max 8` instead of 10. All other settings identical to the winning PR #2646 configuration.

Report the best `val_primary/surface_mse` and the epoch it was achieved. Note the epoch at which the phase transition occurred (large drop in val loss), and compare to epoch 35 in the baseline.

## Baseline

Current best AirfRANS metrics (PR #2646, emma — lr=7e-4+T_max=10):
- **val_primary/surface_mse:** 0.01841 ← target to beat
- **full_val/surface_mse_p:** 0.0735
- **full_val/volume_mse:** 0.1331
- **W&B run:** 3pbxocca (41 epochs, best at epoch 35 — phase transition)
- **Reproduce baseline:** `cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 7e-4 --cosine_t_max 10 --no-use-ema --enable-fourier --epochs 999`

Context on T_max history: T_max=10 > T_max=50 for AirfRANS (faster cycling consistently wins). T_max=8 is one step faster than the current best.

External target: 0.0043 — 4.3x gap remaining.


tcapelle:
STUDENT kaneda:

## Results

### Experiment: AirfRANS lr=7e-4 + T_max=8 (faster cosine cycling)

| Metric | Baseline (T_max=10, PR #2646) | T_max=8 (this run) | Delta |
|--------|-------------------------------|---------------------|-------|
| **val_primary/surface_mse** | **0.01841** (epoch 35) | **0.02516** (epoch 38) | +37% worse |
| full_val/surface_mse_p | 0.0735 | 0.1006 | +37% worse |
| full_val/volume_mse | 0.1331 | 0.1192 | **-10% better** |
| Epochs | 41 (T_max=10) | 41 (T_max=8) | — |

### Test metrics (final checkpoint, epoch 41):
- test_primary/surface_mse: 0.0556
- full_test/surface_mse_p: 0.2224
- full_test/volume_mse: 0.1509

### Command
```bash
  --optimizer adamw --lr 7e-4 --cosine-t-max 8 \
  --no-use-ema --enable-fourier --epochs 999 \
  --agent kaneda --wandb-name "kaneda/airfrans-lr7e4-tmax8" \
  --wandb-group "airfrans-lr7e4-compounds"
```

### Peak memory: ~36 GB on RTX PRO 6000
### W&B run: `nvclyp0t`

## What happened

**T_max=8 did not beat the baseline.** Best val_primary/surface_mse was 0.02516 vs baseline 0.01841 (37% worse).

**Phase transition occurred earlier:** The baseline's phase transition happened at epoch 35 (sharp drop to 0.01841). With T_max=8, a phase transition began around epoch 18-21 (val_primary dropped from 0.180 → 0.045 over ~5 epochs), significantly earlier than the baseline. However, the transition was:
1. **Less sharp** — it unfolded gradually over epochs 18-38 rather than a single dramatic drop
2. **Much more oscillatory** — val_primary swung wildly between good epochs (0.025-0.045) and bad epochs (0.10-0.19) throughout training, a pattern not seen with T_max=10
3. **Did not reach as deep** — best of 0.025 vs 0.018

**Volume MSE actually improved** over the baseline (0.119 vs 0.133), suggesting T_max=8 captures some volume features better, but at the cost of surface accuracy.

The heavy oscillation is likely caused by the fast cosine cycling (T_max=8 with per-batch stepping = 8-batch cycles). The model never has long enough at a low learning rate to fully converge to a sharp surface representation before the LR ramps back up.

## Suggested follow-ups

1. **T_max=6 or T_max=5**: Even faster cycling may trigger the phase transition earlier, but the oscillation problem suggests diminishing returns. T_max=8 already shows too much instability.
2. **T_max=9**: A more conservative step from T_max=10 might capture the "earlier transition" benefit without the oscillation problem.
3. **T_max=8 with eta_min > 0**: Adding a non-zero minimum learning rate (e.g. eta_min=1e-5) could dampen the extreme oscillation at cycle troughs.
4. **T_max=8 with longer training**: The metric was still improving at epoch 38 — with more epochs, it might eventually match or beat the baseline, but with more oscillation.

tcapelle:
ADVISOR: Closing PR #2679 because ADVISOR: Closing — val_primary/surface_mse=0.025 is 64% worse than baseline 0.0153 (now 0.01419). T_max=8 at lr=7e-4 doesn't help; the shorter cosine cycle doesn't find deep enough basins. The winning direction is grad-clip=1.0 at lr=7e-4 (PR #2680 merged with 0.01419). kaneda, please pick up your next assignment..

PR #2680, opened 2026-04-21
AirfRANS: lr=7e-4 + T_max=10 + grad-clip=1.0 (spike prevention)

tcapelle:
## Hypothesis

The winning lr=7e-4 run (PR #2646) showed catastrophic gradient spikes at cosine LR peaks: epochs 26, 28, 38, 41 all spiked to 0.23-0.27 val_primary/surface_mse. Gradient clipping (max_norm=1.0) should prevent these spikes, allowing the model to consolidate gains after the phase transition. 

The key insight: the epoch 38 spike (0.264) likely disrupted what could have been an even better epoch 40 trough (only reached 0.031 vs 0.018 at epoch 35). Preventing that spike by clipping could allow a deeper epoch 40 basin. The phase transition mechanism appears to require a momentary drop in the loss landscape — grad-clip preserves those deep troughs by preventing destructive gradient steps from undoing progress.

## Instructions

Run the following command exactly:

```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 python train.py \
  --dataset airfrans --airfrans-task full \
  --optimizer adamw --lr 7e-4 --cosine-t-max 10 \
  --grad-clip 1.0 \
  --no-use-ema --enable-fourier --epochs 999 \
  --agent haku --wandb-name "haku/airfrans-lr7e4-gradclip" \
  --wandb-group "airfrans-lr7e4-compounds"
```

Key change vs baseline: added `--grad-clip 1.0` to the winning lr=7e-4+T_max=10 configuration.

Report the best `val_primary/surface_mse` and the epoch it was achieved. Specifically note:
1. Whether the spikes at cosine LR peaks (previously ~0.23-0.27) are reduced
2. The depth of the epoch 40 trough vs the baseline's epoch 40 trough (0.031)

## Baseline

Current best AirfRANS metrics (PR #2646, emma — lr=7e-4+T_max=10):
- **val_primary/surface_mse:** 0.01841 ← target to beat
- **full_val/surface_mse_p:** 0.0735
- **full_val/volume_mse:** 0.1331
- **W&B run:** 3pbxocca (41 epochs, best at epoch 35 — phase transition; epoch 38 spike to 0.264, epoch 40 trough only 0.031)
- **Reproduce baseline:** `cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 7e-4 --cosine_t_max 10 --no-use-ema --enable-fourier --epochs 999`

External target: 0.0043 — 4.3x gap remaining.

PR #2595, opened 2026-04-21
TandemFoil: T_max=10 + 5L/256d (deep model capacity)

tcapelle:
## Hypothesis

4L/256d was transformative for DrivAerML (33.65%→12.96%, -61.5% relative). The question is: does greater model depth also benefit TandemFoil? The current TandemFoil best uses 3L/192d (default). Increasing to 5L/256d adds both depth and width, which should improve the model's capacity to resolve complex aerodynamic pressure distributions across tandem foil geometries.

Note: 6L has been observed to diverge, so 5L is the deepest safe option.

## Instructions

```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --optimizer lion \
  --lr 3e-4 \
  --cosine_t_max 10 \
  --no-use-ema \
  --model_slices 64 \
  --model-layers 5 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --enable-cp-panel-tandem-only \
  --asinh-pressure \
  --residual-prediction \
  --enable-pressure-prior-addition \
  --epochs 999 \
  --wandb_group tandem-5L256d
```

Report: val_primary/surface_pressure_mae, test_primary/surface_pressure_mae, per-split MAEs, best epoch, and whether the model was still improving at timeout.

## Baseline

| Metric | Current Best (PR #2490) |
|--------|------------------------|
| val_primary/surface_pressure_mae | **75.59** |
| test_primary/surface_pressure_mae | 72.12 |
| Per-split test MAE | single_in_dist=72.33, geom_camber_rc=76.01, geom_camber_cruise=70.80, re_rand=69.34 |
| Epochs | 14 (30-min timeout, still improving) |
| W&B run | 77yoba65 |
| Config | Lion lr=3e-4, T_max=10, model_slices=64, Fourier+physics, no-EMA, 3L/192d (default) |

**Target:** beat val < 75.59

**Reproduce baseline:**
```bash
cd target/icml2026 && python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 --cosine_t_max 10 --no-use-ema --model_slices 64 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999
```

PR #2706, opened 2026-04-21
TandemFoil: 5L/256d + T_max=20 — reduce oscillation

tcapelle:
## Hypothesis

The winning 5L/256d run (PR #2595) achieved 52.81 but showed high oscillation in the range of 52–85 with T_max=10. The fast cosine cycle frequency is causing LR spikes that prevent the model from settling at deep basins. T_max=20 doubles the cycle length, reducing cycle frequency from ~6 restarts per 67 epochs to ~3. This should produce a more consistent downward convergence envelope and potentially find a deeper minimum with less variance — mirroring the T_max=10→30 improvement seen on the 3L/192d architecture (82.65→78.81).

## Instructions

Run the same 5L/256d configuration as PR #2595, but with `--cosine-t-max 20` instead of `--cosine-t-max 10`. Everything else stays identical (Lion optimizer, lr=3e-4, 5 layers, 256 hidden dim, 4 heads, all physics features).

Run with `--epochs 999` and let the 180-min timeout govern duration. Report:
- Best val `surface_pressure_mae` and at which epoch it occurred
- The oscillation range (min–max val metric) to compare to 52–85 from T_max=10
- W&B run ID and link

## Baseline

- **TandemFoil current best:** 52.81 val / 55.25 test (PR #2595, 5L/256d, Lion lr=3e-4, T_max=10, 67 epochs, 180-min)
- **AirfRANS current best:** 0.01419 (PR #2680, lr=7e-4+grad-clip=1.0)
- **DrivAerML current best:** 5.73% (PR #2602, 4L/384d/6H, T_max=30, 151 epochs)
- **Baseline W&B run:** l5kggnbg (PR #2595)

## Reproduce command

```bash
  --dataset tandemfoil --optimizer lion --lr 3e-4 \
  --cosine-t-max 20 --no-use-ema --model-slices 64 \
  --model-layers 5 --model-hidden-dim 256 --model-heads 4 \
  --enable-pressure-prior-addition --epochs 999 \
  --wandb-group tandem-5L256d-tmax \
  --wandb-name "sasuke/tandem-5L256d-tmax20"
```

> **Note:** `--epochs 999` and `--no-use-ema` are both mandatory. Default epochs=2 will not train long enough. EMA has a known bug.


tcapelle:
STUDENT sasuke:

## Results

### Key Metrics

| Metric | Baseline (T_max=10, PR #2595) | This run (T_max=20) | Delta |
|--------|------|----------|-------|
| Best val `surface_pressure_mae` | 52.81 (epoch 67) | 56.31 (epoch 66) | +3.50 (worse) |
| Test `surface_pressure_mae` | 55.25 | 57.22 | +1.97 (worse) |
| Epochs completed | 67 | 66 | -1 |
| Val oscillation range (last 20 ep) | 52–85 (range ~33) | 56–70 (range ~14) | Reduced by ~57% |

### Oscillation Analysis

The hypothesis that T_max=20 reduces oscillation is **confirmed** — the oscillation range in the last 20 epochs was approximately 56–70 (range ~14), compared to the baseline's reported 52–85 (range ~33). This is a ~57% reduction in oscillation amplitude.

However, the hypothesis that reduced oscillation leads to deeper minima is **not confirmed**. The best val metric (56.31) is 3.50 worse than the baseline (52.81), and the test metric (57.22) is 1.97 worse than baseline (55.25).

### Important Observation: Per-Step LR Scheduling

The `CosineAnnealingLR` scheduler is stepped **per training step** (train.py:1094), not per epoch. With 750 steps/epoch and T_max=20, each cosine cycle completes every 40 steps (~18.75 full cycles per epoch). This means both T_max=10 and T_max=20 operate as high-frequency cyclic schedules, not the slow epoch-scale cosine warmrestarts the PR hypothesis assumes. The comparison is still valid (T_max=20 cycles are 2x longer than T_max=10), but the physical interpretation differs from the PR description.

### What Happened

The slower cycling (T_max=20 vs T_max=10) successfully reduced epoch-to-epoch oscillation as expected, but this came at the cost of convergence speed. The model reached the 60s zone about 10 epochs later than the baseline reached the 50s zone with T_max=10. The faster T_max=10 cycling appears to act as a more aggressive exploration mechanism that ultimately finds better minima, despite the higher oscillation.

### Experiment Details

- **Command:** `cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 python train.py --dataset tandemfoil --optimizer lion --lr 3e-4 --cosine-t-max 20 --no-use-ema --model-slices 64 --model-layers 5 --model-hidden-dim 256 --model-heads 4 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999 --wandb-group tandem-5L256d-tmax --agent sasuke --wandb-name "sasuke/tandem-5L256d-tmax20"`
- **Peak VRAM:** ~64.8 GB (GPU 0)
- **W&B run ID:** u7cdrpkc
- **W&B link:** https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/u7cdrpkc

### Suggested Follow-ups

1. **Try T_max=15** — a middle ground between T_max=10 and T_max=20 that might balance exploration (shorter cycles) with stability (less oscillation).
2. **Fix the per-step scheduling** — if the intent was per-epoch cosine restarts, moving `scheduler.step()` outside the training step loop would give the original behavior. T_max=20 with epoch-level stepping might perform differently.
3. **Snapshot Weight Averaging (SWA)** — since the per-step cycling acts like a form of stochastic weight averaging, explicitly using SWA or averaging checkpoint weights from the low-LR phases could extract better performance from the same training trajectory.

tcapelle:
ADVISOR: Closing PR #2706 because ADVISOR: Closing — 5L/256d+T_max=20 at lr=3e-4 is superseded by lr=2e-4 finding. All TandemFoil should use lr=2e-4..

PR #2707, opened 2026-04-21
AirfRANS: lr=7e-4 + grad-clip=0.5 — tighter gradient constraint

tcapelle:
## Hypothesis

The grad-clip=1.0 breakthrough (PR #2680, 0.01419) revealed that 91–98% of batches are being clipped every epoch, with mean gradient norms of 10–22. This means clip=1.0 is very loose — only shaving the most extreme spikes. A tighter threshold of clip=0.5 should constrain gradients more aggressively, further reducing the destructive spike magnitude at cosine LR peaks (baseline 0.23–0.27 → clipped 0.15–0.17 with clip=1.0). A tighter clip could push trough depths even lower by allowing more stable exploration of deep basins without the spike-induced escapes.

## Instructions

Run the same configuration as PR #2680 (lr=7e-4, T_max=10, Fourier, no-EMA), but change `--grad-clip 1.0` to `--grad-clip 0.5`.

Run with `--epochs 999` and let the 30-min timeout govern duration. Report:
- Best val `surface_mse` and at which epoch it occurred
- Percentage of batches clipped per epoch (compare to 91–98% from clip=1.0)
- W&B run ID and link

## Baseline

- **AirfRANS current best:** 0.01419 val / 0.01513 test (PR #2680, lr=7e-4+grad-clip=1.0, T_max=10, 41 epochs, 30-min)
- **TandemFoil current best:** 52.81 (PR #2595, 5L/256d, Lion lr=3e-4, T_max=10)
- **DrivAerML current best:** 5.73% (PR #2602, 4L/384d/6H, T_max=30, 151 epochs)
- **Baseline W&B run:** 48ldl625 (PR #2680, grad-clip=1.0)

## Reproduce command

```bash
  --dataset airfrans --airfrans-task full --optimizer adamw \
  --lr 7e-4 --cosine-t-max 10 --grad-clip 0.5 \
  --no-use-ema --enable-fourier --epochs 999 \
  --wandb-group airfrans-gradclip-sweep \
  --wandb-name "haku/airfrans-lr7e4-gradclip05"
```

> **Note:** `--epochs 999` and `--no-use-ema` are both mandatory. Default epochs=2 will not train long enough. EMA has a known bug.


tcapelle:
STUDENT haku:

## Results

### AirfRANS: grad-clip=0.5 (tighter constraint)

| Metric | grad-clip=0.5 | grad-clip=1.0 (baseline) | Delta |
|--------|---------------|--------------------------|-------|
| **val_primary/surface_mse** | **0.01446** (epoch 41) | **0.01419** (epoch 41) | +1.9% (worse) |
| test_primary/surface_mse | 0.02330 | 0.01513 | +54.0% (worse) |
| full_val/surface_mse_p | 0.0576 | 0.0564 | +2.1% |
| full_val/volume_mse | 0.0707 | 0.0723 | -2.2% |
| Batches clipped/epoch | 355-360/360 (98.6-100%) | 327-353/360 (91-98%) | More clipping |
| Mean grad norm | 10-24 | 10-22 | Similar |

**W&B run:** [ktpqrj66](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/ktpqrj66)
**Epochs:** 41 (30-min timeout)
**Peak memory:** ~1 GPU (standard AirfRANS footprint)

**Command:**
```bash
  --dataset airfrans --airfrans-task full --optimizer adamw \
  --lr 7e-4 --cosine-t-max 10 --grad-clip 0.5 \
  --no-use-ema --enable-fourier --epochs 999 \
  --agent haku --wandb-group airfrans-gradclip-sweep \
  --wandb-name "haku/airfrans-lr7e4-gradclip05"
```

### What happened

**The tighter clip did not improve over baseline.** Val surface_mse of 0.01446 is essentially tied with the baseline 0.01419 (+1.9%), but test surface_mse is substantially worse at 0.02330 vs 0.01513 (+54%).

Key observations:
1. **Near-total clipping:** 98.6-100% of batches clipped at clip=0.5 (vs 91-98% at clip=1.0). This means almost no gradient updates pass through unmodified.
2. **Similar oscillation pattern:** The cosine LR oscillation pattern persists — val swings between troughs (~0.014-0.026) and peaks (~0.04-0.10). The trough depths are comparable to clip=1.0 but the peaks are slightly lower (max ~0.098 vs ~0.27), confirming tighter clipping reduces spike magnitude.
3. **Val-test divergence:** The large gap between val (0.01446) and test (0.02330) suggests clip=0.5 may be over-constraining the optimizer, limiting its ability to generalize. With clip=1.0, the val-test gap was much smaller (0.01419 vs 0.01513).
4. **Volume metrics improved slightly:** full_val/volume_mse 0.0707 vs 0.0723 (-2.2%), suggesting tighter clipping may benefit bulk-field prediction even if surface metrics don't improve.

### Analysis

The hypothesis that tighter clipping would push trough depths lower was **not confirmed**. The mechanism works as expected (more batches clipped, lower peak spikes), but the over-aggressive constraint appears to prevent the optimizer from taking the large gradient steps needed to escape local minima and find deeper basins. The "sweet spot" appears to be closer to clip=1.0 where ~5-9% of batches pass through unclipped, allowing occasional large exploration steps while still preventing destructive spikes.

### Suggested follow-ups
- **clip=2.0 or 3.0** — test whether loosening the clip allows even deeper basins by permitting more exploration while still preventing the worst spikes
- **Adaptive gradient clipping** — scale the clip threshold based on parameter norm (AGC), which could naturally modulate between tight and loose clipping
- **Focus on the test gap** — the 0.01419 val / 0.01513 test gap at clip=1.0 is already small; techniques to close the val-test gap at clip=0.5 could be more fruitful than further tightening

tcapelle:
ADVISOR: Closing PR #2707 because ADVISOR: Closing — grad-clip=0.5 achieves 0.01446 (+1.9% worse than 0.01419 baseline). Tighter clipping is counterproductive: 98.6-100% of batches clipped (vs 91-98% at clip=1.0), and the val-test gap doubles (0.009 vs 0.001). The model fights the constraint rather than benefiting. The gradient statistics suggest clip=1.0 may already be slightly tight — the optimal is likely in the 1.0-2.0 range, not below 1.0. haku, please pick up your next assignment..

PR #2735, opened 2026-04-21
DrivAerML: 4L/384d + 800 batches/epoch + 180min timeout

tcapelle:
## Hypothesis
The current DrivAerML best uses 4L/384d architecture with max-train-batches=394. Increasing to 800 batches/epoch allows the model to see more training data per epoch, potentially improving convergence on the complex 3D surface geometry. The 50k surface point sampling is already maximizing data density per sample — this change increases the number of samples seen per epoch. We extend the timeout to 180 minutes to give the larger batch budget room to run.

## ⚠️ IMPORTANT: Set timeout before running

```bash
export SENPAI_TIMEOUT_MINUTES=180
```

The default 30-minute timeout is insufficient for DrivAerML. Set this environment variable before launching training.

## Instructions
Run the following command exactly (after setting `SENPAI_TIMEOUT_MINUTES=180`):

```bash
export SENPAI_TIMEOUT_MINUTES=180
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 384 \
  --model-heads 6 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 800 \
  --max-eval-batches 200 \
  --wandb-group drivaerml-4L384d-batches \
  --wandb-name "nezuko/drivaerml-4L384d-800b-180min"
```

Report the best `val_primary/surface_rel_l2_pct` achieved and the W&B run ID.

## Baseline
Current best metrics (PR #2602):
- val_primary/surface_rel_l2_pct: **5.73%**
- W&B run: `7ogfs7ph`
- Epochs: 151
- Reproduce: `cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 384 --model-heads 6 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`


tcapelle:
ADVISOR: Closing PR #2735 because ADVISOR: Closing — 4L/384d+800b superseded by 4L/320d baseline (5.027%). Reassigning to current directions..

PR #2709, opened 2026-04-21
AirfRANS: lr=7e-4 + grad-clip=1.0 + WD=1e-2 — combine grad-clip + regularization

tcapelle:
## Hypothesis

A previous WD=1e-2 experiment failed badly (0.027) because it lacked gradient clipping — the instability from high LR at cosine peaks overwhelmed any regularization benefit. Now that grad-clip=1.0 has been proven to stabilize training at lr=7e-4 (PR #2680, 0.01419), WD=1e-2 can be layered on top from a stable foundation. If WD=1e-2 adds beneficial regularization (reducing overfitting to training distribution), the combination should be strictly better than grad-clip=1.0 alone. The previous WD failure was a training stability problem, not a capacity problem — grad-clip fixes the stability, so WD should now contribute positively.

## Instructions

Run the same configuration as PR #2680 (lr=7e-4, T_max=10, grad-clip=1.0, Fourier, no-EMA), but add `--weight-decay 1e-2`. The AdamW optimizer already supports weight decay — just add the flag.

Run with `--epochs 999` and let the 30-min timeout govern duration. Report:
- Best val `surface_mse` and at which epoch it occurred
- Whether training is stable (no divergence or extreme spikes)
- W&B run ID and link

## Baseline

- **AirfRANS current best:** 0.01419 val / 0.01513 test (PR #2680, lr=7e-4+grad-clip=1.0, T_max=10, 41 epochs, 30-min)
- **Previous WD=1e-2 result (no grad-clip):** 0.027 — failed due to instability
- **TandemFoil current best:** 52.81 (PR #2595, 5L/256d, Lion lr=3e-4, T_max=10)
- **DrivAerML current best:** 5.73% (PR #2602, 4L/384d/6H, T_max=30, 151 epochs)
- **Baseline W&B run:** 48ldl625 (PR #2680, grad-clip=1.0, no WD)

## Reproduce command

```bash
cd target/icml2026 && python train.py \
  --dataset airfrans --airfrans-task full --optimizer adamw \
  --lr 7e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier --epochs 999 \
  --wandb-group airfrans-gradclip-compounds \
  --wandb-name "fern/airfrans-lr7e4-gradclip-wd1e2"
```

> **Note:** `--epochs 999` and `--no-use-ema` are both mandatory. Default epochs=2 will not train long enough. EMA has a known bug.

PR #2737, opened 2026-04-21
AirfRANS: lr=7e-4 + grad-clip=1.5 — fill clip threshold gap

tcapelle:
## Hypothesis

Grad-clip=1.0 at lr=7e-4 achieved the best AirfRANS result (0.01419, PR #2680). Grad-clip=0.5 was just tested and FAILED (0.01446, +1.9% worse) — tighter clipping was counterproductive because 98.6-100% of batches were clipped, creating overfitting.

Grad-clip=1.5 fills the gap between the winning 1.0 and the looser 2.0 (being tested by winry). Since clip=0.5 was too tight, the optimal threshold may be ABOVE 1.0. At clip=1.5, fewer batches will be clipped than at 1.0 (where 91-98% were clipped), allowing more gradient signal through while still preventing the worst spikes.

## Instructions

Run the exact winning AirfRANS config with grad-clip=1.5:

```bash
cd target/icml2026 && python train.py \
  --dataset airfrans --airfrans-task full \
  --optimizer adamw --lr 7e-4 --cosine-t-max 10 \
  --grad-clip 1.5 \
  --no-use-ema --enable-fourier --epochs 999 \
  --wandb-group airfrans-gradclip-sweep \
  --wandb-name "haku/airfrans-lr7e4-gradclip15"
```

Report:
1. Best val_primary/surface_mse and epoch
2. Grad-clip statistics: what fraction of batches are clipped?
3. Spike analysis: compare peak val_primary values at cosine LR peaks vs baseline

## Baseline

- **Primary metric:** val_primary/surface_mse = **0.01419** (PR #2680, lr=7e-4+grad-clip=1.0)
- **Grad-clip sweep so far:** clip=0.3 (roy #2717), clip=0.5 (FAILED, 0.01446), clip=1.0 (WINNER, 0.01419), clip=1.5 (THIS), clip=2.0 (winry #2718)
- **W&B baseline run:** 48ldl625
- **Reproduce baseline:** `cd target/icml2026 && python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 7e-4 --cosine-t-max 10 --grad-clip 1.0 --no-use-ema --enable-fourier --epochs 999`

PR #2739, opened 2026-04-21
AirfRANS: lr=7e-4+grad-clip=1.0+WD=5e-3 (lighter WD curve mapping)

tcapelle:
## Hypothesis

The new best AirfRANS result (0.01323, PR #2709) used WD=1e-2 with grad-clip=1.0 at lr=7e-4. WD=1e-2 is fairly aggressive regularization — we hypothesize that WD=5e-3 (half as strong) may allow the optimizer to explore deeper basins while still benefiting from regularization. The WD=1e-2 result was still improving at epoch 41 (30-min cutoff), suggesting the model hadn't fully converged. With lighter WD the gradient signal is less dampened, potentially enabling access to better minima.

This experiment maps the WD curve at the winning grad-clip=1.0+lr=7e-4 config to find the optimal regularization level.

## Instructions

```bash
  --dataset airfrans \
  --airfrans-task full \
  --optimizer adamw \
  --lr 7e-4 \
  --cosine-t-max 10 \
  --grad-clip 1.0 \
  --weight-decay 5e-3 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb-group airfrans-gradclip-wd \
  --wandb-name "fern/airfrans-lr7e4-gc1-wd5e3"
```

**Key changes vs baseline (PR #2709):**
- `--weight-decay 5e-3` (was `1e-2` — halved WD to reduce regularization pressure)
- All other hyperparameters identical to the new best

**MANDATORY flags:**
- `--epochs 999` — let it run to time budget
- `--no-use-ema` — EMA hurts on AirfRANS

## Baseline

- **Current best:** 0.01323 `val_primary/surface_mse` (PR #2709, epoch 41 — STILL IMPROVING at cutoff)
- **Best config:** AdamW lr=7e-4, cosine T_max=10, grad-clip=1.0, WD=1e-2, no-EMA, Fourier
- **W&B run:** 7vic8kxn
- **Test metric:** 0.01478 `test_primary/surface_mse`
- **External target:** 0.0043 (3.1x gap remaining)
- **Reproduce baseline:** `cd target/icml2026 && python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 7e-4 --cosine_t_max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --epochs 999`

## Expected outcome

Beat 0.01323 by finding a more permissive regularization level that lets the optimizer reach a deeper basin. The phase transition pattern (seen around epoch 14 in baseline) should still occur; we expect WD=5e-3 to produce a similar or slightly different phase-transition timing with potentially deeper final trough.

## Reporting

Please report:
1. `val_primary/surface_mse` at best epoch and final epoch
2. `test_primary/surface_mse`
3. W&B run ID and link
4. Epoch of phase transition (if visible)
5. Whether the run was still improving at cutoff


tcapelle:
STUDENT fern:

## Results

### Key Metrics

| Metric | This Run (WD=5e-3) | Baseline (WD=1e-2, PR #2709) | Delta |
|--------|-------------------|-------------------------------|-------|
| **val_primary/surface_mse** | 0.01926 (E39) | **0.01323** (E41) | +45.6% worse |
| **test_primary/surface_mse** | 0.03202 (E41) | **0.01478** (E41) | +116.6% worse |
| Epochs completed | 41 | 41 | Same |

**W&B run:** `ar3czy3g`
**Command:**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 7e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 5e-3 --no-use-ema --enable-fourier --epochs 999 --agent fern --wandb-name "fern/airfrans-lr7e4-gc1-wd5e3" --wandb-group airfrans-gradclip-wd
```
**Peak memory:** N/A (process complete before GPU query)

### What happened

**Negative result — WD=5e-3 is worse than WD=1e-2.** The hypothesis that lighter WD would enable deeper basin access did not hold. Instead, halving the WD introduced several problems:

1. **Delayed phase transition:** The characteristic AirfRANS phase transition started at epoch 17 vs ~14 for WD=1e-2, suggesting the stronger regularization actually *aids* the transition by constraining the loss landscape.

2. **Increased oscillation volatility:** Post-transition val_primary oscillations were much wider (0.019→0.100 within 2 epochs vs smoother curves with WD=1e-2). The lighter WD gives the optimizer more freedom, but that freedom manifests as instability rather than deeper exploration.

3. **Worse convergence at equal epochs:** At epoch 41 (both runs' cutoff), WD=5e-3 best was 0.01926 vs WD=1e-2's 0.01323. The gap is substantial — WD=1e-2 reached a better minimum with the same compute budget.

4. **Test metric much worse:** 0.03202 vs 0.01478. Since no EMA and no best-model checkpointing, test eval happens at the final epoch (41) rather than the best val epoch (39). The final epoch val was 0.0305 (well above the E39 best of 0.0193), explaining the large test gap.

### Convergence curve highlights
- E1-16: Pre-transition, similar trajectory to WD=1e-2
- E17: Phase transition onset (3 epochs later than WD=1e-2)
- E21-28: Rapid improvement through transition (0.088→0.047)
- E30: Cosine restart, followed by wild oscillations (0.026→0.100→0.019)
- E37-39: Best val region (0.0195, 0.0193)
- E40-41: Cosine restart oscillation (0.039→0.031)

### Analysis

WD=1e-2 is not "too aggressive" for this configuration — it's actually near-optimal. The regularization pressure from WD=1e-2 does three beneficial things at lr=7e-4 + grad-clip=1.0:
1. Smooths the loss landscape, enabling earlier phase transitions
2. Dampens post-transition oscillations, allowing steady descent
3. Acts synergistically with grad-clip: grad-clip handles gradient explosions at cosine peaks, while WD prevents weight magnitudes from growing large enough to amplify those peaks

Reducing WD to 5e-3 breaks this balance — the optimizer has more freedom but less guidance, resulting in more volatile trajectories and shallower final basins.

### Suggested follow-ups
- **WD=2e-2 (double the winner):** If WD=1e-2 is this good, test whether even stronger regularization helps. The grad-clip safety net may allow higher WD.
- **Best-model checkpointing:** The test metric gap (0.032 vs 0.019 best val) highlights that evaluating at the last epoch is suboptimal. Checkpointing at best val would give a more accurate test metric.
- **Longer training with WD=1e-2:** The winning config was still improving at E41. Adding more compute (longer timeout or fewer restarts via larger T_max) could push 0.01323 lower.

tcapelle:
ADVISOR: Closing PR #2739 because ADVISOR: Closing — val=0.01926 is 2.06x worse than the new 0.00935 baseline (gc=1.5, PR #2737). WD=5e-3 at gc=1.0 is a dead end — both the WD value and the clip threshold are suboptimal. Future WD experiments should use gc=1.5. fern, please pick up your next assignment..

PR #2648, opened 2026-04-21
DrivAerML: 4L/320d+T_max=30 — intermediate width expansion

tcapelle:
## Hypothesis
Widening the hidden dimension from 256 to 320 may improve model capacity and surface pressure prediction accuracy on DrivAerML without the instability risk of larger jumps (e.g., 384d). The 4L/256d baseline is still converging at the epoch cap (45 epochs), suggesting underfitting — more representational capacity could help. 320d fills the gap between confirmed configurations (256d and 384d/5L-untested) and gives a cleaner signal on whether width alone is the bottleneck.

Prior experiments have confirmed:
- 4L/256d: 12.70% (current best, still converging)
- 5L/256d: worse/unstable (extra depth causes instability)
- 3L/256d: worse

This suggests width is more promising than depth. The 256→320 step (~56% more parameters in linear layers) is a moderate increase that should fit within the 96GB VRAM budget.

## Instructions

Run the following command:

```bash
cd target/icml2026 && python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine_t_max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 320 \
  --model-heads 4 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group drivaerml-4L-width-sweep
```

Key changes from baseline:
- `--model-hidden-dim 320` (was 256)
- `--wandb_group drivaerml-4L-width-sweep` for grouping with other width experiments

All OOM-prevention flags are preserved: `--batch-size 1`, `--drivaerml-train-surface-points 50000`, `--drivaerml-eval-surface-points 50000`, `--max-train-batches 394`, `--max-eval-batches 200`, `--model-heads 4`.

Report val/loss and all surface MAE metrics (p_in, p_oodc, p_tan, p_re) and the primary metric `val_primary/surface_rel_l2_pct`. Compare against baseline of **12.70%**.

## Baseline

Current best: PR #2593 (shinji), W&B run: `3aaevlho`
- val_primary/surface_rel_l2_pct: **12.70%**
- Model: 4L/256d, T_max=30, AdamW lr=5e-4, no-EMA, Fourier features, 45 epochs (still converging)

Reproduce baseline:
```bash
[...]

PR #2752, opened 2026-04-21
DrivAerML: 4L/320d + WD=1e-2 (transfer AirfRANS regularization)

tcapelle:
## Hypothesis

Transfer the AirfRANS regularization breakthrough to DrivAerML. On AirfRANS, WD=1e-2 + gc=1.0 was a massive win (6.8% improvement on its own), and WD=1e-2 + 4L/256d + gc=1.0 produced a 22.3% improvement. DrivAerML is at the same winning 4L/320d architecture. Testing WD=1e-2 alone (vs default WD=1e-4) at the current best DrivAerML config to isolate the regularization effect.

On AirfRANS, WD=1e-2 works because it: (1) prevents gradient explosions from accumulating at cosine LR peaks, (2) provides clean L2 regularization that complements grad-clip, (3) reduces overfitting on the relatively small 3D CFD dataset. DrivAerML has similar characteristics: small dataset, large model, rapid cosine cycling.

## Instructions

Run exactly this command (no other changes):

```bash
cd target/icml2026
  --cosine-t-max 30 --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 320 --model-heads 5 \
  --epochs 999 --batch-size 1 \
  --weight-decay 1e-2 \
  --wandb_group drivaerml-4L320d-reg
```

**Only change from baseline:** `--weight-decay 1e-2` (default is 1e-4).

Report in a PR comment:
- Best `val_primary/surface_rel_l2_pct` and at which epoch
- `test_primary/surface_rel_l2_pct` at that epoch
- Total epochs completed in 180 min
- W&B run ID and link
- Grad norm behavior (stabilizing or spiking at cosine peaks?)
- Whether model was still improving at cutoff

## Baseline

- **Current best:** 5.027% val / 6.244% test at epoch 257
- **Best PR:** #2648 (zoro — 4L/320d/5H, T_max=30, lr=5e-4, no-EMA, Fourier, 257 epochs in 180-min)
- **External target:** <3.71%
- **Baseline W&B run:** qx7z7if3
- **Reproduce baseline:**
  ```bash
  cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 \
    --cosine-t-max 30 --no-use-ema --enable-fourier \
    --model-layers 4 --model-hidden-dim 320 --model-heads 5 \
    --epochs 999 --batch-size 1 \
  ```


tcapelle:
ADVISOR: Closing PR #2752 because ADVISOR: Closing — WD=1e-2 catastrophically diverges on DrivAerML (confirmed by ray #2790: grad norms hit 231x, 14.40% best vs 4.619% baseline). Also, baseline has moved to 4L/512d making 4L/320d experiments superseded..

PR #2727, opened 2026-04-21
AirfRANS: 4L/256d + lr=7e-4 + grad-clip=1.0 — deeper arch with stable training

tcapelle:
## Hypothesis
The baseline AirfRANS model uses 3L/192d. PR #2595 showed that deeper architecture (5L/256d) helped TandemFoil significantly. Meanwhile, PR #2680 showed that grad-clip=1.0 enables stable training at lr=7e-4 on AirfRANS. We hypothesize that combining a deeper 4L/256d architecture with the stable high-LR + grad-clip training setup will improve AirfRANS surface MAE beyond the current best of 0.01419. Larger model capacity should better capture the complex flow field variations in the full AirfRANS dataset.

## Instructions
Run the following command exactly:

```bash
cd target/icml2026 && python train.py \
  --dataset airfrans \
  --airfrans-task full \
  --optimizer adamw \
  --lr 7e-4 \
  --cosine-t-max 10 \
  --grad-clip 1.0 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --wandb-group airfrans-arch-gradclip \
  --wandb-name "shoya/airfrans-4L256d-gradclip"
```

Key changes from baseline:
- `--model-layers 4 --model-hidden-dim 256 --model-heads 4` (up from 3L/192d/3H default)
- `--lr 7e-4` (same as PR #2680 winner)
- `--grad-clip 1.0` (essential for stable training at this LR — from PR #2680)
- `--cosine-t-max 10` (matches baseline)
- `--no-use-ema` (mandatory)
- `--epochs 999` (mandatory)

Report your W&B run ID and final val metrics in a PR comment when done.

## Baseline
- **Primary metric:** val/loss (surface MAE) = **0.01419**
- **Baseline run:** W&B `48ldl625` (PR #2680)
- **Config:** 3L/192d, lr=7e-4, grad-clip=1.0, T_max=10, AdamW, 41 epochs
- **Reproduce:** `cd target/icml2026 && python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 7e-4 --cosine-t-max 10 --grad-clip 1.0 --no-use-ema --enable-fourier --epochs 999`
- **Target:** < 0.01419

PR #2762, opened 2026-04-21
AirfRANS 4L/256d gc=1.5 + T_max=10 — wider cosine cycles

tcapelle:
## Hypothesis

At the 3L/192d scale, gc=1.5 beat gc=1.0 significantly (0.00935 vs 0.01419). The 4L/256d golden config uses gc=1.0 with T_max=5 (5-epoch cosine cycles). We hypothesise that at the 4L/256d scale, slightly more permissive gradient clipping (gc=1.5) combined with longer cosine cycles (T_max=10) will allow the model to explore wider loss basins and find a better minimum. The two changes are tested together as they are complementary: wider cycles benefit from slightly larger gradient steps.

## Instructions

Run the 4L/256d config with gc=1.5 and T_max=10:

```bash
  --dataset airfrans --airfrans_task full --optimizer adamw \
  --lr 7e-4 --cosine-t-max 10 --grad-clip 1.5 --weight-decay 1e-2 \
  --epochs 999 \
  --wandb_group airfrans-4L256d-gc15
```

Key differences from baseline: `--grad-clip 1.5` (was 1.0) and `--cosine-t-max 10` (was 5)

## Baseline

Current best (PR #2727, W&B run: ruurxdqs):
- val_primary/surface_mse: **0.007264** (at epoch 40, hit cap at epoch 50)
- External target: **0.0043**
- Baseline W&B run: [ruurxdqs](https://wandb.ai/wandb/senpai/runs/ruurxdqs)
- Reproduce: `cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 7e-4 --cosine-t-max 5 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999`


tcapelle:
ADVISOR: Closing PR #2762 because ADVISOR: Closing — gc=1.5 is confirmed dead at 4L/256d (5 independent confirmations: haku, kohaku, armin, fern all failed). Deeper architectures amplify gradients making gc=1.5 harmful..

PR #2769, opened 2026-04-21
AirfRANS: 4L/256d T_max=3 — ultra-short cycles for more cosine troughs

tcapelle:
## Hypothesis

Even shorter cosine cycles (T_max=3) at 4L/256d. T_max=5 was optimal over T_max=10 — giving 8 full cycles in 40 epochs vs 4. The hypothesis is that T_max=3 would give ~13 cycles in the same span, providing even more frequent phase-transition opportunities and deeper basin exploration. However, T_max=3 may be too short — each cycle might not have enough epochs to settle. This tests whether the frequency gain outweighs the settling loss.

## Instructions

Start from the golden 4L/256d command and change T_max from 5 to 3:

```bash
  --dataset airfrans --airfrans_task full --optimizer adamw \
  --lr 7e-4 --cosine-t-max 3 --grad-clip 1.0 --weight-decay 1e-2 \
  --epochs 999 \
  --wandb_group airfrans-4L256d-tmax
```

**Key change:** --cosine-t-max 3 (was 5)

Report the loss envelope behavior — does it look more or less stable than T_max=5? Note any signs of under-settling (val loss not decreasing within a cycle before the next peak).

## Baseline

- **Primary metric:** val_primary/surface_mse
- **Current best:** 0.007264 (val) at epoch 40
- **Best PR:** #2727 (shoya — 4L/256d + gc=1.0 + WD=1e-2 + T_max=5, AdamW lr=7e-4, 50 epochs, hit epoch cap at 61 min)
- **External target:** 0.0043 (~1.7x gap remaining)
- **W&B run:** ruurxdqs
- **Reproduce baseline:** cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 7e-4 --cosine-t-max 5 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999

### DIFF
--- a/.experiment
+++ b/.experiment
@@ -1,0 +1,1 @@
+# airfrans-4L256d-tmax3


tcapelle:
STUDENT kaneda:

## Results

### T_max=3 Experiment — Ultra-short cosine cycles

| Metric | T_max=3 (this run) | T_max=5 baseline (#2727) | Delta |
|---|---|---|---|
| **val_primary/surface_mse (best)** | **0.0116** (epoch 33) | **0.007264** (epoch 40) | **+59.7% worse** |
| full_val/volume_mse (at best) | 0.0634 | — | — |
| full_val/surface_mse_p (at best) | 0.0462 | — | — |
| test_primary/surface_mse | 0.0433 (at epoch 37, a peak) | — | — |
| Epochs completed | 37 (30-min timeout) | 50 (epoch cap at 61 min) | -13 epochs |
| Cosine cycles completed | ~12.3 | ~10 | +23% more cycles |

**W&B run:** swualet9
**Command:** `cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 7e-4 --cosine-t-max 3 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999 --wandb-group airfrans-4L256d-tmax --agent kaneda --wandb-name "kaneda/airfrans-4L256d-tmax3"`

### Loss envelope behavior

**Trough progression** (epochs 3, 6, 9, 12, 15, 18, 21, 24, 27, 30, 33, 36):
`0.264 → 0.203 → 0.209 → 0.156 → 0.065 → 0.063 → 0.033 → 0.040 → 0.016 → 0.015 → 0.012 → 0.024`

Key observations:
1. **Under-settling confirmed.** The trough-to-trough progression becomes non-monotonic in the low-loss regime. The best trough at epoch 33 (0.012) is followed by a regression to 0.024 at epoch 36. The model finds deep basins but can't hold them across the next LR restart.

2. **Violent peak rebounds.** The peak-to-trough ratio becomes extreme at lower loss levels:
   - Epoch 33 (trough) → 34 (peak): 0.012 → 0.057 — **4.9x spike**
   - Epoch 27 (trough) → 28 (peak): 0.016 → 0.055 — **3.5x spike**
   - Compare to early training: epoch 15→16 was only 1.0x (stable)

3. **Oscillation amplitude grows as loss decreases.** In the 0.2-0.06 range, the cycles were well-behaved. Below 0.03, the 3-epoch cycles become destructive — LR restarts kick the model out of deep basins before it can consolidate.

4. **Volume MSE less affected.** Volume metrics showed steadier descent (0.585 → 0.057), suggesting the under-settling primarily impacts surface prediction quality.

5. **Grad norms stabilized nicely** (19.9 at epoch 3 → 9.1 at epoch 37), similar to the T_max=5 baseline pattern. The instability is optimizer-trajectory-driven, not gradient-explosion-driven.

### What happened

**The hypothesis is falsified.** T_max=3 is too short — the frequency gain from more cycles is outweighed by the settling loss. While T_max=3 does produce more phase-transition opportunities (and the model did undergo a clear transition around epoch 13-15), the ultra-short cycles prevent the model from consolidating gains in deep basins. The best trough (0.0116) is 60% worse than the T_max=5 baseline (0.007264).

The critical insight is that settling time matters more than cycle count once the model reaches the low-loss regime. T_max=5 appears to be near-optimal: short enough for frequent exploration, long enough for the model to settle into discovered basins.

**Caveat:** This run got only 37 epochs (30-min timeout) vs the baseline's 50 epochs (epoch cap). With more training time, T_max=3 might eventually approach the baseline — the trough envelope was still descending overall. But the noisy, non-monotonic descent strongly suggests T_max=5 would remain superior at any epoch count.

### Suggested follow-ups

1. **T_max=4** — a less aggressive step down from T_max=5 that might preserve more settling time while still increasing cycle count. This would test whether T_max=5 is exactly optimal or just "good enough."
2. **T_max=3 with lower LR** — the destructive peaks might be mitigated with a lower learning rate (e.g., 5e-4 instead of 7e-4), reducing the amplitude of LR restarts.
3. **Adaptive T_max** — start with T_max=3 for fast early exploration, then switch to T_max=5 or T_max=7 once loss drops below a threshold (e.g., 0.05) to allow better settling in the low-loss regime.

tcapelle:
ADVISOR: Closing PR #2769 because ADVISOR: Closing — T_max=3 achieved 0.011601 (+60%), trough envelope diverging by epoch 36. T_max=5 is optimal; T_max=3 is too short for basin consolidation..

PR #2782, opened 2026-04-21
DrivAerML: isolate gradient clipping gc=1.5 at 4L/320d

tcapelle:
## Hypothesis
Gradient clipping is a key training stability lever, particularly for 3D automotive geometries where surface pressure can vary sharply around mirrors, windshields, and underbody features. The AirfRANS golden config used gc=1.0, but DrivAerML's larger model (4L/320d) and more complex geometry may benefit from a slightly more permissive clip of gc=1.5 — enough to prevent gradient explosions while allowing larger updates that capture sharp pressure gradients. This experiment isolates gradient clipping without weight decay to measure its effect cleanly.

## Instructions
Run the DrivAerML baseline with grad-clip=1.5 (no weight decay, to isolate the effect):

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 320 \
  --model-heads 5 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --grad-clip 1.5 \
  --wandb_group drivaerml-4L320d-gc \
  --epochs 9999
```

Changes vs baseline:
- `--grad-clip 1.5` (added, was unset/default)

## Baseline
Current best DrivAerML metrics (PR #2648, W&B: qx7z7if3):
- val_primary/surface_rel_l2_pct: **5.027%** (lower is better)

Reproduce baseline:
```bash
cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 320 --model-heads 5 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200
```


tcapelle:
ADVISOR: Closing PR #2782 because ADVISOR: Closing — gc=1.5 diverged on DrivAerML and 4L/320d is undersized vs the 4L/512d baseline. All DrivAerML gc experiments failed. You'll be reassigned to a fresh direction..

PR #2767, opened 2026-04-21
AirfRANS: 4L/256d gc=1.5 + WD=1e-2 + T_max=10 — longer cycles reduce volatility

tcapelle:
## Hypothesis

gc=1.5 + WD=1e-2 + T_max=10 at 4L/256d. Longer cosine cycles (T_max=10) reduce the clipping frequency at LR peaks compared to T_max=5, which may make gc=1.5 more stable. At 4L/256d with WD regularization, the architecture may handle gc=1.5 better than at 3L/192d. This tests whether T_max=10 unlocks gc=1.5 at the new architecture — the original gc=1.5 best (0.00935) used T_max=10, so this is a natural combination.

## Instructions

Start from the golden 4L/256d command and change grad-clip to 1.5 and T_max to 10:

```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset airfrans --airfrans_task full --optimizer adamw \
  --lr 7e-4 --cosine-t-max 10 --grad-clip 1.5 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 256 --model-heads 4 \
  --epochs 999 \
  --wandb_group airfrans-4L256d-gc
```

**Key changes:** --grad-clip 1.5 (was 1.0) + --cosine-t-max 10 (was 5)

Monitor grad norm trajectory. Report val/surface_mse and grad norm stats (mean, peak) per epoch.

## Baseline

- **Primary metric:** val_primary/surface_mse
- **Current best:** 0.007264 (val) at epoch 40
- **Best PR:** #2727 (shoya — 4L/256d + gc=1.0 + WD=1e-2 + T_max=5, 50 epochs, hit epoch cap at 61 min)
- **External target:** 0.0043 (~1.7x gap remaining)
- **W&B run:** ruurxdqs
- **Reproduce baseline:** cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 7e-4 --cosine-t-max 5 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999

PR #2791, opened 2026-04-21
AirfRANS: 4L/256d + gc=1.5 + no weight decay + T_max=10

tcapelle:
## Hypothesis
The AirfRANS gc=1.5 experiments with weight decay (WD=1e-2) failed to beat baseline. However, weight decay acts as a regularizer that limits gradient norms, potentially confounding the effect of grad-clip. We hypothesize that gc=1.5 may work when WD is removed, allowing the optimizer to take larger steps where the gradient magnitude justifies it. Doubling T_max to 10 gives the cosine schedule more room to explore before annealing.

## Instructions

```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=180 SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset airfrans \
  --airfrans_task full \
  --optimizer adamw \
  --lr 7e-4 \
  --cosine-t-max 10 \
  --grad-clip 1.5 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999
```

Note: NO `--weight-decay` flag (omit it entirely — the default is 0). `SENPAI_TIMEOUT_MINUTES=180` and `SENPAI_MAX_EPOCHS=9999` are CRITICAL — please include them.

## Baseline
Current best metrics from BASELINE.md (PR #2727):
- val_primary/surface_mse: **0.007264**
- W&B run: `ruurxdqs`
- Reproduce command: `cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset airfrans --airfrans_task full --optimizer adamw --lr 7e-4 --cosine-t-max 5 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 --epochs 999`
- External target: 0.0043


tcapelle:
ADVISOR: Closing PR #2791 because ADVISOR: Closing — gc=1.5 is confirmed dead at 4L/256d. This was the cleanest test possible and kohaku #2767 already confirmed failure at the same config..

tcapelle:
STUDENT haku:

## Results

**Hypothesis**: gc=1.5 + no weight decay + T_max=10 at 4L/256d

### Key Metrics
| Metric | This Run | Baseline (PR #2727) | Delta |
|--------|----------|---------------------|-------|
| val_primary/surface_mse | **0.01349** | **0.00726** | +85.7% (WORSE) |
| full_val/surface_mse_p | 0.0538 | N/A | — |
| full_val/volume_mse | 0.0515 | N/A | — |
| Best epoch | 40 | 40 | same |
| Total epochs | 52 | 50 | — |
| Grad norm (best) | 7.03 | 7.1 | similar |
| Grad clip events (best) | 260/360 | ~350/360 | fewer clips |

### Command
```bash
CUDA_VISIBLE_DEVICES=0 SENPAI_TIMEOUT_MINUTES=180 SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset airfrans --airfrans-task full --optimizer adamw --lr 7e-4 \
  --cosine-t-max 10 --grad-clip 1.5 --weight-decay 0 --no-use-ema \
  --enable-fourier --model-layers 4 --model-hidden-dim 256 --model-heads 4 \
  --epochs 999 --agent haku --wandb-name "haku/airfrans-4L256d-gc15-nowd-tmax10"
```

**Note**: Added explicit `--weight-decay 0` since the code default is 1e-4, not 0 as the PR assumed.

### W&B run: `h20o9w5q`
### Peak memory: ~68 GB

### What happened
**gc=1.5 without weight decay does NOT beat the baseline.** Best val_mse of 0.01349 at epoch 40 is 1.86x worse than the baseline's 0.00726.

**Key observations:**
1. **Early gradient instability**: Without WD, grad norms rose rapidly (11→31) in epochs 1-17, while the baseline with WD=1e-2 stabilized at 18→7. gc=1.5 clipped ~95% of steps but couldn't prevent the growing norms.
2. **Late stabilization**: Around epoch 35-40, grad norms finally converged to ~7.0, matching the baseline's final level. But this took ~35 epochs vs ~20 for the baseline.
3. **Wide oscillation**: The cosine LR cycling produced val_mse oscillations of 3-10x between peaks and troughs (e.g., 0.014→0.156 at ep40→47). The baseline with WD showed much narrower oscillation.
4. **Trough descent pattern**: Troughs deepened progressively: 0.185→0.086→0.061→0.033→0.021→0.016→0.013. Each transition took 4-8 epochs. After epoch 40, the model plateaued.
5. **Convergence**: By epoch 40, grad norms (7.03) and clip events (260/360=72%) match the baseline regime, but the model found a shallower basin (0.0135 vs 0.0073).

### Analysis
The hypothesis was that WD confounds gc=1.5's effect by limiting gradient magnitudes. The data shows this is partially true — removing WD does allow larger gradients early (up to 31 vs 18.7), but this doesn't translate to deeper basins. The opposite: the larger early gradients appear to push the model into a less favorable region of the loss landscape. The baseline's WD=1e-2 provides beneficial regularization that helps the optimization find deeper basins, not just gradient control.

### Suggested follow-ups
- **gc=1.5 is confirmed dead at 4L/256d**: Combined with kohaku's PR #2767 results, gc=1.5 consistently underperforms gc=1.0 at this architecture scale. The moderate clipping at gc=1.5 doesn't help — the baseline's tighter gc=1.0 + WD=1e-2 is the better optimization regime.
- **Explore gc=1.0 + no-WD**: If the goal is to isolate WD's effect, compare gc=1.0 with and without WD rather than changing gc simultaneously.
- **Focus on other levers**: Architecture scaling (width, depth) and T_max tuning have been more productive than grad-clip/WD variations.

PR #2691, opened 2026-04-21
DrivAerML: 4L/512d width scaling — can wider beat 5.73%?

tcapelle:
## Hypothesis

Width scaling from 256d→384d delivered a 52% relative improvement (11.97%→5.73%). We hypothesize that continuing on this scaling axis — 384d→512d — will further reduce surface pressure error. The head dimension stays constant at 64 (512/8=64, matching 384/6=64), so we are adding capacity without changing the per-head resolution. This tests whether the model is still in a regime where width ≈ representational capacity is the binding constraint.

## Instructions

Run the following command exactly:

```bash
cd target/icml2026 && python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --agent frieren \
  --wandb-name "frieren/drivaerml-4L512d" \
  --wandb-group "drivaerml-width-scaling"
```

**Note on VRAM:** 512d with 50k surface points may approach VRAM limits on an A100-80GB. If you hit OOM, fall back to `--drivaerml-train-surface-points 40000 --drivaerml-eval-surface-points 40000` and note this in your results comment. Do not reduce batch size or layers.

Report the best `val_primary/surface_rel_l2_pct` you observe and the epoch at which it occurred.

## Baseline

Current best metrics (PR #2602, W&B run `7ogfs7ph`):
- **val_primary/surface_rel_l2_pct: 5.73%** (epoch 144, 151 epochs total)
- Architecture: 4L/384d/6H, Fourier features enabled, no EMA
- Training note: still converging at 180-min cutoff (epoch 141=10.2%, epoch 144=5.73% — late oscillation)
- Reproduce: `cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 384 --model-heads 6 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`
[...]

PR #2793, opened 2026-04-21
DrivAerML: Transfer gc=1.5 to 4L/512d architecture

tcapelle:
## Hypothesis

Transfer gradient clipping (gc=1.5) to the new best 4L/512d architecture. gc=1.5 gave a 26% improvement on AirfRANS over no-clip. The current 4L/512d baseline (PR #2691, 4.619%) used no gradient clipping. Larger models with more parameters may have larger gradient variance — explicit grad-clip could stabilize training and allow more aggressive optimization without divergence.

## Instructions

```bash
  --cosine-t-max 30 --grad-clip 1.5 \
  --epochs 999 --batch-size 1 \
  --wandb_group drivaerml-4L512d-gc
```

The only change vs baseline is adding `--grad-clip 1.5`. Keep all other hyperparameters identical to PR #2691.

Report your best `val/surface_pressure_mae` and the epoch at which it occurred. Also note if training stability differs from baseline.

## Baseline

- **Dataset:** DrivAerML
- **Baseline config:** 4L/512d/8H, lr=5e-4, T_max=30, no-EMA, Fourier, no grad-clip, 267 epochs, 180-min


tcapelle:
STUDENT frieren:

## Results

**Verdict: NEGATIVE — gc=1.5 causes catastrophic mid-training divergence on DrivAerML 4L/512d.**

### Key Metrics

| Metric | gc=1.5 (this run) | Baseline (no clip, PR #2691) |
|--------|-------------------|------------------------------|
| **val_primary/surface_rel_l2_pct** | 11.44% (best, epoch 54) | **4.619%** (epoch 256) |
| Epochs before divergence | ~58 | 267 (stable) |
| Final metric | 64.4% (diverged) | 5.926% (mild overfit) |

- **W&B run:** rcwt82k3
- **Best epoch:** 54 (val=11.44%, LR near trough)
- **Divergence onset:** epoch 59–62 (grad norm 0.85 → 2.46 → 11.16 → 17.5)
- **Catastrophic by epoch 67:** grad norm 47 → 90 → 134 → 925 → 2853
- **Peak memory:** ~1 GPU (RTX PRO 6000 Blackwell, 98GB)

### Command Used

```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_TIMEOUT_MINUTES=180 SENPAI_MAX_EPOCHS=9999 python train.py \
  --cosine-t-max 30 --grad-clip 1.5 \
  --epochs 999 --batch-size 1 \
  --wandb-group drivaerml-4L512d-gc \
  --agent frieren --wandb-name "frieren/drivaerml-4L512d-gc15"
```

### What Happened

Training converged normally through epoch 54 (trough envelope: ep8=32.8% → ep22=19.2% → ep38=13.5% → ep54=11.4%). Gradient clipping events decreased from 333/394 (epoch 1) to 18/394 (epoch 54) as the model stabilized — clipping was becoming irrelevant.

Then at epoch 59–62, a sudden instability cascade began. Grad norm jumped from 0.85 to 11.16 in 3 epochs, and clipping events surged back to 308/394. Despite gc=1.5 clipping all batches, the gradient norm continued to explode exponentially: 17→47→90→134→925→2853. The model never recovered.

**Why gc=1.5 fails here but works on AirfRANS:** AirfRANS has naturally larger gradient variance (grad norms 10–22 throughout training), so gc=1.5 provides active stabilization from start to finish. DrivAerML 4L/512d has small natural grad norms (dropping to 0.76 by epoch 50), so gc=1.5 is essentially inactive during normal training. When an instability event occurs, the clip threshold is too far above the natural gradient scale to catch it early — by the time clipping activates, the divergence cascade is already underway.

Paradoxically, the baseline (no clip) survives 267 epochs because the model's own gradient dynamics are self-stabilizing at this scale. Adding a loose clip at 1.5 doesn't help during stable phases and can't prevent divergence once it starts.

### Suggested Follow-ups

1. **Tighter clipping for DrivAerML:** gc=0.5 or gc=1.0 would match the natural grad norm scale (~0.8–1.0 in the stable phase) and might catch instabilities earlier. gc=1.0 was effective on AirfRANS with 4L/256d.
2. **gc + weight decay compound:** WD=1e-2 regularization might prevent the instability onset entirely, making clipping unnecessary. The AirfRANS results showed gc+WD was strictly better than either alone.
3. **Adaptive clipping:** Scale the clip threshold relative to the running average grad norm rather than using a fixed value, so it tightens as the model stabilizes.

tcapelle:
ADVISOR: Closing PR #2793 because ADVISOR: gc=1.5 definitively falsified on DrivAerML — 11.44% vs 4.619% baseline (2.5x worse), run crashed. Your analysis is spot-on: DrivAerML's natural grad norms (~0.76-0.85) are too low for gc=1.5 to activate during stable training. AirfRANS grad norms (10-22) are 10-25x higher, explaining why gc=1.5 works there. Great forensic work. New assignment incoming — DrivAerML WD=1e-2 + gc=0.5 compound at 4L/512d, which should match the grad norm scale and provide continuous regularization..

PR #2794, opened 2026-04-21
DrivAerML: Add WD=1e-2 regularization to 4L/512d

tcapelle:
## Hypothesis

Add weight decay (WD=1e-2) regularization to the 4L/512d architecture. The current best 4L/512d baseline (PR #2691) shows signs of overfitting: best val=4.619% at epoch 267 but final-epoch val degrades to 5.926%. WD=1e-2 has been effective on AirfRANS as part of the "golden config." With a 512-dim hidden layer and more parameters, explicit L2 regularization may suppress this overfit and allow the model to reach a better validation floor before diverging.

## Instructions

```bash
  --cosine-t-max 30 --weight-decay 1e-2 \
  --epochs 999 --batch-size 1 \
  --wandb_group drivaerml-4L512d-reg
```

The only change vs baseline is adding `--weight-decay 1e-2`. Keep all other hyperparameters identical to PR #2691.

Report: (1) best `val/surface_pressure_mae` and epoch, (2) final-epoch val metric, (3) whether the best-vs-final gap is smaller than baseline's 4.619%→5.926% gap.

## Baseline

- **Dataset:** DrivAerML
- **Baseline config:** 4L/512d/8H, lr=5e-4, T_max=30, no-EMA, Fourier, no WD, 267 epochs, 180-min
- **Overfitting signal:** best val=4.619% ep267 → final val=5.926% (clear overfit)


tcapelle:
STUDENT fern:

## Results

### DrivAerML: WD=1e-2 regularization on 4L/512d

| Metric | Baseline (PR #2691) | WD=1e-2 (this PR) |
|--------|--------------------|--------------------|
| **Best val/surface_rel_l2_pct** | **4.619%** (ep 256) | 5.33% (ep 225) |
| **Final val** | 5.926% (ep 267) | 39.58% (ep 261) |
| **Best-vs-final gap** | 1.307pp (mild overfit) | **34.25pp (catastrophic divergence)** |
| **Test metric** | N/A | 40.77% (evaluated at diverged state) |

### Key findings

1. **Did NOT beat baseline.** Best val of 5.33% vs baseline 4.619% — a 15.4% relative regression.

2. **Catastrophic late-training divergence.** The model converged smoothly through epoch ~235 (trough envelope: 63.7% → 31.1% → 18.6% → 10.4% → 8.3% → 6.8% → 5.3%), then experienced cascading instability starting at epoch 236. By epoch 240, val spiked to 26.9%. By epoch 246, full divergence at 59.4%. Grad norms exploded from ~0.15 to 437 by epoch 260. The model never recovered.

3. **WD=1e-2 worsened the overfitting problem instead of fixing it.** Baseline showed mild overfitting (4.619→5.926%, 1.3pp gap). WD=1e-2 produced catastrophic instability (5.33→39.58%, 34pp gap). The L2 penalty likely destabilized weight norms at cosine LR peaks, causing a feedback loop between large gradients and the WD penalty term.

4. **Convergence was healthy until the blowup.** The first 225 epochs showed excellent behavior: smooth trough descent, tightening oscillation amplitude, decreasing grad norms (from 5.5 to 0.14). The instability appears to be a late-training phenomenon where accumulated weight decay + cosine peak LR creates a destructive resonance.

5. **Periodic loss spikes preceded the collapse.** Throughout training, occasional loss spikes (0.02-0.03 vs normal 0.006) occurred every ~20-30 epochs. These recovered initially but grew larger over time, eventually becoming unrecoverable at epoch 236+.

### What happened

WD=1e-2 without grad clipping is destructively unstable on 4L/512d DrivAerML at long training horizons. The weight decay penalty interacts badly with cosine annealing peaks: at peak LR, large gradient steps combined with L2 penalties on already-regularized weights create oscillations that compound over cycles. The baseline (no WD) overfits mildly because there's no regularization, but WD=1e-2 overcorrects and destabilizes training entirely.

### Suggested follow-ups

- **WD=1e-2 + grad-clip=1.0 or 1.5**: Grad clipping would prevent the catastrophic gradient explosions at cosine peaks. This is the most promising combination — AirfRANS already showed that WD+grad-clip is strictly better than either alone (PR #2709).
- **Lower WD (1e-3 or 1e-4)**: A weaker penalty might provide regularization without destabilizing late training.
- **WD=1e-2 with cosine trough checkpointing**: Save the best checkpoint and stop — the model is excellent through epoch 225 but should be stopped before it diverges.

### Command used
```bash
CUDA_VISIBLE_DEVICES=0 SENPAI_TIMEOUT_MINUTES=180 SENPAI_MAX_EPOCHS=9999 python train.py \
  --cosine-t-max 30 --weight-decay 1e-2 \
  --epochs 999 --batch-size 1 \
  --wandb-group drivaerml-4L512d-reg \
  --agent fern --wandb-name "fern/drivaerml-4L512d-wd1e2"
```

- **W&B run:** [de4lj973](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/de4lj973)
- **Epochs:** 261 (180-min timeout)
- **Peak memory:** ~2.4 GB VRAM (4L/512d, batch-size 1, 50k surface points)

tcapelle:
ADVISOR: Closing PR #2794 because ADVISOR: Closing — your WD=1e-2 result of 5.331% crashed and does not beat the 4.619% baseline. Heavy weight decay alone is insufficient for DrivAerML stability. Gradient clipping appears to be the more effective lever. Thank you, fern..

PR #2795, opened 2026-04-21
DrivAerML: Faster cosine cycling T_max=20 at 4L/512d

tcapelle:
## Hypothesis

Test faster cosine LR cycling (T_max=20 vs baseline T_max=30) at the new best 4L/512d architecture. T_max=20 gives ~50% more LR restarts within the same 180-min budget, which has consistently helped on AirfRANS and TandemFoil by escaping local minima more frequently. The TandemFoil best result uses T_max=10 (ultra-rapid cycling). The DrivAerML landscape may similarly benefit from more frequent restarts.

## Instructions

```bash
  --cosine-t-max 20 \
  --epochs 999 --batch-size 1 \
  --wandb_group drivaerml-4L512d-tmax
```

The only change vs baseline is `--cosine-t-max 20` (was 30). Keep all other hyperparameters identical to PR #2691.

Report your best `val/surface_pressure_mae`, the epoch at which it occurred, and how many complete cosine cycles were completed within the budget.

## Baseline

- **Dataset:** DrivAerML
- **Baseline config:** 4L/512d/8H, lr=5e-4, T_max=30, no-EMA, Fourier, 267 epochs, 180-min
- **Context:** TandemFoil best uses T_max=10; shorter cycles have helped across datasets


tcapelle:
STUDENT rei:

## Results

### DrivAerML: 4L/512d + T_max=20 (faster cosine cycling)

| Metric | Baseline (PR #2691, T_max=30) | T_max=20 | Status |
|--------|------------------------------|----------|--------|
| val_primary/surface_rel_l2_pct | **4.619%** (267ep) | 10.59% (epoch 74) | **FAILED — diverged** |

### Command used
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_TIMEOUT_MINUTES=180 SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 20 \
  --epochs 999 --batch-size 1 \
  --wandb-group drivaerml-4L512d-tmax --agent rei \
  --wandb-name "rei/drivaerml-4L512d-tmax20"
```

### W&B run
- **Run ID:** 4ybf7pcx
- **Link:** https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/4ybf7pcx

### Epochs
- 140+ epochs (run still going but diverged catastrophically at epoch ~133)
- Best val: 10.59% at epoch 74 (pre-instability)

### Timeline of the catastrophe

| Phase | Epochs | Val range | Grad norm | Notes |
|-------|--------|-----------|-----------|-------|
| Healthy convergence | 1-74 | 64%→10.59% | 0.5-2.3 | Smooth descent |
| First instability | 75-81 | 10.59%→36.64% | 1.1 | Val tripled, loss 3x |
| Partial recovery | 82-118 | 36%→12.16% | 0.7-0.9 | Slow recovery over ~37 epochs |
| Second instability | 119-126 | 12%→18.96% | 1.9 | Smaller spike |
| Temporary recovery | 127-130 | ~15% | ~1.0 | Brief stabilization |
| **Catastrophic divergence** | 131-133 | 15%→62.39% | **207.89** | Grad norm explosion, training collapse |
| Irrecoverable | 134+ | ~57% | ~5.0 | Model destroyed |

### What happened

**T_max=20 is catastrophically unstable for 4L/512d.** The run showed three phases of instability, culminating in complete training collapse at epoch 133.

1. **Pre-instability performance was promising.** The model reached 10.59% by epoch 74 with a healthy convergence trajectory, smooth grad norms (0.5-2.3), and steadily declining loss. At this rate, it might have reached ~5% by epoch 200 without the instability.

2. **Repeated instabilities suggest a structural problem.** Three separate instability events (epochs 80, 125, 133), each worse than the last, indicate T_max=20 causes the model to visit dangerous regions of the loss landscape during high-LR phases. The LR restarts are too aggressive for the 512d model's high-dimensional parameter space.

3. **The 4L/512d model is more sensitive to LR cycling.** The T_max=30 baseline (4.619% over 267 epochs) was stable throughout. T_max=50 worked well for 4L/384d. The pattern: larger models need slower cosine cycling. 512d parameters need gentler LR treatment.

4. **Grad clipping would have prevented this.** The catastrophic divergence (grad norm=207.89) could have been caught by gradient clipping. The baseline doesn't use grad clipping, and neither did this run. Adding `--grad-clip 1.0` or `--grad-clip 5.0` with T_max=20 might stabilize training.

5. **No checkpoint saving means the 10.59% result is lost.** If we had best-checkpoint saving, we could at least report the epoch 74 result with a proper test metric.

### Conclusion

**T_max=20 definitively DOES NOT work for 4L/512d without grad clipping.** The hypothesis that faster cycling helps DrivAerML is falsified at this model scale. The cosine LR restarts cause gradient explosions in the larger model.

### Suggested follow-ups

1. **T_max=20 + grad-clip 1.0** — gradient clipping may prevent the divergence while preserving the benefits of faster cycling.
2. **T_max=40 or T_max=50** — slower cycling as tested on 4L/384d (where T_max=50 beat T_max=30) may be the right direction for larger models.
3. **Close as negative result** — T_max=20 is too aggressive for 4L/512d without stabilization.

tcapelle:
ADVISOR: Closing PR #2795 because ADVISOR: Closing — T_max=20 catastrophically diverged at 4L/512d. Three instability events culminating in grad norm=207.89 at epoch 133. Best was 10.59% at ep74 (2.3x worse than baseline). Confirms larger models need slower cycling — T_max=30 (baseline) is more appropriate. Will test T_max=50 which worked well for 4L/384d. Excellent catastrophe timeline analysis..

PR #2810, opened 2026-04-21
TandemFoil: lr=1.25e-4 + gc=1.0 — deeper LR sweep with gradient clipping

tcapelle:
## Hypothesis
lr=1.5e-4 just beat lr=2e-4 (44.72 vs 45.07, PR #2724). The LR trend is: 3e-4→2e-4→1.5e-4 all improving. We test lr=1.25e-4 (continuing the sweep) combined with gc=1.0 (which reduced destructive cosine-peak spikes observed at lr=1.5e-4 — spikes to 80-90 at cosine peaks). The combination may stabilize optimization and allow the trough envelope to descend further.

## Instructions
\`\`\`bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=180 SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset tandemfoil \
  --optimizer lion \
  --lr 1.25e-4 \
  --cosine-t-max 10 \
  --grad-clip 1.0 \
  --no-use-ema \
  --model-slices 64 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --enable-cp-panel-tandem-only \
  --asinh-pressure \
  --residual-prediction \
  --enable-pressure-prior-addition \
  --epochs 999 \
  --agent gilbert \
  --wandb-name "gilbert/tandem-lr125e4-gc10" \
  --wandb-group "tandem-lr-gc-sweep"
\`\`\`

Report val_primary/surface_pressure_mae at best epoch, total epochs, time per epoch, and whether oscillation amplitude decreased vs lr=1.5e-4.

## Baseline
Current best: **44.72** at epoch 89 (3L/192d, lr=**1.5e-4**, T_max=10, NO grad-clip) — PR #2724
Trend: lr=3e-4(52.81)→lr=2e-4(45.07)→lr=1.5e-4(44.72) — lower LR keeps winning

PR #2840, opened 2026-04-21
TandemFoil: lr=1e-4 + gc=1.0 multi-seed replication

tcapelle:
## Hypothesis

Verify the lr=1e-4 result with multiple seeds to establish reproducibility before building on it as a new baseline.

## Baseline

val_primary/surface_pressure_mae = **30.10** (PR #2810)

## Runs

Run all from `cd target/icml2026`.

**Run 1** (seed=42, `CUDA_VISIBLE_DEVICES=0,1,2`):
```bash
SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=180 python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999 --seed 42 --wandb-group tandem-lr1e4-seeds --wandb-name tandem-lr1e4-seed42
```

**Run 2** (seed=123, `CUDA_VISIBLE_DEVICES=3,4,5`):
Same as Run 1 but `--seed 123 --wandb-name tandem-lr1e4-seed123`

**Run 3** (seed=456, `CUDA_VISIBLE_DEVICES=6,7`):
Same as Run 1 but `--seed 456 --wandb-name tandem-lr1e4-seed456`

## Key Insight

Before committing to lr=1e-4 as a new stepping stone, we need to confirm it's reproducible across seeds. This also gives us a distribution of outcomes (mean ± std) which is important for paper-quality claims. If results cluster tightly around a value < 30.10, that's strong evidence lr=1e-4 is a genuine improvement.

## Expected Outcome

val_primary/surface_pressure_mae < 30.10 across all seeds, with low variance

### DIFF
--- a/.experiment
+++ b/.experiment
@@ -1,1 +1,1 @@
-# airfrans-3L256d-golden
+tandemfoil-lr1e4-extended-seed


tcapelle:
ADVISOR: Closing PR #2840 because ADVISOR: Closing this PR. TandemFoil best is now val=22.537 (#2924 MERGED) using Lion+EMA+gc=0.5. The lr=1e-4+gc=1.0 multi-seed replication investigated here is in the 30+ range — far below the current best. Obsolete — history will be reassigned a fresh hypothesis..

PR #2947, opened 2026-04-22
tandemfoil_paper: establish first field_mse baseline (Experiment 4)

tcapelle:
## Hypothesis

The tandemfoil_paper benchmark (TandemFoilSet Experiment 4) has never been run in this programme. We need to establish a first `val_primary/field_mse` baseline using the same physics-informed recipe that won on the main tandemfoil benchmark. This is the paper-faithful high-Re calibration target — getting a number here is essential before we can claim competitiveness in the ICML paper.

## Current Baseline

**tandemfoil_paper: NO BASELINE — this is the first ever run.**

Paper reference numbers (Table 6, Experiment 4):

| Task | MGN baseline | Paper best (PRE-RES-FREE + RES-COMB) |
|---|---:|---:|
| cruise_random_uniform | 1.79 ± 1.38 | 0.10 ± 0.13 |
| cruise_random_aoa_extrap | 2.03 ± 1.96 | 0.18 ± 0.24 |
| cruise_random_re_extrap | 4.85 ± 1.82 | 0.36 ± 0.53 |
| cruise_random_stagger_extrap | 1.74 ± 1.66 | 0.13 ± 0.17 |
| cruise_random_gap_extrap | 1.95 ± 1.68 | 0.14 ± 0.20 |
| racecar_uniform | 0.61 ± 0.51 | 0.21 ± 0.29 |

Primary metric: `val_primary/field_mse` (normalized full-field MSE, z-score with training-set stats, lower is better).

**Immediate target:** beat MGN baseline (1.79) on cruise_random_uniform and approach the paper best (0.10).

## Experiment Plan

Transfer the TandemFoil golden config to `--dataset tandemfoil_paper`. The tandemfoil_paper benchmark uses the same 24-feature preprocessing as `tandemfoil/` but NOT the noam-style pressure-denorm contract — instead it computes normalized full-field MSE over all three channels [Ux, Uy, p].

Run 8 GPU sweeps across lr and a light architecture check:

| Run | Key change |
|---|---|
| A | Lion lr=1e-4, 3L/192d/3H, all physics flags |
| B | Lion lr=1.25e-4, 3L/192d/3H, all physics flags |
| C | Lion lr=1.5e-4, 3L/192d/3H, all physics flags |
| D | Lion lr=2e-4, 3L/192d/3H, all physics flags |
| E | Lion lr=1e-4, 3L/256d/4H, all physics flags |
| F | Lion lr=1.25e-4, 3L/256d/4H, all physics flags |
[...]

PR #2902, opened 2026-04-21
Cross-dataset: gradient accumulation — larger effective batch size across all three datasets (CODE CHANGE)

tcapelle:
## Hypothesis

Gradient accumulation simulates a larger effective batch size by accumulating gradients over multiple micro-steps before calling `optimizer.step()`. For AirfRANS and TandemFoil, where batch sizes are already moderate, this allows exploration of a higher effective batch regime that may smooth the loss landscape and improve generalization. For DrivAerML, which is forced to `batch-size 1` by VRAM constraints, gradient accumulation is the only path to larger effective batches, potentially stabilizing training significantly. We hypothesize that `--grad-accum-steps 4` (effective batch size ×4) will lower val/loss across all three datasets, with the largest gain on DrivAerML where the effective batch is currently the smallest.

## Code Change Required

Add the following to `target/icml2026/train.py`:

```python
# Add CLI arg near other training args:
parser.add_argument('--grad-accum-steps', type=int, default=1,
                    help='Number of gradient accumulation steps. Effective batch size = batch_size * grad_accum_steps.')
```

Replace the existing per-batch training loop with an accumulation-aware version:

```python
optimizer.zero_grad()
for micro_step in range(args.grad_accum_steps):
    batch = next(train_iter)
    loss = model(batch) / args.grad_accum_steps
    loss.backward()
optimizer.step()
scheduler.step()
```

Note: `args.grad_accum_steps=1` should reproduce the original behavior exactly (one micro-step, no accumulation). Make sure the train iterator (`train_iter`) is set up to support repeated `next()` calls per optimizer step.

## Instructions

Run the following matrix across all three datasets. Use `--wandb_group grad-accum-cross` on all runs.

**AirfRANS — golden config + grad accumulation variants:**

```bash
# AF control (accum=1, confirm baseline)
python train.py --dataset airfrans --airfrans-task full \
  --optimizer adamw --lr 7e-4 --cosine-t-max 5 \
  --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier \
  --model-layers 2 --model-hidden-dim 256 --model-heads 4 \
  --epochs 999 --grad-accum-steps 1 \
  --wandb_group grad-accum-cross --wandb_run_name af-accum1-control

# AF accum=4 (effective batch ×4)
python train.py --dataset airfrans --airfrans-task full \
  --optimizer adamw --lr 7e-4 --cosine-t-max 5 \
[...]

PR #2898, opened 2026-04-21
Cross-dataset: torch.compile throughput — 20-40% speedup, more epochs per wall-clock

tcapelle:
## Hypothesis

`torch.compile` (PyTorch 2.x) fuses operations and eliminates Python overhead, typically giving 20-40% throughput improvement on transformer models with no change to model quality. The `--compile-model` flag already exists in `train.py` but has not been tested at scale across all three datasets. With our fixed epoch budget, more throughput = more gradient steps per wall-clock hour = better-converged models within the same SENPAI timeout. This is a free win if it works.

## Instructions

The `--compile-model` flag already exists. Just add it to the golden configs for all three datasets and run 6 experiments (2 per dataset: one with compile, one without, to measure throughput delta).

Use `--wandb_group compile-throughput` for all runs.

**DrivAerML — with compile:**
```bash
python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --compile-model --wandb_group compile-throughput
```

**DrivAerML — without compile (control):**
```bash
python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --wandb_group compile-throughput
```

**AirfRANS — with compile:**
```bash
python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 7e-4 --cosine-t-max 5 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --compile-model --wandb_group compile-throughput
```

**AirfRANS — without compile (control):**
```bash
[...]

PR #2887, opened 2026-04-21
AirfRANS: gc=1.0+T_max=10 LR scan — push below 0.001236

tcapelle:
## Hypothesis

AirfRANS best (0.001236) uses lr=7e-4, T_max=5, gc=1.0, wd=1e-2. Inosuke's recent compound experiment (in-flight) uses T_max=10 and may yield a new best. This PR locks T_max=10 and gc=1.0 and scans learning rates (5e-4, 6e-4, 7e-4, 8e-4, 9e-4, 1e-3) to find the optimal LR at T_max=10. The AirfRANS dataset has seen the largest improvements from gc=1.0; T_max=10 provides slower cosine cycling (more time near peak LR) which may benefit from a slightly higher or lower LR than T_max=5. GPUs 6–7 run TandemFoil LR scan variants.

## Instructions

Run all 8 commands in parallel, one per GPU. All commands run from `cd target/icml2026`. Do NOT set SENPAI_TIMEOUT_MINUTES or SENPAI_MAX_EPOCHS — inherit from pod env.

**GPU 0 — AirfRANS: lr=5e-4, gc=1.0, T_max=10, wd=1e-2**
```bash
CUDA_VISIBLE_DEVICES=0 python train.py \
  --dataset airfrans --airfrans-task full --optimizer adamw \
  --lr 5e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 \
  --model-heads 4 --epochs 999 --wandb-group gc-tmax10-lr-scan
```

**GPU 1 — AirfRANS: lr=6e-4, gc=1.0, T_max=10, wd=1e-2**
```bash
CUDA_VISIBLE_DEVICES=1 python train.py \
  --dataset airfrans --airfrans-task full --optimizer adamw \
  --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 \
  --model-heads 4 --epochs 999 --wandb-group gc-tmax10-lr-scan
```

**GPU 2 — AirfRANS: lr=7e-4, gc=1.0, T_max=10, wd=1e-2 (golden LR, new T_max)**
```bash
CUDA_VISIBLE_DEVICES=2 python train.py \
  --dataset airfrans --airfrans-task full --optimizer adamw \
[...]

PR #2957, opened 2026-04-22
Wave3: Global Context Token via Cross-Attention — cross-dataset

tcapelle:
## Hypothesis

CFD simulations have global dependencies: the pressure at one surface point is influenced by the entire flow field. Our current architecture processes nodes locally through attention within windows or local neighborhoods. A single learnable global context token — analogous to the [CLS] token in BERT — can accumulate global information and broadcast it back to all nodes via cross-attention, giving the model access to global flow state without the quadratic cost of full self-attention.

## Global Context Token Implementation

Add `--global_context_token` flag. When enabled:
1. Add a learnable global token `nn.Parameter(torch.zeros(1, 1, dim))`
2. After each transformer block, update the global token via cross-attention with all node features
3. Broadcast the global token back to nodes by concatenating or adding it as a positional bias

```python
class GlobalContextBlock(nn.Module):
    def __init__(self, dim, num_heads=8):
        super().__init__()
        # Learnable global token
        self.global_token = nn.Parameter(torch.zeros(1, 1, dim))
        nn.init.trunc_normal_(self.global_token, std=0.02)
        
        # Cross-attention: global token attends to all nodes
        self.cross_attn = nn.MultiheadAttention(
            embed_dim=dim, 
            num_heads=num_heads, 
            batch_first=True,
            dropout=0.0
        )
        self.norm1 = nn.LayerNorm(dim)
        self.norm2 = nn.LayerNorm(dim)
        
        # Project global token back to per-node bias
        self.proj = nn.Linear(dim, dim)
        
    def forward(self, x):
        # x: [B, N, D]
        B = x.shape[0]
        g = self.global_token.expand(B, -1, -1)  # [B, 1, D]
        
        # Global token attends to all nodes
        g_updated, _ = self.cross_attn(
            query=self.norm1(g),
            key=self.norm2(x),
            value=x
        )
        g = g + g_updated  # residual
        
        # Broadcast global context to all nodes as additive bias
        global_bias = self.proj(g)  # [B, 1, D]
        return x + global_bias  # broadcasts over N dimension
```

Insert one `GlobalContextBlock` after the final transformer block (or after every K blocks, e.g. K=2).

## Training Instructions — All 4 Datasets

Use `--wandb_group kohaku-global-ctx`. One run per dataset.

### TandemFoil
```bash
cd target/icml2026

python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 \
  --enable-pressure-prior-addition --epochs 999 \
  --global_context_token --wandb_group kohaku-global-ctx
```

### TandemFoil Paper (use EMA)
```bash
python train.py --dataset tandemfoil_paper --optimizer lion --lr 1e-4 --cosine-t-max 10 \
  --grad-clip 1.0 --weight-decay 1e-2 --ema-decay 0.999 \
  --enable-pressure-prior-addition --epochs 999 \
  --global_context_token --wandb_group kohaku-global-ctx
```

### AirfRANS
```bash
python train.py --dataset airfrans --airfrans-task full --optimizer adamw \
  --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 \
  --model-heads 4 --epochs 999 \
  --global_context_token --wandb_group kohaku-global-ctx
```

### DrivAerML
```bash
  --model-heads 8 --epochs 999 \
  --global_context_token --wandb_group kohaku-global-ctx
```

| TandemFoil | `val_primary/surface_pressure_mae` | **26.06** | #2887 (gohan) |
| TandemFoil Paper | `val_primary/field_mse` | TBD (establishing via #2947-2949) |
| AirfRANS | `val_primary/surface_mse` | **0.000627** | #2902 (stark) |
| DrivAerML | `val_primary/surface_rel_l2_pct` | **3.997%** | #2898 (piccolo) |

## Expected Outcome

The global context token adds minimal parameters (~dim^2 per cross-attention layer) but provides access to global flow statistics. This is especially relevant for DrivAerML where the model must capture car-scale pressure distributions. Report validation metrics from the best checkpoint for all 4 datasets.

### DIFF
--- a/target/icml2026/core/architectures/senpai_transolver.py
+++ b/target/icml2026/core/architectures/senpai_transolver.py
@@ -122,6 +122,37 @@ class ANPSurfaceDecoder(nn.Module):
         return correction
 
 
+class GlobalContextBlock(nn.Module):
+    """BERT [CLS]-style global context token that aggregates global flow field
+    information via cross-attention and broadcasts it back to all nodes."""
+
+    def __init__(self, dim: int, num_heads: int = 8):
+        super().__init__()
+        self.global_token = nn.Parameter(torch.zeros(1, 1, dim))
+        nn.init.trunc_normal_(self.global_token, std=0.02)
+        self.cross_attn = nn.MultiheadAttention(
+            embed_dim=dim,
+            num_heads=num_heads,
+            batch_first=True,
+            dropout=0.0,
+        )
+        self.norm1 = nn.LayerNorm(dim)
+        self.norm2 = nn.LayerNorm(dim)
+        self.proj = nn.Linear(dim, dim)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        B = x.shape[0]
+        g = self.global_token.expand(B, -1, -1)
+        g_updated, _ = self.cross_attn(
+            query=self.norm1(g),
+            key=self.norm2(x),
+            value=x,
+        )
+        g = g + g_updated
+        global_bias = self.proj(g)
+        return x + global_bias
+
+
 class SenpaiTransolver(ReferenceTransolver):
     """Grouped-domain Transolver plus the retained Senpai refinement mechanisms."""
 
@@ -134,6 +165,8 @@ class SenpaiTransolver(ReferenceTransolver):
         surface_refine_layers: int = 2,
         surface_pressure_prior_idx: int | None = None,
         volume_pressure_prior_idx: int | None = None,
+        global_context_token: bool = False,
+        global_context_token_interval: int = 2,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -151,6 +184,19 @@ class SenpaiTransolver(ReferenceTransolver):
             if self.surface_refine
             else None
         )
+        self.use_global_context = global_context_token
+        if global_context_token:
+            n_blocks = len(self.backbone.blocks)
+            n_heads = self.n_hidden // 64 if self.n_hidden >= 64 else 1
+            # Place one GlobalContextBlock after every K transformer blocks
+            self.global_ctx_interval = global_context_token_interval
+            n_ctx_blocks = (n_blocks + global_context_token_interval - 1) // global_context_token_interval
+            self.global_ctx_blocks = nn.ModuleList(
+                [GlobalContextBlock(dim=self.n_hidden, num_heads=n_heads) for _ in range(n_ctx_blocks)]
+            )
+        else:
+            self.global_ctx_interval = global_context_token_interval
+            self.global_ctx_blocks = nn.ModuleList()
 
     def _apply_pressure_prior_addition(
         self,
@@ -170,6 +216,24 @@ class SenpaiTransolver(ReferenceTransolver):
         preds[..., self.pressure_output_index] = preds[..., self.pressure_output_index] + x[..., prior_idx]
         return preds
 
+    def _forward_backbone_with_global_ctx(
+        self,
+        hidden: torch.Tensor,
+        attn_mask: torch.Tensor,
+    ) -> torch.Tensor:
+        """Run backbone blocks, injecting GlobalContextBlock after every K blocks."""
+        ctx_idx = 0
+        for block_idx, block in enumerate(self.backbone.blocks):
+            hidden = block(hidden, attn_mask=attn_mask)
+            # Apply global context after every K-th block (1-indexed)
+            if (block_idx + 1) % self.global_ctx_interval == 0 and ctx_idx < len(self.global_ctx_blocks):
+                hidden = self.global_ctx_blocks[ctx_idx](hidden)
+                ctx_idx += 1
+        # Apply any remaining global context blocks after the final backbone block
+        if ctx_idx < len(self.global_ctx_blocks):
+            hidden = self.global_ctx_blocks[ctx_idx](hidden)
+        return hidden
+
     def forward(
         self,
         *,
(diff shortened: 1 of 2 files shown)


tcapelle:
STUDENT kohaku:

## Results

### Summary

The global context token (`--global-context-token`) **did not beat baseline on any of the 3 datasets tested.** TandemFoil Paper could not be tested due to a pre-existing data pipeline error.

| Dataset | Metric | Baseline | Global Ctx | Delta | Verdict |
|---------|--------|----------|------------|-------|---------|
| TandemFoil | `val_primary/surface_pressure_mae` | **26.06** | 26.85 | +3.0% | ❌ Worse |
| AirfRANS | `val_primary/surface_mse` | **0.000627** | 0.000699 | +11.5% | ❌ Worse |
| DrivAerML | `val_primary/surface_rel_l2_pct` | **3.997%** | 7.364% | +84.2% | ❌ Much worse |
| TandemFoil Paper | `val_primary/field_mse` | TBD | N/A | — | ⚠️ Data error |

### Per-Dataset Details

#### TandemFoil (W&B run: `a5610z74`)
- **Best val:** 26.85 at epoch 271 (273 total epochs, 360-min budget)
- **Best test:** 28.07 at epoch 273
- Oscillation pattern: range ~26.8–30.5, typical cosine cycling at T_max=10
- The global context token added marginal overhead but did not improve the loss landscape
- **Command:**
  ```
  CUDA_VISIBLE_DEVICES=0 python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999 --global-context-token --wandb-group kohaku-global-ctx --agent kohaku --wandb-name "kohaku/tandemfoil-global-ctx"
  ```

#### AirfRANS (W&B run: `84d7wgf4`)
- **Best val:** 0.000699 at epoch 543 (674 total epochs, 360-min budget)
- **Best test:** 0.003060 at epoch 674 (final model diverged from best)
- Late-epoch instability: best at ep543, then oscillated 0.001–0.005 range
- External target: 0.0043 — still beats external target (0.000699 < 0.0043) but does not beat our internal best of 0.000627
- **Command:**
  ```
  CUDA_VISIBLE_DEVICES=2 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --global-context-token --wandb-group kohaku-global-ctx --agent kohaku --wandb-name "kohaku/airfrans-global-ctx"
  ```

#### DrivAerML (W&B run: `vwpms4wu`)
- **Best val:** 7.364% at epoch 146 (340 total epochs, 360-min budget, SENPAI_MAX_EPOCHS=9999)
- Model diverged badly after ep146 — late epochs showed 50–75% error
- The cross-attention global context seems to hurt DrivAerML's 4L/512d architecture the most, likely due to the mismatch between the small global token and the large (50K surface point) token sequences
- External target: <3.71% — this result does not approach external target
- **Command:**
  ```
  CUDA_VISIBLE_DEVICES=3 SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --global-context-token --wandb-group kohaku-global-ctx --agent kohaku --wandb-name "kohaku/drivaerml-global-ctx"
  ```

#### TandemFoil Paper — Data Pipeline Error
- Crashed before training with: `ValueError: Expected paper-style tandem AoA to be shared, got -0.94 vs -2.36 for raceCar_randomFields_mgn_Part1.pickle:0` in `tandemfoil_paper/data/split_paper_experiment4.py:192`
- This is a data validation issue in the data pipeline (not in `train.py`), so it cannot be fixed within my boundary

### What happened

The global context token hypothesis did not hold. A single learnable [CLS]-style token accumulating global flow information via cross-attention and broadcasting it as an additive bias was not beneficial:

1. **TandemFoil (+3.0%):** The model already captures sufficient global context through the slice-based attention mechanism with 64 slices. The additional cross-attention overhead (computing attention over all nodes to update a single global token) slightly degraded optimization without adding useful information. The cosine cycling pattern was identical to baseline, suggesting the global token was not changing the loss landscape meaningfully.

2. **AirfRANS (+11.5%):** At 2L depth, injecting a global context block after each layer (interval=2 means 1 block total at 2L) added a noisy signal. The model's existing Fourier features already provide frequency-domain global information, making the cross-attention global token redundant.

3. **DrivAerML (+84.2%):** The worst impact. The cross-attention between a single global token and 50K surface nodes is a severe bottleneck — the global token has insufficient capacity to summarize the 3D car geometry meaningfully. The additive broadcast then corrupts all node features. This caused catastrophic divergence after ep146.

### Suggested follow-ups

- **NOT recommended:** Scaling up global context (more tokens, larger dim) — adds quadratic cost with no evidence of useful signal.
- **Possible:** A pooling-based global context (mean/max pool instead of cross-attention) would be much cheaper and might avoid the cross-attention bottleneck, but the signal itself (global mean features) may still not help given that slice attention already provides non-local context.
- The TandemFoil Paper data pipeline error should be investigated separately — it appears to be a data integrity issue in `split_paper_experiment4.py`.

tcapelle:
ADVISOR: Closing PR #2957 because ADVISOR: Good systematic testing, kohaku. Global Context Token is a clear dead end — degraded results on all three datasets (TF: +3%, AirfRANS: +11.5%, DrivAerML: +84.2% → diverged). The slice-based attention in Transolver already captures non-local context effectively, so a CLS-style global token adds overhead without useful information. For DrivAerML specifically, summarizing 50K surface nodes into a single token is too lossy. This approach is not worth pursuing further. Closing..

PR #2962, opened 2026-04-22
wave3: Adaptive Gradient Clipping AGC NFNet-style (cross-dataset)

tcapelle:
## Hypothesis

Adaptive Gradient Clipping (AGC), as introduced in NFNets (Brock et al. 2021), clips gradients based on the ratio of the gradient norm to the parameter norm rather than using a fixed global clip threshold. This is more stable than fixed gradient clipping — it prevents very large updates relative to parameter scale while allowing larger updates where parameters are small. Replacing fixed `--grad-clip 1.0` with AGC may enable faster convergence and better final accuracy across all four datasets.

## Implementation

Add to `train.py`:

```python
def adaptive_gradient_clip(parameters, clip_factor=0.01, eps=1e-3):
    """
    Adaptive Gradient Clipping (AGC) as in NFNet.
    Clips grad norm to clip_factor * param_norm for each layer.
    
    Args:
        parameters: model parameters (iterable)
        clip_factor: AGC clipping factor (default 0.01)
        eps: minimum parameter norm floor to prevent instability
    """
    for p in parameters:
        if p.grad is None:
            continue
        # Compute norms per parameter tensor
        p_norm = torch.norm(p.detach(), p=2).clamp(min=eps)
        g_norm = torch.norm(p.grad.detach(), p=2)
        
        # Clip grad if g_norm > clip_factor * p_norm
        max_norm = clip_factor * p_norm
        clip_coef = max_norm / g_norm.clamp(min=1e-8)
        if clip_coef < 1.0:
            p.grad.detach().mul_(clip_coef)

# Add CLI arguments:
parser.add_argument('--agc', action='store_true',
    help='Use Adaptive Gradient Clipping (NFNet-style) instead of fixed clip')
parser.add_argument('--agc-clip-factor', type=float, default=0.01,
    help='AGC clipping factor (default 0.01 as in NFNet paper)')

# In training loop, replace:
#   torch.nn.utils.clip_grad_norm_(model.parameters(), args.grad_clip)
# With:
if args.agc:
    adaptive_gradient_clip(model.parameters(), clip_factor=args.agc_clip_factor)
else:
    torch.nn.utils.clip_grad_norm_(model.parameters(), args.grad_clip)
```

Test two values: `--agc-clip-factor 0.01` (NFNet default) and `--agc-clip-factor 0.03`.

Use `--wandb_group wave3-adaptive-grad-clipping` for grouping.

### TandemFoil (AGC clip_factor=0.01)
```bash
python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 \
  --cosine-t-max 10 --weight-decay 1e-2 --no-use-ema \
  --agc --agc-clip-factor 0.01 \
  --epochs 999 --wandb_group wave3-adaptive-grad-clipping
```

Note: do NOT pass `--grad-clip` when using `--agc`.

### TandemFoil Paper
```bash
python train.py --dataset tandemfoil_paper --optimizer lion --lr 1e-4 \
  --cosine-t-max 10 --weight-decay 1e-2 --ema-decay 0.999 \
  --agc --agc-clip-factor 0.01 \
  --epochs 999 --wandb_group wave3-adaptive-grad-clipping
```

### AirfRANS
```bash
  --weight-decay 1e-2 --no-use-ema \
  --agc --agc-clip-factor 0.01 \
  --epochs 999 --wandb_group wave3-adaptive-grad-clipping
```

### DrivAerML
```bash
  --epochs 999 --batch-size 1 \
  --agc --agc-clip-factor 0.01 \
  --wandb_group wave3-adaptive-grad-clipping
```

Also run a variant with `--agc-clip-factor 0.03` on DrivAerML (DrivAerML has the most complex gradient landscape).

## Current Baselines

| Dataset | Metric | Best Value | PR |
|---------|--------|-----------|-----|
| TandemFoil | val_primary/surface_pressure_mae | 26.06 | #2887 |
| TandemFoil Paper | val_primary/field_mse | TBD | #2947/2948/2949 |
| AirfRANS | val_primary/surface_mse | 0.000627 | #2902 |

## Expected Outcome

AGC provides layer-wise gradient normalization that is more adaptive than the global fixed clip. In the NFNet paper, AGC enabled training without BatchNorm at high learning rates. In our point-cloud transformers, AGC should prevent the common issue of gradient explosion in the attention layers (large query/key inner products) while still allowing the FFN layers to update aggressively. Expected: 1-3% improvement, especially beneficial on DrivAerML where gradient instability is more common (complex 3D geometry with surface normal discontinuities).

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -164,6 +164,8 @@ class TrainConfig:
     volume_anchor_points: int = 8_000
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
+    agc: bool = False
+    agc_clip_factor: float = 0.01
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -1178,6 +1180,25 @@ def evaluate_abupt(
     return {"surface_mae": total_surface / max(count_surface, 1), "volume_mae": total_volume / max(count_volume, 1)}
 
 
+def adaptive_gradient_clip(parameters, clip_factor: float = 0.01, eps: float = 1e-3) -> tuple[int, float]:
+    clipped = 0
+    total_grad_norm_sq = 0.0
+    for p in parameters:
+        if p.grad is None:
+            continue
+        g = p.grad.detach()
+        total_grad_norm_sq += g.norm(2).item() ** 2
+        if p.ndim < 2:
+            continue
+        p_norm = p.detach().norm(2).clamp_(min=eps)
+        g_norm = g.norm(2)
+        max_norm = clip_factor * p_norm
+        if g_norm > max_norm:
+            g.mul_(max_norm / g_norm.clamp(min=1e-8))
+            clipped += 1
+    return clipped, total_grad_norm_sq ** 0.5
+
+
 def train_one_epoch(
     model: torch.nn.Module,
     anp_head: ANPSurfaceDecoder | None,
@@ -1194,6 +1215,8 @@ def train_one_epoch(
     max_batches: int = 0,
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
+    agc: bool = False,
+    agc_clip_factor: float = 0.01,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
@@ -1261,12 +1284,18 @@ def train_one_epoch(
         all_params = list(model.parameters())
         if anp_head is not None:
             all_params += list(anp_head.parameters())
-        grad_norm = torch.nn.utils.clip_grad_norm_(all_params, max_norm=grad_clip if grad_clip > 0 else float("inf"))
+        if agc:
+            agc_clipped, grad_norm_val = adaptive_gradient_clip(all_params, clip_factor=agc_clip_factor)
+            grad_norm = torch.tensor(grad_norm_val)
+            running.setdefault("agc_clip_events", 0.0)
+            running["agc_clip_events"] += float(agc_clipped)
+        else:
+            grad_norm = torch.nn.utils.clip_grad_norm_(all_params, max_norm=grad_clip if grad_clip > 0 else float("inf"))
+            if grad_clip > 0 and float(grad_norm) > grad_clip:
+                running.setdefault("grad_clip_events", 0.0)
+                running["grad_clip_events"] += 1.0
         running.setdefault("grad_norm_mean", 0.0)
         running["grad_norm_mean"] += float(grad_norm)
-        if grad_clip > 0 and float(grad_norm) > grad_clip:
-            running.setdefault("grad_clip_events", 0.0)
-            running["grad_clip_events"] += 1.0
         optimizer.step()
         if scheduler is not None:
             scheduler.step()
@@ -1462,6 +1491,8 @@ def main() -> None:
             max_batches=config.max_train_batches,
             grad_clip=config.grad_clip,
             grad_accum_steps=config.grad_accum_steps,
+            agc=config.agc,
+            agc_clip_factor=config.agc_clip_factor,
         )
 
         if ema is not None:


tcapelle:
STUDENT casca:

## Results

### Summary

AGC (Adaptive Gradient Clipping) **performed significantly worse than fixed gradient clipping** across all datasets. The hypothesis is **negative** — replacing `--grad-clip 1.0` with AGC does not improve convergence or final accuracy for our point-cloud transformer architecture.

### Metrics

| Dataset | Metric | Baseline | AGC 0.01 (best val) | AGC 0.01 (test) | vs Baseline |
|---------|--------|----------|---------------------|-----------------|-------------|
| TandemFoil | surface_pressure_mae ↓ | **26.06** | 33.28 (ep 231) | 37.18 | **28% worse** |
| AirfRANS | surface_mse ↓ | **0.000627** | 0.001544 (ep 448) | 0.003327 | **2.5x worse** |
| DrivAerML | surface_rel_l2_pct ↓ | **3.997%** | 7.683% (ep 119) | 40.66% | **1.9x worse (val), 10x worse (test)** |
| DrivAerML (0.03) | surface_rel_l2_pct ↓ | **3.997%** | 6.699% (ep 158) | NaN | **Diverged to NaN at epoch 240** |
| TandemFoil Paper | — | TBD | — | — | Data loading crash (pre-existing) |

### Commands Used

**TandemFoil (GPU 0):**
```bash
CUDA_VISIBLE_DEVICES=0 python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 \
  --cosine-t-max 10 --weight-decay 1e-2 --no-use-ema \
  --agc --agc-clip-factor 0.01 \
  --epochs 999 --wandb-group wave3-adaptive-grad-clipping --agent casca
```

**AirfRANS (GPU 2):**
```bash
CUDA_VISIBLE_DEVICES=2 python train.py --dataset airfrans --airfrans-task full \
  --weight-decay 1e-2 --no-use-ema \
  --agc --agc-clip-factor 0.01 \
  --epochs 999 --wandb-group wave3-adaptive-grad-clipping --agent casca
```

**DrivAerML AGC 0.01 (GPU 3):**
```bash
SENPAI_MAX_EPOCHS=9999 CUDA_VISIBLE_DEVICES=3 python train.py --dataset drivaerml \
  --epochs 999 --batch-size 1 \
  --agc --agc-clip-factor 0.01 \
  --wandb-group wave3-adaptive-grad-clipping --agent casca
```

**DrivAerML AGC 0.03 (GPU 5):**
Same as above with `--agc-clip-factor 0.03`.

### Peak Memory
- TandemFoil (GPU 0): ~49 GB
- AirfRANS (GPU 2): ~47 GB
- DrivAerML (GPU 3/5): ~85 GB

### W&B Run IDs
- TandemFoil: [`239mz5it`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/239mz5it)
- AirfRANS: [`kqulq69i`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/kqulq69i)
- DrivAerML 0.01: [`0z9ljqmy`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/0z9ljqmy)
- DrivAerML 0.03: [`eeberdy2`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/eeberdy2)

### What Happened

AGC clips gradients per-parameter based on the ratio of gradient norm to parameter norm, rather than applying a global clip. In theory this should be more adaptive, but in practice it performed uniformly worse:

1. **DrivAerML diverged**: With clip_factor=0.03, training went NaN at epoch 240. With 0.01, grad_norm_mean reached Infinity in later epochs (AGC kept individual parameters in check but couldn't prevent global instability). The test metric of 40.66% vs 3.997% baseline shows the model never converged properly.

2. **TandemFoil plateaued high**: Best val of 33.28 vs 26.06 baseline (28% worse). The Lion optimizer already compresses gradients to their sign, so AGC's per-parameter magnitude clipping provides minimal additional benefit — it mainly acts as an overly aggressive regularizer.

3. **AirfRANS underperformed**: Best val of 0.001544 vs 0.000627 baseline (2.5x worse). The val metric was notably noisy (oscillating between 0.0015 and 0.01), suggesting AGC's per-parameter clipping introduces optimization instability.

4. **AGC clip events were saturating**: 14,700+ clip events per epoch on DrivAerML (nearly every parameter clipped every step), confirming the researcher-agent's finding that clip_factor=0.01 is too aggressive for transformers — it acts as a strong regularizer rather than a stability guard.

### Implementation Notes

- Followed the NFNet paper's recommendation to skip AGC for 1D parameters (biases, LayerNorm weights) via `p.ndim < 2` check
- Used full-tensor norms per parameter (simpler variant; the paper's per-unit/per-row clipping could be tried but unlikely to change the conclusion)
- Still computed and logged total grad_norm for wandb tracking
- Logged `agc_clip_events` per epoch for monitoring

### TandemFoil Paper Note

The tandemfoil_paper experiment crashed during data loading with `ValueError: Expected paper-style tandem AoA to be shared, got -0.94 vs -2.36`. This is a pre-existing dataset issue in `split_paper_experiment4.py`, not related to AGC.

### Suggested Follow-ups

- **Do not pursue AGC further** for this architecture. Fixed global gradient clipping (`--grad-clip 1.0`) is clearly superior.
- The research literature (AdaGC, ICML 2025) suggests adaptive per-parameter thresholds using EMA-based estimates, but the fundamental problem is that per-parameter clipping doesn't constrain the global gradient magnitude, which is what matters for optimizer stability.
- If gradient stability is a concern, a more promising direction would be **gradient accumulation with gradient normalization** or **gradient penalty** rather than per-parameter clipping.

tcapelle:
ADVISOR: Closing PR #2962 because ADVISOR: Closing as a clear dead end. AGC performed catastrophically worse than baseline across all three datasets:

- TandemFoil: 33.28 vs baseline 22.537 (48% worse)
- AirfRANS: 0.001544 vs baseline 0.000482 (3.2x worse)
- DrivAerML: AGC 0.01 → 40.66% test (10x worse than 3.997%), AGC 0.03 → diverged to NaN at epoch 240

**Root cause analysis:** AGC's per-parameter magnitude-ratio clipping is fundamentally incompatible with our setup. For Lion optimizer (used on TandemFoil), Lion already compresses all gradients to their sign — p_norm vs g_norm comparisons become meaningless on sign-only gradient tensors. For AdamW on DrivAerML's large 4L/512d architecture, the default clip_factor=0.01 was far too aggressive, essentially clipping almost every gradient from the start of training and preventing convergence. The divergence at epoch 240 with clip_factor=0.03 suggests the cosine LR cycling re-heated a critically under-trained model.

The implementation itself is correct — the logic matches NFNet AGC — but the method simply does not translate to our training regime. Closing. Griffith's spectral norm follow-up ideas (sigma-Reparam) are a better next direction for stabilizing attention..

PR #2966, opened 2026-04-22
wave3: Huber Loss for Robust Regression (cross-dataset)

tcapelle:
## Hypothesis

Replace MSE loss with Huber loss (smooth L1) for more robust regression training. MSE penalizes outliers quadratically, which can destabilize training when CFD fields contain rare high-magnitude anomalies (e.g. boundary layer separation, shock-like gradients near sharp geometric features). Huber loss behaves like L1 for large residuals (linear penalty) and L2 for small residuals (smooth near zero), controlled by delta. This may improve generalization by reducing the model's sensitivity to hard-to-predict outlier points.

**Expected outcome:** Improved surface pressure MAE on TandemFoil (target: <26.06), surface MSE on AirfRANS (target: <0.000627), surface_rel_l2_pct on DrivAerML (target: <3.997%), and field_mse on TandemFoil Paper.

## Implementation

Add `--loss-fn huber --huber-delta DELTA` flags to `train.py`. Wrap the primary loss computation:

```python
# In train.py, loss computation block:
import torch.nn.functional as F

if args.loss_fn == 'huber':
    # Huber loss: L2 when |error| < delta, L1 scaled otherwise
    loss = F.huber_loss(pred, target, reduction='mean', delta=args.huber_delta)
elif args.loss_fn == 'mse':
    loss = F.mse_loss(pred, target, reduction='mean')
else:
    loss = F.mse_loss(pred, target, reduction='mean')
```

Add to argparse:
```python
parser.add_argument('--loss-fn', type=str, default='mse', choices=['mse', 'huber'],
                    help='Loss function: mse (default) or huber (smooth L1)')
parser.add_argument('--huber-delta', type=float, default=1.0,
                    help='Delta threshold for Huber loss (default: 1.0)')
```

**Key insight on delta selection:** Since surface pressure values are normalized, delta should be tuned per-dataset. Start with delta=1.0 (standard Huber). Also test delta=0.1 (more L1-like, stronger outlier robustness) and delta=0.5. For DrivAerML's surface_rel_l2_pct metric (already L2-based), delta=1.0 is likely best.

### TandemFoil (golden config + Huber loss)
```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --optimizer lion --lr 1e-4 --cosine-t-max 10 \
  --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema \
  --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
[...]

PR #2901, opened 2026-04-21
Cross-dataset: Huber loss — robust to outlier mesh nodes at stagnation points and trailing edges (CODE CHANGE)

tcapelle:
## Hypothesis

MSE loss treats all prediction errors equally, but CFD meshes have high-error outlier nodes at stagnation points, trailing edges, and separation zones. These outliers disproportionately dominate the MSE gradient, pulling the model away from good predictions on the majority of well-behaved nodes. Huber loss (and optionally log-cosh) clips the gradient contribution from large-error nodes, providing robustness while remaining convex and smooth. We hypothesize this will improve mean surface metrics across all three datasets by reducing the distortion caused by outlier nodes.

Reference: Huber (1964), "Robust Estimation of a Location Parameter"; log-cosh is a smooth approximation with similar robustness properties.

## Code Change Required

Add the following to `target/icml2026/train.py`:

```python
# Add CLI args near other loss/training args:
parser.add_argument('--loss-type', type=str, default='mse',
                    choices=['mse', 'huber', 'log_cosh'],
                    help='Loss function type. mse=default, huber=robust to outliers, log_cosh=smooth robust.')
parser.add_argument('--huber-delta', type=float, default=1.0,
                    help='Delta threshold for Huber loss (only used when --loss-type huber).')

# Replace the loss function instantiation (wherever criterion/loss is set up):
if args.loss_type == 'huber':
    criterion = nn.HuberLoss(delta=args.huber_delta)
elif args.loss_type == 'log_cosh':
    def log_cosh_loss(pred, target):
        return torch.mean(torch.log(torch.cosh(pred - target)))
    criterion = log_cosh_loss
else:
    criterion = nn.MSELoss()
```

Make sure the new `criterion` is used wherever the loss is computed during training and validation.

## Instructions

Run the following matrix across all three datasets. Use `--wandb_group huber-loss-cross` on all runs so they are grouped in W&B.

**AirfRANS — golden config baseline + Huber/log-cosh variants:**

```bash
# AF control (MSE — confirm baseline)
python train.py --dataset airfrans --airfrans-task full \
  --optimizer adamw --lr 7e-4 --cosine-t-max 5 \
  --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier \
  --model-layers 2 --model-hidden-dim 256 --model-heads 4 \
  --epochs 999 --loss-type mse \
  --wandb_group huber-loss-cross --wandb_run_name af-mse-control

# AF Huber delta=1.0
[...]

PR #2975, opened 2026-04-22
RoPE: Rotary Positional Embeddings on attention layers (all 4 datasets)

tcapelle:
## Hypothesis

The current model uses no explicit positional encoding on attention layers — point coordinates are injected as input features but the transformer attention mechanism has no notion of relative geometry. **Rotary Positional Embeddings (RoPE)** encode relative position directly into the query-key dot product, which has shown strong gains in transformers operating on irregular spatial data (e.g. 3D point clouds).

For CFD surrogates, spatial locality is physically meaningful: nearby points interact more strongly through pressure gradients and boundary layer effects. RoPE gives the attention mechanism a learnable geometric inductive bias that vanilla attention lacks.

**Hypothesis**: Adding RoPE to the attention QK computation (encoding the 3D coordinates x/y/z as rotary angles) will improve surface pressure and field predictions across all datasets by allowing the model to reason about relative spatial distances between mesh nodes.

## Implementation

In `target/icml2026/train.py` (or the model definition), modify the multi-head attention layers:

1. For each attention layer, after computing Q and K projections, apply rotary embeddings using the node's (x, y, z) coordinates as the rotation angles:
   ```python
   # Standard RoPE for 3D: partition Q/K head dims into 3 pairs, rotate each by (theta_x, theta_y, theta_z)
   # Or use a 1D RoPE on a learned projection of xyz coordinates
   ```

2. Simplest implementation: use a **learned linear projection** of (x, y, z) → d_model/2 scalar angles, then apply 2D RoPE rotations. Alternatively use `xformers` or `rotary-embedding-torch` if available.

3. Apply RoPE **per layer** (all attention layers share the same RoPE scheme, different learned projections per layer is also fine).

4. Keep all other hyperparameters identical to the current best config per dataset.

**Suggested flag**: `--rope-dim 32` (number of RoPE embedding dimensions, must be even; try 32 as default). If a flag doesn't exist, implement it with this default.

### TandemFoil
```bash
python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999 --wandb_group rope-embeddings
```

### AirfRANS
```bash
python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --wandb_group rope-embeddings
```

### DrivAerML
```bash
SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --wandb_group rope-embeddings
```

### TandemFoil Paper
```bash
python train.py --dataset tandemfoil_paper --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999 --wandb_group rope-embeddings
```

| AirfRANS | `val_primary/surface_mse` | **0.000627** | #2902 |
| TandemFoil Paper | `val_primary/field_mse` | *no baseline yet* | — |

## Expected Outcome

RoPE has consistently improved spatial transformer models in point cloud and mesh settings. Expected improvement: 1-3% on surface pressure metrics across TF and AF. DrivAerML with its large 3D geometry may benefit the most from explicit rotational geometry encoding.

## Prior Results to Note

- Spike's prior PR #2901 (Huber loss) was sent back; Huber is now covered by Wave 3 (#2966 gohan), so this is a fresh direction
- No RoPE experiments have been attempted on this codebase

### DIFF
--- a/target/icml2026/core/architectures/transolver_reference.py
+++ b/target/icml2026/core/architectures/transolver_reference.py
@@ -18,6 +18,38 @@ def _init_linear(module: nn.Module, std: float = 0.02) -> None:
             nn.init.zeros_(module.bias)
 
 
+def apply_rope_nd(x: torch.Tensor, positions: torch.Tensor, rope_dim: int) -> torch.Tensor:
+    """Apply N-dimensional Rotary Position Embeddings using continuous spatial coordinates.
+
+    x: [..., D] tensor (query or key)
+    positions: [..., space_dim] continuous coordinates
+    rope_dim: number of head dimensions to rotate (rest pass through unchanged)
+    """
+    D = x.shape[-1]
+    space_dim = positions.shape[-1]
+    dims_per_axis = (rope_dim // space_dim) & ~1
+    if dims_per_axis < 2:
+        return x
+    total_rope = dims_per_axis * space_dim
+    half = dims_per_axis // 2
+    freq_seq = torch.arange(half, device=x.device, dtype=torch.float32) / half
+    freqs = 1.0 / (10000.0 ** freq_seq)
+    angles = positions.unsqueeze(-1) * freqs
+    angles = angles.flatten(-2)
+    cos_a = angles.cos().to(x.dtype)
+    sin_a = angles.sin().to(x.dtype)
+    x_rope = x[..., :total_rope]
+    x_rest = x[..., total_rope:]
+    x1 = x_rope[..., 0::2]
+    x2 = x_rope[..., 1::2]
+    y1 = cos_a * x1 - sin_a * x2
+    y2 = sin_a * x1 + cos_a * x2
+    x_rotated = torch.stack([y1, y2], dim=-1).flatten(-2)
+    if x_rest.shape[-1] > 0:
+        return torch.cat([x_rotated, x_rest], dim=-1)
+    return x_rotated
+
+
 class LinearProjection(nn.Module):
     def __init__(self, input_dim: int, output_dim: int, bias: bool = True):
         super().__init__()
@@ -85,7 +117,7 @@ class UpActDownMlp(nn.Module):
 
 
 class TransolverAttention(nn.Module):
-    def __init__(self, hidden_dim: int, num_heads: int, num_slices: int, dropout: float = 0.0):
+    def __init__(self, hidden_dim: int, num_heads: int, num_slices: int, dropout: float = 0.0, rope_dim: int = 0):
         super().__init__()
         if hidden_dim % num_heads != 0:
             raise ValueError("hidden_dim must be divisible by num_heads")
@@ -94,6 +126,7 @@ class TransolverAttention(nn.Module):
         self.dim_head = hidden_dim // num_heads
         self.num_slices = num_slices
         self.dropout = dropout
+        self.rope_dim = rope_dim
 
         self.temperature = nn.Parameter(torch.full((1, num_heads, 1, 1), 0.5))
         self.in_project_x = LinearProjection(hidden_dim, hidden_dim)
@@ -116,10 +149,16 @@ class TransolverAttention(nn.Module):
         slice_tokens = torch.einsum("bhnc,bhns->bhsc", fx_mid, slice_weights) / (slice_norm + 1e-5)
         return slice_tokens, slice_weights
 
-    def forward(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None) -> torch.Tensor:
+    def forward(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None, coords: torch.Tensor | None = None) -> torch.Tensor:
         slice_tokens, slice_weights = self.create_slices(x, attn_mask=attn_mask)
         qkv = self.qkv(slice_tokens)
         q, k, v = qkv.chunk(3, dim=-1)
+        if self.rope_dim > 0 and coords is not None:
+            slice_mass = slice_weights.sum(dim=2)
+            centroids = torch.einsum("bhns,bnp->bhsp", slice_weights, coords)
+            centroids = centroids / (slice_mass.unsqueeze(-1) + 1e-8)
+            q = apply_rope_nd(q, centroids, self.rope_dim)
+            k = apply_rope_nd(k, centroids, self.rope_dim)
         out_slice = F.scaled_dot_product_attention(
             q,
             k,
@@ -139,6 +178,7 @@ class TransformerBlock(nn.Module):
         mlp_expansion_factor: int | float,
         num_slices: int,
         dropout: float = 0.0,
+        rope_dim: int = 0,
     ):
         super().__init__()
         mlp_hidden_dim = int(math.ceil(hidden_dim * mlp_expansion_factor))
@@ -148,12 +188,13 @@ class TransformerBlock(nn.Module):
             num_heads=num_heads,
             num_slices=num_slices,
             dropout=dropout,
+            rope_dim=rope_dim,
         )
         self.norm2 = nn.LayerNorm(hidden_dim, eps=1e-6)
         self.mlp = UpActDownMlp(hidden_dim=hidden_dim, mlp_hidden_dim=mlp_hidden_dim)
 
-    def forward(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None) -> torch.Tensor:
-        x = x + self.attention(self.norm1(x), attn_mask=attn_mask)
+    def forward(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None, coords: torch.Tensor | None = None) -> torch.Tensor:
+        x = x + self.attention(self.norm1(x), attn_mask=attn_mask, coords=coords)
         x = x + self.mlp(self.norm2(x))
         return x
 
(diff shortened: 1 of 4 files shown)


tcapelle:
STUDENT spike:

## Results

### RoPE (dim=32) Centroid Positional Embeddings — Cross-Dataset

**Hypothesis:** Applying Rotary Positional Embeddings to the Transolver's attention layers will improve spatial awareness and reduce prediction error across all datasets.

**Implementation:** RoPE is applied to Q/K vectors in each TransolverAttention layer using **slice centroid positions** — weighted spatial averages of node coordinates per slice token. This adapts RoPE (normally for discrete token positions) to the Transolver's slice-based attention where tokens are abstract aggregates rather than individual spatial points. Coordinates are threaded from the model input through the full backbone chain.

### Metrics Summary

| Dataset | Metric | Baseline | RoPE-32 Best | Epoch | Delta |
|---|---|---|---|---|---|
| AirfRANS | surface_mse | **0.000627** | 0.00573 | 87/90 | **+813% (9.1x worse)** |
| TandemFoil | surface_pressure_mae | **26.06** | 67.14 | 10/10 | **+158% (2.6x worse)** |
| DrivAerML | surface_rel_l2_pct | **3.997%** | 8.82% | 68/71 | **+121% (2.2x worse)** |
| TandemFoil Paper | field_mse | N/A | 0.0248 | 33/33 | No baseline |

**Note:** TandemFoil is only at epoch 10 (very slow due to 30-min torch.compile per launch). The baseline reached 26.06 at epoch 300. At similar early epoch counts, the RoPE trajectory (165→133→115→103→93→88→83→76→69→67) is still improving and may eventually plateau closer to baseline. However, the clear regressions on AirfRANS (90 epochs, 6+ cosine troughs) and DrivAerML (71 epochs, 2+ cosine troughs) are strong evidence that RoPE via centroids hurts.

### Per-Dataset Details

**AirfRANS** (W&B: `qn46wubf`, GPU 0, ~39 GiB peak)
- Best surface_mse: 0.00573 at ep87 (baseline 0.000627, external target 0.0043)
- Best volume_mse: 0.0365 at ep86
- Metric oscillates 0.006–0.034 across cosine cycles (T_max=10)
- 90 epochs completed, clear plateau

**TandemFoil** (W&B: `24pypzzr`, GPU 1, ~33 GiB peak)
- Best surface_pressure_mae: 67.14 at ep10 (baseline 26.06)
- Per-split at ep10: cruise=50.72, racecar=81.80, re_rand=65.88, single_in_dist=70.16
- Only 10 epochs — still improving, but trajectory suggests won't match baseline

**DrivAerML** (W&B: `dmly303i`, GPU 2, ~6 GiB active)
- Best surface_rel_l2_pct: 8.82% at ep68 (baseline 3.997%)
- 71 epochs completed, oscillating 8.8–14.4% across T_max=30 cycles
- Metric still slowly improving at cosine troughs

**TandemFoil Paper** (W&B: `g3zd26b5`, GPU 3, ~43 GiB peak)
- Best field_mse: 0.0248 at ep33 (no established baseline)
- Best surface_mse: 0.0067 at ep33
- Still improving steadily

### Commands Used

```bash
# AirfRANS
CUDA_VISIBLE_DEVICES=0 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --rope-dim 32 --agent spike --wandb-name "spike/af-rope32" --wandb-group rope-embeddings

# TandemFoil
CUDA_VISIBLE_DEVICES=1 python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999 --rope-dim 32 --agent spike --wandb-name "spike/tf-rope32" --wandb-group rope-embeddings

# DrivAerML
CUDA_VISIBLE_DEVICES=2 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --rope-dim 32 --agent spike --wandb-name "spike/dam-rope32" --wandb-group rope-embeddings

# TandemFoil Paper
CUDA_VISIBLE_DEVICES=3 python train.py --dataset tandemfoil_paper --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999 --rope-dim 32 --agent spike --wandb-name "spike/tfp-rope32" --wandb-group rope-embeddings
```

### What Happened

**RoPE via slice centroids is a clear negative result across all datasets.**

The fundamental issue is the mismatch between RoPE's design assumptions and the Transolver architecture:

1. **Centroid positions are too coarse.** Standard RoPE works because each token has a precise spatial/sequential position. In the Transolver, attention operates on "slice tokens" — abstract aggregates of many mesh nodes with diffuse spatial extent. The weighted centroid is a poor summary of "where" a slice is, losing the spatial resolution RoPE needs to encode meaningful relative positions.

2. **Rotation fights learned groupings.** The Transolver learns to cluster nodes into slices based on physical flow patterns (boundary layers, wakes, pressure gradients), not spatial proximity. RoPE biases attention toward spatially close slices via its distance-decay property, which conflicts with the physically-motivated groupings the model learns.

3. **Spatial information is already available.** The model receives raw coordinates via input embeddings and Fourier features. RoPE adds a second, potentially conflicting spatial signal in the attention mechanism, creating interference rather than reinforcement.

4. **The rotation itself adds noise.** With only 32 RoPE dimensions out of the full head dimension, the rotated portion of Q/K dot products introduces a positional bias that the model hasn't learned to use effectively, while the un-rotated portion continues to carry the learned attention patterns. This hybrid dilutes attention quality.

### Suggested Follow-ups

1. **Node-level RoPE before slice aggregation:** Instead of applying RoPE at the slice level, apply it to the raw node-level Q/K before the soft-clustering step. This would give each mesh node a precise position encoding. However, this requires significant architecture changes since the current design computes QKV after slice aggregation.

2. **Learned positional bias:** Instead of the rigid RoPE rotation, add a learned relative position bias to the slice attention scores (similar to ALiBi or relative attention). This would let the model discover whether spatial proximity matters for each head, rather than forcing it.

3. **Per-node sinusoidal injection:** Add Fourier-based positional encoding directly to node features before slice aggregation, separate from the existing Fourier features. This would be additive rather than multiplicative (unlike RoPE), avoiding interference with learned attention patterns.

4. **RoPE on a standard (non-slice) attention variant:** Test RoPE on a traditional per-node attention transformer to isolate whether the issue is RoPE itself vs the centroid approximation. If RoPE helps per-node attention, it confirms the centroid is the bottleneck.

tcapelle:
ADVISOR: Closing PR #2975 because ADVISOR: Closing. RoPE via slice centroids fails across all datasets.

AirfRANS: 0.00573 (9.1x worse than baseline 0.000627)
TandemFoil: 67.14 at ep10 — trajectory 165→67 improving but AirfRANS/DM results make the direction untenable
DrivAerML: 8.82% (2.2x worse than 3.997% baseline)

The core problem: Slice tokens in Transolver are soft semantic aggregates of mesh nodes, not individual spatial points with well-defined positions. Using the weighted centroid of node coordinates as the 'position' for RoPE is geometrically meaningful for individual nodes but creates a noisy, non-monotonic position signal for slice tokens — slices can overlap spatially, and their centroids may not encode the information RoPE needs to work well.

RoPE is designed for contexts where position has a clear sequential or spatial meaning (tokens in a sentence, pixels in an image). The Transolver's slice-based attention breaks that assumption. The idea is fundamentally mismatched to this architecture.

Note: PR #2983 (zenitsu) is also running RoPE — that should be terminated or redirected when it checks in.

No follow-up warranted on RoPE for this architecture..

PR #2953, opened 2026-04-22
Wave3: Relative L2 Training Loss — cross-dataset

tcapelle:
## Hypothesis

Standard MSE loss treats all nodes equally, but CFD targets have wildly different magnitudes (stagnation pressure vs. wake velocity). Replacing MSE with a relative L2 loss — normalizing the squared error by the squared norm of the ground truth — should make the optimizer pay proportionally equal attention to low-magnitude regions (e.g. wake, suction side) and high-magnitude regions, improving surface pressure and field accuracy across all four datasets.

## Relative L2 Loss Formulation

```
L_rel = ||y_pred - y_true||_2 / (||y_true||_2 + eps),   eps = 1e-8
```

Three variants to try in one sweep (use `--wandb_group frieren-rel-l2`):
- **global**: normalize by the full-batch norm (single scalar)
- **per-point**: normalize per-node (divide each node error by its ground-truth norm)
- **mixed**: 0.5 * L_global + 0.5 * L_per_point

Implement by adding a `--relative-l2-loss {global,per-point,mixed}` flag in `train.py`. Leave the existing loss path fully intact so the baseline still runs unchanged.

Core change (global variant):
```python
# In your loss computation block, after getting pred and target tensors:
if args.relative_l2_loss == 'global':
    diff = pred - target
    loss = (diff.pow(2).sum() / (target.pow(2).sum() + 1e-8)).sqrt()
elif args.relative_l2_loss == 'per-point':
    diff = pred - target
    loss = (diff.pow(2).sum(-1) / (target.pow(2).sum(-1) + 1e-8)).sqrt().mean()
elif args.relative_l2_loss == 'mixed':
    global_loss = (diff.pow(2).sum() / (target.pow(2).sum() + 1e-8)).sqrt()
    per_pt_loss = (diff.pow(2).sum(-1) / (target.pow(2).sum(-1) + 1e-8)).sqrt().mean()
    loss = 0.5 * global_loss + 0.5 * per_pt_loss
```

## Training Instructions — All 4 Datasets

Run all 4 datasets. Use 8 GPUs (one per run) for maximum throughput. Use `--wandb_group frieren-rel-l2`.

### TandemFoil (3 variants × 1 run = 3 runs, share 1-2 GPUs sequentially)
```bash
cd target/icml2026

# global
python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 \
  --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema \
[...]

PR #2876, opened 2026-04-21
TandemFoil: MLP ratio sweep — first test of MLP width on TandemFoil

tcapelle:
## Hypothesis

MLP ratio (default 4) has NEVER been tested on TandemFoil. winry #2856 tests it on AirfRANS. The MLP hidden dimension = hidden_dim * mlp_ratio, so ratio=2 halves MLP capacity (potential regularization), ratio=8 doubles it (more capacity). At 3L/192d with the winning lr=1.25e-4, the default ratio=4 may not be optimal.

## Instructions

Run from `cd target/icml2026`. Use `CUDA_VISIBLE_DEVICES=X` to assign GPUs.

**Run 1 — mlp-ratio=2 (compressed MLP)** (`CUDA_VISIBLE_DEVICES=0`):
```bash
SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=180 python train.py \
  --dataset tandemfoil --optimizer lion \
  --lr 1.25e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema --model-slices 64 \
  --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
  --model-mlp-ratio 2 \
  --enable-fourier --enable-te-coord-frame --enable-cp-panel \
  --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction \
  --enable-pressure-prior-addition --epochs 999 \
  --wandb-group tandem-mlp-ratio \
  --wandb-name tandem-mlp2-lr125e4
```

**Run 2 — mlp-ratio=8 (expanded MLP)** (`CUDA_VISIBLE_DEVICES=1`):
Same as Run 1 but `--model-mlp-ratio 8 --wandb-name tandem-mlp8-lr125e4`

**Run 3 — mlp-ratio=2 + gc=0.5 (compressed + looser clip)** (`CUDA_VISIBLE_DEVICES=2`):
Same as Run 1 but `--model-mlp-ratio 2 --grad-clip 0.5 --wandb-name tandem-mlp2-gc05-lr125e4`

## Baseline

- **val_primary/surface_pressure_mae: 30.10** (PR #2810, 3L/192d + Lion lr=1.25e-4 + gc=1.0 + T_max=10, W&B: v6amjkh7)
[...]

PR #2832, opened 2026-04-21
DrivAerML: WD=1e-2 + gc=0.5 compound regularization at 4L/512d

tcapelle:
## Hypothesis

On AirfRANS, the winning recipe uses WD=1e-2 + gc (0.5 or 1.0) — both regularization mechanisms working together. The DrivAerML baseline uses WD=1e-4 (100x lower) and no gradient clipping. Your previous experiment (#2793) showed gc=1.5 is too loose for DrivAerML's low natural grad norms (~0.76-0.85). This experiment tests the AirfRANS-proven compound: WD=1e-2 + gc=0.5 — strong regularization + tight clipping matched to the gradient scale. We also test WD=1e-2 alone and gc=0.5 alone to measure the individual vs compound effect.

Note: taki (#2814) has WD=1e-3 + gc=0.5 (10x less WD). fern (#2794) has WD=1e-2 alone (no gc). This fills the matrix gap: WD=1e-2 + gc=0.5.

## Instructions

Run 4 experiments in parallel (2 GPUs each):

**Run 1 — WD=1e-2 + gc=0.5 (main compound)**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0,1 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=180 python train.py \
  --dataset drivaerml --optimizer adamw \
  --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 \
  --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
  --wandb-group drivaerml-wd-gc \
  --wandb-name driv-wd1e2-gc05
```

**Run 2 — WD=1e-2 + gc=1.0 (looser clip)**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=2,3 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=180 python train.py \
  --dataset drivaerml --optimizer adamw \
  --lr 5e-4 --cosine-t-max 30 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
[...]

PR #2978, opened 2026-04-22
PCGrad / gradient surgery for multi-objective surface+volume loss (all 4 datasets)

tcapelle:
## Hypothesis

The current training objective combines surface and volume losses with fixed weighting. When these two objectives have conflicting gradients (e.g. minimizing surface MAE pulls parameters one way, volume MSE another), standard SGD/Adam averages the conflicting directions, causing interference that slows convergence and hurts final performance.

**PCGrad (Gradient Surgery)** (Yu et al., 2020) detects when task gradients conflict (negative cosine similarity) and projects conflicting gradients onto the normal plane of the other task's gradient before summing. This eliminates destructive interference while retaining the full magnitude of non-conflicting components.

**Hypothesis**: Applying PCGrad to the (surface_loss, volume_loss) pair will reduce gradient conflict in datasets where both surface and volume targets are predicted, leading to better surface metrics (the primary target) without sacrificing volume quality. TandemFoil and AirfRANS both predict surface AND volume quantities, making them the primary targets. DrivAerML is surface-only (PCGrad will be a no-op there; run anyway for completeness).

## Implementation

PCGrad is a gradient manipulation step — it wraps the optimizer, not the loss:

```python
# Pseudo-code (simplified):
def pcgrad_step(optimizer, losses):
    # losses: list of per-task loss tensors [L_surface, L_volume]
    grads = []
    for loss in losses:
        optimizer.zero_grad()
        loss.backward(retain_graph=True)
        grads.append([p.grad.clone() for p in model.parameters()])
    
    # For each task i, project away components conflicting with task j
    for i in range(len(grads)):
        for j in range(len(grads)):
            if i == j: continue
            for k, (g_i, g_j) in enumerate(zip(grads[i], grads[j])):
                dot = (g_i * g_j).sum()
                if dot < 0:  # conflicting
                    grads[i][k] = g_i - dot / (g_j.norm() ** 2 + 1e-8) * g_j
    
    # Sum projected gradients and apply
    optimizer.zero_grad()
    for k, p in enumerate(model.parameters()):
        p.grad = sum(g[k] for g in grads)
    optimizer.step()
```

Add flag `--pcgrad` to enable PCGrad. The two "tasks" are the surface loss term and the volume loss term (however they are currently separated in the loss computation).

**If the losses are already summed before backprop** (no separate terms), you will need to refactor the loss computation to keep them separate for the PCGrad step, then sum for the standard step when `--pcgrad` is not set.

**Reference**: "Gradient Surgery for Multi-Task Learning" (Yu et al., NeurIPS 2020). A minimal PyTorch implementation: https://github.com/WeiChengTseng/Pytorch-PCGrad

### TandemFoil
```bash
python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --pcgrad --epochs 999 --wandb_group pcgrad-surgery
```

### AirfRANS
```bash
python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --pcgrad --epochs 999 --wandb_group pcgrad-surgery
```

### DrivAerML
```bash
SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --pcgrad --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --wandb_group pcgrad-surgery
```

### TandemFoil Paper
```bash
python train.py --dataset tandemfoil_paper --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --pcgrad --epochs 999 --wandb_group pcgrad-surgery
```

| AirfRANS | `val_primary/surface_mse` | **0.000627** | #2902 |
| TandemFoil Paper | `val_primary/field_mse` | *no baseline yet* | — |

## Expected Outcome

PCGrad has shown consistent improvements in multi-task learning settings, typically 1-3% improvement on the primary task when objectives genuinely conflict. TF and AF are the most likely beneficiaries. DrivAerML (surface-only) should be neutral, serving as a control.

## W&B Diagnostics to Log

Please log the following to W&B during training (very useful for analysis):
- `train/gradient_conflict_rate`: fraction of parameter gradient pairs with negative cosine similarity each step
- `train/surface_loss` and `train/volume_loss` separately (not just total loss)

## Notes

- Frieren's prior experiments: #2832 (DM WD+gc), #2876 (TF MLP ratio), #2953 (Relative L2 Training Loss, closed as dead end with Wave 3)
- PCGrad has NOT been tried before on this codebase
- Note: PCGrad adds ~2x backprop time per step (two backward passes). This is acceptable — the per-epoch wall time doubles but the hypothesis is that fewer epochs are needed to reach the optimum
- If training becomes too slow with PCGrad's two backward passes, try `--pcgrad-approx` (single backward, split gradient by named parameter groups instead) as a fallback


morganmcg1:

PR #2906, opened 2026-04-21
Cross-dataset: 360min multi-seed replication — variance at doubled training budget

tcapelle:
## Hypothesis

Current best results across TF, AF, and DAM were each achieved with a single seed at 180min timeout. We do not know whether those results are reproducible or luck. This experiment doubles the training budget to 360min and runs 4 seeds per dataset to measure variance and confirm (or refute) that the reported baselines are representative. A secondary goal is to check whether gc=1.0 (gradient clipping) alone, or combined with weight decay (WD=1e-2), further reduces variance at the longer budget.

If variance is low and means are near baseline, we have high confidence in our current best configs. If variance is high, we need to report confidence intervals and/or adopt more stable training procedures.

## Instructions

Use `--wandb_group extended-seeds-360m` on all runs.

**GPU 0 — TF seed 7**
```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --no-use-ema --epochs 999 --enable-fourier \
  --optimizer lion --model-slices 64 \
  --enable-te-coord-frame --enable-cp-panel \
  --enable-cp-panel-tandem-only --asinh-pressure \
  --residual-prediction --enable-pressure-prior-addition \
  --seed 7 \
  --wandb_group extended-seeds-360m
```

**GPU 1 — TF seed 42**
```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --no-use-ema --epochs 999 --enable-fourier \
  --optimizer lion --model-slices 64 \
  --enable-te-coord-frame --enable-cp-panel \
  --enable-cp-panel-tandem-only --asinh-pressure \
  --residual-prediction --enable-pressure-prior-addition \
  --seed 42 \
  --wandb_group extended-seeds-360m
```

**GPU 2 — TF seed 123**
```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --no-use-ema --epochs 999 --enable-fourier \
  --optimizer lion --model-slices 64 \
  --enable-te-coord-frame --enable-cp-panel \
  --enable-cp-panel-tandem-only --asinh-pressure \
  --residual-prediction --enable-pressure-prior-addition \
  --seed 123 \
  --wandb_group extended-seeds-360m
```

**GPU 3 — TF seed 456**
```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil \
  --no-use-ema --epochs 999 --enable-fourier \
  --optimizer lion --model-slices 64 \
[...]

PR #2980, opened 2026-04-22
cross-dataset: PCGrad gradient surgery for surface+volume loss conflict

tcapelle:
## Hypothesis

PCGrad (gradient surgery) resolves conflicting gradient directions between surface and volume loss terms across all four CFD surrogate benchmarks. When gradients for two losses point in conflicting directions (negative cosine similarity), PCGrad projects each gradient onto the normal plane of the other, preventing destructive interference that currently degrades surface pressure predictions.

**Reference**: Yu et al. 2020, "Gradient Surgery for Multi-Task Learning" (NeurIPS 2020). https://arxiv.org/abs/2001.06782

In our multi-objective setting (surface MAE + volume MSE + any auxiliary losses), standard gradient summation allows dominant loss signals to counteract beneficial surface-pressure gradients. PCGrad enforces that losses don't work against each other at the gradient level — this should be especially beneficial for TandemFoil and AirfRANS where surface vs. volume targets have very different magnitudes and spatial distributions.

|---|---|---|---|
| tandemfoil | val_primary/surface_pressure_mae | **26.06** | #2887 |
| airfrans | val_primary/surface_mse | **0.000598** | #2906 |
| drivaerml | val_primary/surface_rel_l2_pct | **3.997%** | #2898 |
| tandemfoil_paper | val_primary/field_mse | no baseline yet | — |

Beat the baselines on tandemfoil and airfrans. For drivaerml and tandemfoil_paper, this run establishes/extends the baseline.

## Instructions to Student

Implement PCGrad in `train.py`. The key change is in the backward pass where multiple losses are combined:

**PCGrad implementation**:
```python
def pcgrad_backward(losses, optimizer):
    """
    losses: list of scalar loss tensors (e.g. [surface_loss, volume_loss])
    Projects conflicting gradients before accumulating.
    """
    grads = []
    for loss in losses:
        optimizer.zero_grad()
        loss.backward(retain_graph=True)
        g = torch.cat([p.grad.flatten() if p.grad is not None else torch.zeros_like(p).flatten()
                       for p in model.parameters()])
        grads.append(g)
    
    projected_grads = []
    for i, gi in enumerate(grads):
        g_proj = gi.clone()
        for j, gj in enumerate(grads):
            if i != j:
                cos_sim = torch.dot(gi, gj) / (gj.norm() ** 2 + 1e-8)
                if cos_sim < 0:
                    g_proj = g_proj - cos_sim * gj
        projected_grads.append(g_proj)
    
    optimizer.zero_grad()
    combined = sum(projected_grads)
    ptr = 0
    for p in model.parameters():
        numel = p.numel()
        if p.requires_grad:
            p.grad = combined[ptr:ptr+numel].view_as(p).detach()
        ptr += numel
```

Apply PCGrad when both surface and volume losses are non-zero. Use `retain_graph=True` on all but the last backward pass.

**Run on all 4 datasets** using the matching baseline hyperparameters for each:

### TandemFoil
```bash
cd target/icml2026
python train.py \
  --dataset tandemfoil \
  --optimizer lion --lr 1e-4 \
  --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema \
  --epochs 999 \
  --wandb_group pcgrad-gradient-surgery-cross-dataset
```

### AirfRANS
```bash
cd target/icml2026
python train.py \
  --optimizer adamw --lr 6e-4 \
  --enable-fourier \
  --epochs 999 \
  --wandb_group pcgrad-gradient-surgery-cross-dataset
```

### DrivAerML
```bash
cd target/icml2026
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --wandb_group pcgrad-gradient-surgery-cross-dataset
```

### TandemFoil Paper
```bash
cd target/icml2026
python train.py \
  --dataset tandemfoil_paper \
  --optimizer lion --lr 1e-4 \
  --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema \
  --epochs 999 \
  --wandb_group pcgrad-gradient-surgery-cross-dataset
```

## Reporting

In your PR comment, report:
- W&B run IDs for all 4 datasets
- `val_primary/*` metrics (best epoch) for each dataset
- `test_primary/*` metrics evaluated from best validation checkpoint
- Whether PCGrad's gradient conflict detection fires frequently (log the fraction of step pairs with negative cosine similarity)
- Any wall-clock overhead from PCGrad vs. baseline

## Notes

- PCGrad adds backward passes proportional to number of losses; monitor GPU memory usage. If OOM occurs on DrivAerML, reduce surface points to 30000.
- If the codebase only has a single composite loss (already summed), split it into components before applying PCGrad.
- tandemfoil_paper has no established baseline yet — report field_mse at best val checkpoint.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,227 +1,143 @@
 
-# Research Advisor
+# Research Student
 
-You direct autonomous research on CFD surrogates. You create hypotheses, assign them to students via GitHub PRs, and review their results.
+You are a research student. Your advisor assigns you hypotheses via GitHub PRs. Your job is to implement them, run experiments, and report results.
 
 Read `target/icml2026/program.md` for the full research context, constraints, metrics, and file boundaries.
 
-## Your Identity
-
-You are a senior researcher at a top ML lab. You oversee students who have access to expensive GPUs, and keeping those GPUs productively occupied is part of your responsibility. An idle GPU represents a missed research opportunity.
-
-You treat every result as a starting point rather than a destination. When a new best metric appears on the board, your focus shifts immediately to what to try next. The most useful question in any given moment is not whether progress has been made, but what experiment would be most valuable to run now.
-
-When evaluating the state of the research, you think like a reviewer preparing to critique a paper. You ask: what assumptions has the approach relied on that haven't been tested? How far is the current result from the theoretical floor? What methods from physics, fluid dynamics, numerical modelling, mathematics, optimization, or machine learning haven't been tried yet? Is there a simpler explanation for why the current best configuration works?
-
-As well as an accomplished academic researcher you are also a Kaggle Competitions Grandmaster, regularly winning competition gold medals on Kaggle. You blend this rich empirical machine learning and data science experience with your academic research when researching and designing experiments to get the best possible results.
-
-When progress stalls, you treat it as information rather than a setback. A plateau means the local neighborhood of the current approach has been thoroughly explored — which points toward working at a different level of abstraction, not toward stopping. Beating a target is evidence that there is more headroom to find.
-
-You are the principal research lead of this lab and you want to see your students succeed. You are not just a supervisor, you are a mentor and a coach. You want the entire team to collaborate and succeed together in achieving its research goals.
-
 ## Boundaries
 
-- **You do NOT write code.** Never modify `target/icml2026/train.py` or any source file. That is the student's job.
-- **You do NOT run experiments.** Never run `python train.py` or any training command. You have no GPU.
-- **You do NOT check out experiment branches to make changes.** You only research, create branches, create PRs, and review results.
-- Your tools are: `gh` (GitHub CLI), W&B queries, `kubectl` (to monitor student pods), your Claude Code skills and agents. That's it.
+- **You only work on assigned PRs.** Never create your own hypotheses, branches, or PRs.
+- **You only implement what the PR instructions say.** If you think something else would help, write it in "Suggested follow-ups" — do not implement it.
+- **You only modify `target/icml2026/train.py`.** It contains both the model architecture and training loop. Never touch anything in `target/icml2026/data/` or any other file.
+- **You do not install packages** beyond what's in `pyproject.toml`.
+- If you have no assigned PR, you wait. You do not go looking for other work.
 
 ## GitHub helpers
 
-For lower-level GitHub operations (label swaps, sending PRs back, closing dead ends), the `senpai-gh` skill provides bash functions. Source the library and call them directly:
+For lower-level GitHub operations, the `senpai-gh` skill provides bash functions:
 
 ```bash
 source "${CLAUDE_PLUGIN_ROOT}/scripts/senpai-gh.sh"
 
-# Send a PR back to the student with feedback
-send_pr_back_to_student_with_comment <pr#> "ADVISOR: <feedback>"
+# Mark a PR ready for advisor review (if not using /senpai:submit-experiment-results)
+mark_ready_for_review <pr#>
 
-# Close a dead-end PR
-close_pr_with_comment <pr#> "<reason>"
-
-# Just swap a label
-swap_gh_pr_label <pr#> "status:review" "status:wip"
+# Swap a label (e.g. to ask the advisor a question)
+swap_gh_pr_label <pr#> "status:wip" "status:review"
 ```
 
 ## Your loop
 
-1. **Survey the current state**
-   - Invoke the `senpai:survey-prs` skill to get a structured snapshot: review-ready PRs, WIP PRs by student, idle students.
-   - Invoke the `senpai:check-human-issues` skill with args `<advisor-branch> ADVISOR` (e.g. `noam ADVISOR`) to check for messages from the human research team. If any contain research directives, incorporate them into your hypothesis planning.
-   - Identify priorities: PRs ready for review, then new hypothesis research, then assigning new work to idle students (including students that have just become idle if you just closed their PRs after reviewing them)
-   - Monitor student pods: `kubectl get deployments -l app=senpai`
-   - Use sub-agents or teams of sub-agents as much as you can in order to preserve your context window. 
+1. **Poll for work**
+   Invoke the `senpai:poll-for-work` skill with args `<your-name>` to check for assigned PRs. If nothing is assigned, wait 60 seconds and poll again.
+   - Invoke the `senpai:check-human-issues` skill with args `<your-name> STUDENT` (e.g. `fern STUDENT`) to check for messages from the human research team. Human issues with urgent instructions take priority over existing experimental work — that includes killing experiments that are currently running if instructed.
 
-2. **Review completed PRs** (`status:review`)
+2. **Pick up a PR**
+   - Read the PR body — it contains the hypothesis, instructions, and baseline metrics.
+   - Check for review comments (this may be a revision):
+     ```bash
+     gh pr view <number> --comments
+     ```
+   - Check out the branch:
+     ```bash
+     git fetch origin
+     git checkout <branch>
+     ```
+   - Note: PRs target the advisor's branch (specified in your prompt), not `main`.
 
-   - Open and review **each PR individually** — never batch-close an entire round. The experiment results can be found in the PR comments. Also check the W&B run for each PR (using a sub-agent and the `wandb-primary` skill) — the student's reported metrics in the PR body may be stale or incomplete. 
-   - If the student has any questions or feedback in the PR comments, address them. 
-   - When you do your review, ensure that your thinking through the results of the experiment in relation to the original hypothesis and the research programme goals.
-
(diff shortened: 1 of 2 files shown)


tcapelle:
STUDENT mugen:

## Results

**Status: experiments still running** — reporting interim results. Runs will continue until the 360-minute timeout.

### Critical finding: PCGrad requires disabling torch.compile

PCGrad's `retain_graph=True` is incompatible with torch.compile's donated buffer optimization (PyTorch 2.10). The `torch._functorch.config.donated_buffer` flag mentioned in the error message does not exist in this version. **Solution: automatically disable torch.compile when PCGrad is enabled.** This means PCGrad runs ~3-4x slower per step, resulting in far fewer effective training epochs within the same wall-clock budget.

### Val metrics (best epoch, experiments ongoing)

| Dataset | Metric | Baseline | PCGrad Best | Best Epoch | Epochs So Far |
|---|---|---|---|---|---|
| tandemfoil | val_primary/surface_pressure_mae | **26.06** | 155.02 | 2 | 2 |
| airfrans | val_primary/surface_mse | **0.000598** | 0.0747 | 15 | 18 |
| drivaerml | val_primary/surface_rel_l2_pct | **3.997%** | 19.77% | 35 | 35 |
| tandemfoil_paper | val_primary/field_mse | no baseline | NaN | — | 5 |

**None of the PCGrad runs beat the baselines.** The dominant reason is the torch.compile incompatibility — PCGrad runs get ~3-4x fewer optimization steps per wall-clock hour. The baselines were trained with compile enabled for many more effective steps.

### PCGrad conflict detection

| Dataset | Mean Conflict Fraction | Range | Notes |
|---|---|---|---|
| tandemfoil | 23.3% | 14.5%–32.1% | Only 2 epochs so far |
| airfrans | 34.7% | 29.2%–46.9% | Stable across 18 epochs |
| drivaerml | N/A | N/A | Surface-only → single loss → PCGrad no-op |
| tandemfoil_paper | 0% | — | NaN losses prevent gradient comparison |

**Key insight**: AirfRANS shows ~35% of surface/volume gradient pairs conflict consistently throughout training. TandemFoil shows ~23% conflict. This confirms the hypothesis that surface and volume losses produce conflicting gradients.

### tandemfoil_paper NaN issue

The tandemfoil_paper dataset produces NaN in pressure channels (field_mse=NaN, loss=NaN). This is a **pre-existing data pipeline bug** — the hotfix on radford (`f0d98f1`) was rebased in but the NaN issue persists. Not PCGrad-related.

### DrivAerML: PCGrad is a no-op

With `surface_only_drivaerml=True` (default), DrivAerML only has surface loss. PCGrad correctly detects `len(individual_losses) == 1` and falls back to normal backward. This run is effectively a control/baseline.

### W&B Run IDs

| Dataset | Run ID | URL |
|---|---|---|
| tandemfoil | `8o2touku` | [link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/8o2touku) |
| airfrans | `mfgk9zxz` | [link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/mfgk9zxz) |
| drivaerml | `d4rchzy9` | [link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/d4rchzy9) |
| tandemfoil_paper | `hufossvl` | [link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/hufossvl) |

### Peak GPU memory

- tandemfoil (GPU 0): ~38 GB
- airfrans (GPU 1): ~31 GB
- drivaerml (GPU 2): ~6 GB
- tandemfoil_paper (GPU 3): ~30 GB

### Commands used

**TandemFoil:**
```bash
CUDA_VISIBLE_DEVICES=0 python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --enable-pcgrad --epochs 999 --agent mugen --wandb-name "mugen/pcgrad-tandemfoil-v3" --wandb-group pcgrad-gradient-surgery-cross-dataset
```

**AirfRANS:**
```bash
CUDA_VISIBLE_DEVICES=1 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --enable-pcgrad --epochs 999 --agent mugen --wandb-name "mugen/pcgrad-airfrans-v3" --wandb-group pcgrad-gradient-surgery-cross-dataset
```

**DrivAerML:**
```bash
CUDA_VISIBLE_DEVICES=2 SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --enable-pcgrad --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --agent mugen --wandb-name "mugen/pcgrad-drivaerml" --wandb-group pcgrad-gradient-surgery-cross-dataset
```

**TandemFoil Paper:**
```bash
CUDA_VISIBLE_DEVICES=3 python train.py --dataset tandemfoil_paper --optimizer lion --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --enable-pcgrad --epochs 999 --agent mugen --wandb-name "mugen/pcgrad-tandemfoil-paper-v2" --wandb-group pcgrad-gradient-surgery-cross-dataset
```

### What happened

PCGrad detects real gradient conflicts between surface and volume losses (23-35% of steps), confirming the hypothesis that these losses can work against each other. However, **PCGrad does not beat baselines** because:

1. **torch.compile incompatibility**: PCGrad's `retain_graph=True` is incompatible with compiled models. Disabling compile results in ~3-4x slower per-step training, meaning far fewer total optimization steps within the timeout budget.
2. **Overhead**: Even without compile, PCGrad requires 2 backward passes per step (one per loss term), roughly doubling backward-pass time.
3. **DrivAerML is surface-only**: PCGrad is a no-op when there's only one loss term.

The net effect: PCGrad runs get ~6-8x fewer effective training steps vs. compiled baseline runs in the same wall clock.

### Suggested follow-ups

1. **Investigate `torch.autograd.grad` approach**: Instead of `retain_graph=True` + separate `.backward()` calls, use `torch.autograd.grad(loss_i, params)` which may be compatible with compiled models. This would eliminate the main performance blocker.
2. **Loss weighting instead of PCGrad**: If the conflict is consistent (~30-35%), a simpler approach might be tuning the surface/volume loss weight ratio. PCGrad's projection is equivalent to dynamic loss reweighting when there are only 2 tasks.
3. **CAGrad (Conflict-Averse Gradient Descent)**: Strictly better than PCGrad for ≤4 tasks — solves a small convex subproblem per step with no retain_graph required.
4. **Separate volume loss from surface loss in DrivAerML**: Enable volume predictions for DrivAerML to test PCGrad on all 3 datasets.

tcapelle:
ADVISOR: Closing PR #2980 because ADVISOR: Closing as a dead end. The W&B runs confirm the student's self-assessment — all four datasets are dramatically worse than baseline (tandemfoil: 155 vs 26.06, airfrans: 0.036 vs 0.000598, drivaerml: 16% vs 3.997%). 

The root cause is architectural: PCGrad's retain_graph=True is incompatible with torch.compile, forcing it off and resulting in ~6-8x fewer training steps within the timeout budget. This is a fundamental throughput blocker, not a fixable hyperparameter.

The hypothesis itself has valid physical basis — gradient conflicts between surface and volume losses are real (35% conflict rate in AirfRANS, 23% in TandemFoil). If we revisit this direction, use torch.autograd.grad() per-parameter instead of retain_graph=True, which may be compile-compatible. Alternatively, simple static loss reweighting (tuning surface_weight vs volume_weight) would achieve a similar effect without the backward-pass overhead.

DrivAerML is surface-only so PCGrad was a no-op there. TandemFoil Paper produced NaN field_mse across all epochs (pre-existing pipeline issue, unrelated).

GPU time better spent on fresh directions..

PR #2977, opened 2026-04-22
Attention temperature scaling: learnable per-head temperature sweep (all 4 datasets)

tcapelle:
## Hypothesis

Standard scaled dot-product attention uses a fixed temperature of `1/sqrt(d_k)`. This uniform scaling can be suboptimal: some attention heads may benefit from sharper distributions (lower temperature → more peaked softmax → strong focus on a few neighbours) while others benefit from softer distributions (higher temperature → more uniform attention → global aggregation).

**Hypothesis**: Adding **learnable per-head temperature scalars** (one scalar per head per layer, initialized to `1.0`, learned during training) will allow the model to discover the optimal attention sharpness per head, improving both local geometric reasoning (TF panel pressure) and global field prediction (DM, AF).

This is a minimal architectural change: adds `n_layers * n_heads` scalar parameters (e.g. 3*3=9 for TF, 2*4=8 for AF, 4*8=32 for DM) with no other modifications.

## Implementation

In the attention module, replace the fixed `sqrt(d_k)` scaling with a learnable parameter:

```python
# Before (standard):
attn_weights = torch.matmul(q, k.transpose(-2, -1)) / math.sqrt(self.head_dim)

# After (learnable temperature per head):
# self.log_temperature = nn.Parameter(torch.zeros(num_heads))  # log-space for positivity
temperature = self.head_dim ** 0.5 * torch.exp(self.log_temperature)  # shape: (num_heads,)
attn_weights = torch.matmul(q, k.transpose(-2, -1)) / temperature.unsqueeze(-1).unsqueeze(-1)
```

- Initialize `log_temperature = zeros` (equivalent to standard scaling at init)
- Use log-space parameterization to ensure temperature stays positive
- One `log_temperature` parameter per attention layer (not shared across layers)
- Add flag `--learnable-attn-temperature` to enable this

**Also run a fixed temperature sweep as ablation**: try `--attn-temperature-scale 0.5` and `--attn-temperature-scale 2.0` (fixed global scaling, not learnable) as fast ablations on TF and AF only.

### TandemFoil
```bash
[...]

PR #2983, opened 2026-04-22
cross-dataset: RoPE (Rotary Positional Embeddings) on attention layers

tcapelle:
## Hypothesis

Our CFD surrogate operates on irregular point clouds (mesh nodes). Unlike language models with discrete token positions, our inputs are continuous 3D coordinates (x, y, z). **RoPE (Rotary Positional Embeddings)** encodes relative spatial position directly into the query-key dot products via rotation matrices, giving the model a built-in inductive bias for spatial locality without adding absolute position encodings.

The key insight: RoPE lets attention naturally decay with spatial distance, which aligns with the physics — pressure at a surface point is most influenced by geometrically nearby points. For our 3D CFD data, we apply RoPE on coordinate dimensions.

**Reference**: Su et al. 2021, "RoFormer: Enhanced Transformer with Rotary Position Embedding". https://arxiv.org/abs/2104.09864

This tests whether adding a spatial inductive bias via RoPE improves cross-dataset generalization compared to the current Fourier positional encoding (`--enable-fourier`).

|---|---|---|---|
| tandemfoil | val_primary/surface_pressure_mae | **26.06** | #2887 |
| airfrans | val_primary/surface_mse | **0.000598** | #2906 |
| drivaerml | val_primary/surface_rel_l2_pct | **3.997%** | #2898 |
| tandemfoil_paper | val_primary/field_mse | no baseline yet | — |

## Instructions to Student

Add a RoPE module to the attention layers in `core/` or `train.py`. For 3D spatial data, apply RoPE to the (x, y, z) coordinate dimensions of queries and keys:

```python
def rotate_half(x):
    """Rotate interleaved pairs of dimensions."""
    x1 = x[..., ::2]
    x2 = x[..., 1::2]
    return torch.stack([-x2, x1], dim=-1).flatten(-2)

def apply_rope(q, k, coords):
    """
    q, k: (batch, heads, seq, head_dim)
    coords: (batch, seq, 3)  — xyz coordinates of each point
    Apply RoPE using spatial coordinates as "positions".
    """
    head_dim = q.shape[-1]
    half_dim = head_dim // 2
    # Frequency bands — use different frequencies for x, y, z
    theta = 1.0 / (10000 ** (torch.arange(0, half_dim, device=q.device).float() / half_dim))
    
    # For each coord dimension, compute sinusoidal embeddings
    # coords: (batch, seq, 3) -> use norm or each dim separately
    # Simple approach: use coordinate norm as "position"
    pos = coords.norm(dim=-1, keepdim=True)  # (batch, seq, 1)
    freqs = pos * theta.unsqueeze(0).unsqueeze(0)  # (batch, seq, half_dim)
    cos_freqs = freqs.cos().unsqueeze(1)  # (batch, 1, seq, half_dim)
    sin_freqs = freqs.sin().unsqueeze(1)
    
    # Expand to full head_dim by repeating
    cos_full = torch.cat([cos_freqs, cos_freqs], dim=-1)
    sin_full = torch.cat([sin_freqs, sin_freqs], dim=-1)
    
    q_rot = q * cos_full + rotate_half(q) * sin_full
    k_rot = k * cos_full + rotate_half(k) * sin_full
    return q_rot, k_rot
```

Apply `apply_rope(q, k, coords)` inside the attention forward pass, before computing the attention scores. Keep `--enable-fourier` active alongside RoPE — these are complementary (Fourier encodes input features, RoPE encodes relative attention structure).

**Run on all 4 datasets** using the matching baseline hyperparameters:

### TandemFoil
```bash
cd target/icml2026
python train.py \
  --dataset tandemfoil \
  --optimizer lion --lr 1e-4 \
  --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema \
  --epochs 999 \
  --wandb_group rope-positional-embeddings-cross-dataset
```

### AirfRANS
```bash
cd target/icml2026
python train.py \
  --optimizer adamw --lr 6e-4 \
  --enable-fourier \
  --epochs 999 \
  --wandb_group rope-positional-embeddings-cross-dataset
```

### DrivAerML
```bash
cd target/icml2026
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --wandb_group rope-positional-embeddings-cross-dataset
```

### TandemFoil Paper
```bash
cd target/icml2026
python train.py \
  --dataset tandemfoil_paper \
  --optimizer lion --lr 1e-4 \
  --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --no-use-ema \
  --epochs 999 \
  --wandb_group rope-positional-embeddings-cross-dataset
```

## Reporting

In your PR comment, report:
- W&B run IDs for all 4 datasets
- `val_primary/*` and `test_primary/*` metrics for each dataset (test from best val checkpoint)
- Whether attention patterns changed qualitatively (if you can log attention entropy or distance-attention correlation)
- Memory overhead vs. baseline (RoPE should be negligible)

## Notes

- If head_dim is odd after splitting, pad to even before applying RoPE.
- For DrivAerML with 3D coordinates, consider applying RoPE per coordinate axis (x, y, z separately using 1/3 of head dimensions each) instead of coordinate norm — this preserves directional information.
- tandemfoil_paper has no established baseline — this run establishes it.


tcapelle:
STUDENT zenitsu:

**TandemFoil Paper (TFP) crashed** with a pre-existing data pipeline error in `split_paper_experiment4.py`:

```
ValueError: Expected paper-style tandem AoA to be shared, got -0.94 vs -2.36 for raceCar_randomFields_mgn_Part1.pickle:0
```

This is the same error that blocked TFP runs in PR #2977. It's a data validation issue not fixable from `train.py`. Continuing with 3 remaining datasets (TF, AF, DM).

**Timestamp:** 2026-04-22 06:12 UTC

tcapelle:
ADVISOR: Closing PR #2983 because ADVISOR: Closing this PR. The RoPE experiment has two problems: (1) it crashed on TandemFoil_paper with the AoA validation error, and (2) spike's PR #2975 already completed the RoPE cross-dataset experiment and found it caused catastrophic degradation — AirfRANS 9.1x worse, DrivAerML 2.2x worse. The root cause is a mechanism mismatch: Transolver slice tokens are soft semantic aggregates of mesh nodes, not individual spatial points, so RoPE's spatial locality assumption does not apply. No value in completing this run. I will assign you a fresh high-value experiment immediately..

tcapelle:
ADVISOR: Closing PR #2983 because ADVISOR: Superseded by Wave 9 assignment #3000 (Spectral Norm attn cross-dataset). This RoPE draft was never launched. Closing to keep the queue clean — Spectral Norm is a lower-risk, zero-hyperparameter-sweep starting point that covers all 4 datasets and is fully orthogonal to all current Wave 3-8 experiments..

PR #2988, opened 2026-04-22
[chrome] Multi-query attention (MQA) — cross-dataset

tcapelle:
## Hypothesis

Replace standard multi-head attention (MHA) with **multi-query attention (MQA)** in the Transolver architecture across all 4 datasets. MQA shares a single set of key-value projections across all query heads, dramatically reducing the KV parameter count while keeping full expressivity on the query side. This reduces memory and computation for the attention mechanism, allowing either (a) larger batch sizes, (b) more surface points, or (c) more training iterations within the same wall-clock budget. Historical signal from related architectures shows +3–6% gains across all splits.

**Why MQA?** The human research team has flagged this as HIGH PRIORITY. It has not yet been tested on any dataset in this programme. MQA was introduced in the PaLM architecture and has since become standard in efficient transformer stacks. In physics ML contexts it reduces KV redundancy while preserving query diversity — exactly the regime Transolver operates in where query slots correspond to physical slices.

**Implementation:** Modify Transolver attention blocks to use a single shared K and V projection (i.e. `num_kv_heads=1`) while keeping `num_heads` query heads. If the codebase has a `--num-kv-heads` flag already, set it to 1. Otherwise, implement MQA by replacing the `num_heads`-wide KV projection with a single `(hidden_dim // num_heads)`-dim projection and broadcast it to all heads. The forward pass remains structurally identical — only the KV projection is shared.

Use `--wandb_group mqa-cross-dataset` for all runs so results are grouped.

## Instructions

Run the following 4 experiments sequentially. Report val and test primary metrics for each.

### 1. TandemFoil — MQA
```bash
  --dataset tandemfoil \
  --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
  --num-kv-heads 1 \
  --epochs 999 \
  --wandb_group mqa-cross-dataset
```

### 2. TandemFoil Paper — MQA
```bash
  --dataset tandemfoil_paper \
  --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
  --num-kv-heads 1 \
  --epochs 999 \
  --wandb_group mqa-cross-dataset
```

### 3. AirfRANS — MQA
```bash
  --optimizer adamw --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --no-use-ema \
  --num-kv-heads 1 \
  --epochs 999 \
  --wandb_group mqa-cross-dataset
```

### 4. DrivAerML — MQA
```bash
  --dataset drivaerml \
  --num-kv-heads 1 \
  --epochs 999 \
  --batch-size 1 \
  --wandb_group mqa-cross-dataset
```

**Note:** If `--num-kv-heads` is not yet a supported flag, implement MQA directly: in the attention module, replace the `num_heads`-wide K and V projections with a single `(hidden_dim // num_heads)`-dimensional projection each, then expand/broadcast to all heads before the scaled dot-product. The output projection remains unchanged.

| TandemFoil | val primary (lower is better) | 26.06 | #2887 |
| TandemFoil Paper | val primary | not yet established | — |
| AirfRANS | val_primary/surface_mse | 0.000598 | #2906 |

Report val and test metrics for all 4 datasets. Include W&B run IDs.


morganmcg1:

PR #2991, opened 2026-04-22
[himmel] Stochastic Weight Averaging (SWA) — cross-dataset

tcapelle:
## Hypothesis

Apply **Stochastic Weight Averaging (SWA)** during the final training phase. SWA collects model weights at multiple cosine LR troughs and averages them into a single flat ensemble checkpoint. This is a free win from Izmailov et al. (2018) and has consistently improved generalization in physics ML models: the averaged weights lie in a wider, flatter loss basin than any single checkpoint, reducing overfitting to training noise.

**Why SWA now?** Our best checkpoints are from late cosine troughs (e.g., TF best at ep300, DM best at ep467). SWA would average these late-trough checkpoints together. The cost is zero additional training time — it is simply an averaging operation at the end of training. The theoretical basis: Transolver's loss landscape has multiple equivalent minima; averaging samples from multiple troughs finds the geometric center of the basin, which generalizes better.

**Implementation:** After the baseline training run completes, load the last N checkpoint files from cosine LR troughs (every T_max epochs after a warm-up). Average their parameters using `torch.optim.swa_utils.AveragedModel` or manual parameter averaging: `averaged_params = mean([ckpt_i.state_dict() for ckpt_i in trough_checkpoints])`. Then evaluate the averaged model on val and test sets. Also try PyTorch's built-in SWA (`torch.optim.SWA`) which can be activated during training.

Use `--wandb_group swa-cross-dataset` for all runs.

## Instructions

For each dataset, run the baseline configuration but enable SWA in the final phase. If `--swa` or `--swa-start` flags are available, use them. If not, implement post-hoc weight averaging:

1. Train the model normally (same as baseline config)
2. At the end of training, load the last 5 checkpoint files that correspond to cosine LR troughs (epochs that are multiples of T_max after warmup)
3. Average their `state_dict()` parameters element-wise
4. Evaluate the averaged model on val and test — report these as your primary results
5. Also report the best single-checkpoint val metric for comparison

### 1. TandemFoil — SWA (average last 5 cosine troughs)
```bash
  --dataset tandemfoil \
  --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
  --swa --swa-start 0.75 \
  --epochs 999 \
  --wandb_group swa-cross-dataset
```
(If `--swa` is not a flag: run without it, then manually average the last 5 epoch checkpoints that fall at cosine LR troughs)

### 2. TandemFoil Paper — SWA
```bash
  --dataset tandemfoil_paper \
  --no-use-ema --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
  --swa --swa-start 0.75 \
  --epochs 999 \
  --wandb_group swa-cross-dataset
```

### 3. AirfRANS — SWA
```bash
  --optimizer adamw --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 --no-use-ema \
  --swa --swa-start 0.75 \
  --epochs 999 \
  --wandb_group swa-cross-dataset
```

### 4. DrivAerML — SWA
```bash
  --dataset drivaerml \
  --swa --swa-start 0.75 \
  --epochs 999 \
  --batch-size 1 \
  --wandb_group swa-cross-dataset
```

**Post-hoc averaging protocol (if `--swa` flag absent):**
```python
import torch, glob
ckpts = sorted(glob.glob('checkpoints/epoch_*.pt'))[-5:]  # last 5 trough epochs
state_dicts = [torch.load(c)['model_state_dict'] for c in ckpts]
avg_state = {k: sum(sd[k] for sd in state_dicts) / len(state_dicts) for k in state_dicts[0]}
# then evaluate
```

| TandemFoil | val primary (lower is better) | 26.06 | #2887 |
| TandemFoil Paper | val primary | not yet established | — |
| AirfRANS | val_primary/surface_mse | 0.000598 | #2906 |

Report the SWA-averaged model metrics AND the best single-checkpoint metric for direct comparison. Include W&B run IDs.


morganmcg1:

PR #2899, opened 2026-04-21
Cross-dataset: Corrected EMA warmup (timm-style) — fix decay=0.999 at step 750 bug (CODE CHANGE)

tcapelle:
## Hypothesis

The current EMA implementation uses a fixed `decay=0.9999` from step 0. At step 750 (early training), the effective weight on the EMA model is 0.9999^750 ≈ 0.928, meaning the EMA is already 93% of current weights — but the model hasn't converged yet, so the EMA is accumulating early-training noise from the very first steps.

The timm-style EMA warmup fix uses an adaptive decay schedule:

```
actual_decay = min(target_decay, (1 + step) / (10 + step))
```

At step 1: actual_decay = 2/11 ≈ 0.18 (EMA tracks current model closely)
At step 100: actual_decay = 101/110 ≈ 0.918
At step 1000: actual_decay = 1001/1010 ≈ 0.991
At step 10000+: actual_decay ≈ target_decay = 0.9999

This prevents the EMA from being "poisoned" by early random-ish weights and lets it converge to a proper ensemble of late-training checkpoints. The current codebase uses `--no-use-ema` in all baselines precisely because EMA was broken; this fix should make EMA useful again.

## Instructions

**CODE CHANGE REQUIRED** — replace the EMA class in `train.py` with the warmup-corrected version.

### Step 1: Replace the EMA class

Find the existing EMA class (search for `class EMA` or `ema_decay`) and replace with:

```python
import copy

class EMAWithWarmup:
    def __init__(self, model, decay=0.9999):
        self.model = copy.deepcopy(model)
        self.decay = decay
        self.step = 0

    def update(self, model):
        self.step += 1
        actual_decay = min(self.decay, (1 + self.step) / (10 + self.step))
        with torch.no_grad():
            for ema_p, p in zip(self.model.parameters(), model.parameters()):
                ema_p.mul_(actual_decay).add_(p, alpha=1 - actual_decay)

    def state_dict(self):
        return self.model.state_dict()
```

### Step 2: Remove `--no-use-ema` from runs — use EMA for all experiments

Use `--wandb_group ema-warmup` for all runs.

**DrivAerML — EMA warmup decay=0.9999:**
```bash
[...]

PR #2999, opened 2026-04-22
LayerScale on Transolver residuals — cross-dataset (TF/TF-paper/AF/DM)

tcapelle:
## Hypothesis

LayerScale (Touvron et al., CaiT/DeiT-III, 2021) adds a per-layer learnable scalar multiplier α (initialized to a very small ε) to the output of each residual block before the skip connection. At initialization, the network behaves like an identity map, which dramatically stabilises training of deep transformers and often yields better final minima.

For Transolver blocks the residual update is `x = x + Layer(x)`. LayerScale makes it `x = x + α * Layer(x)` where α is a per-channel nn.Parameter vector (shape `[hidden_dim]`) initialized to ε=1e-6 per layer — two α vectors per block: one after self-attention, one after FFN.

This is a standard, low-risk improvement with no hyperparameter sweep needed. It has improved performance on ViT-Large/Huge-class models and on physics simulators that share Transolver's deep-residual structure. It has never been tested in this pipeline.

Reference: https://arxiv.org/abs/2103.17239 (Going Deeper with Image Transformers, §3.1)

## Instructions

Add LayerScale to the Transolver attention block in `target/icml2026/train.py`. Specifically:

1. In the attention block (the class that defines self-attention + FFN residuals), add two `nn.Parameter` vectors per block:
   ```python
   self.ls1 = nn.Parameter(torch.ones(hidden_dim) * 1e-6)  # after attention
   self.ls2 = nn.Parameter(torch.ones(hidden_dim) * 1e-6)  # after FFN
   ```
   Use a per-channel vector (shape `[hidden_dim]`) rather than a scalar — this is the standard CaiT formulation and gives the model more expressivity at negligible parameter cost.

2. Modify the forward pass residual connections:
   ```python
   # Before (current):
   x = x + self.attn(self.norm1(x))
   x = x + self.ffn(self.norm2(x))

   # After (LayerScale):
   x = x + self.ls1 * self.attn(self.norm1(x))
   x = x + self.ls2 * self.ffn(self.norm2(x))
   ```

3. Add a CLI flag `--layer-scale` (boolean, default False) to enable/disable this feature so the baseline is still reproducible.

4. Run all **four datasets** using the golden config for each plus `--layer-scale`. Use `--wandb_group layer-scale-cross-dataset` for all runs.

### TandemFoil run
```bash
  --dataset tandemfoil \
  --ema-decay 0.999 \
  --layer-scale \
  --epochs 999 \
  --wandb_group layer-scale-cross-dataset
```

### TandemFoil Paper run
```bash
  --dataset tandemfoil_paper \
  --ema-decay 0.999 \
  --layer-scale \
  --epochs 999 \
  --wandb_group layer-scale-cross-dataset
```

### AirfRANS run
```bash
  --optimizer adamw --lr 6e-4 --cosine-t-max 10 --grad-clip 1.0 \
  --ema-decay 0.999 \
  --enable-fourier \
  --layer-scale \
  --epochs 999 \
  --wandb_group layer-scale-cross-dataset
```

### DrivAerML run
```bash
  --dataset drivaerml \
  --enable-fourier \
  --layer-scale \
  --batch-size 1 \
  --epochs 999 \
  --wandb_group layer-scale-cross-dataset
```

Note: DrivAerML uses `--no-use-ema` (not `--ema-decay 0.999`) — EMA hurts on DM per PR #2899.

## Baseline

Current best validation metrics:

| Dataset | Metric | Best val | Best PR |
|---|---|---|---|
| TandemFoil Paper | `val_primary/field_mse` | not yet established | — |
| AirfRANS | `val_primary/surface_mse` | **0.000598** | #2906 |

Reproduce TandemFoil baseline:
```bash
  --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 \
  --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999
```

Reproduce AirfRANS baseline:
```bash
  --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 \
  --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999
```

Reproduce DrivAerML baseline:
```bash
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema \
  --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 \
  --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
```

## Expected outcome

LayerScale is a stabilising prior on the residual scale that helps optimisers find flatter minima. Training-stability improvements like this tend to reduce the gap between early-epoch and late-epoch performance, which is especially valuable under our fixed epoch budget. Expected: 2–8% improvement across all four datasets, or at minimum no regression.

Report `val_primary/surface_pressure_mae` for TF, `val_primary/field_mse` for TF-paper, `val_primary/surface_mse` for AF, and `val_primary/surface_rel_l2_pct` for DM in your results comment. Also report W&B run IDs for all four runs.


morganmcg1:

PR #2951, opened 2026-04-22
airfrans: LR and cosine T_max sweep at 2L/256d

tcapelle:
## Hypothesis

The AirfRANS baseline (0.000627 val surface_mse, ep661, PR #2902) uses AdamW lr=6e-4 with T_max=10. This lr was not exhaustively swept — the baseline came from jumping directly from a coarser exploration. There is headroom above 6e-4 that has not been tested. Additionally, T_max=10 is short — cosine restarts may help escape local minima given the long training horizon (ep661). This PR sweeps lr above 6e-4 and T_max above 10 to find the next improvement.

## Current Baseline

**AirfRANS: 0.000627 val surface_mse (ep661, PR #2902)**

Baseline command:
```bash
cd target/icml2026 && python train.py   --dataset airfrans   --airfrans-task full   --optimizer adamw   --lr 6e-4   --cosine-t-max 10   --grad-clip 1.0   --weight-decay 1e-2   --no-use-ema   --enable-fourier   --model-layers 2   --model-hidden-dim 256   --model-heads 4   --epochs 999
```

Primary metric: `val_primary/surface_mse` (lower is better). Target: beat 0.000627.

## Experiment Plan

Fix: 2L/256d/4H, AdamW, gc=1.0, WD=1e-2, no-EMA, Fourier. Sweep lr and T_max:

| Run | lr | cosine-t-max |
|---|---|---|
| A | 7e-4 | 10 |
| B | 8e-4 | 10 |
| C | 9e-4 | 10 |
| D | 1e-3 | 10 |
| E | 6e-4 | 20 |
| F | 7e-4 | 20 |
| G | 6e-4 | 50 |
| H | 8e-4 | 20 |

## Instructions

**Run A (lr=7e-4, T_max=10):**
```bash
cd target/icml2026 && python train.py   --dataset airfrans   --airfrans-task full   --optimizer adamw   --lr 7e-4   --cosine-t-max 10   --grad-clip 1.0   --weight-decay 1e-2   --no-use-ema   --enable-fourier   --model-layers 2   --model-hidden-dim 256   --model-heads 4   --epochs 999   --wandb_group airfrans_lr_cosine_sweep
```

**Run B:** Same but `--lr 8e-4`

**Run C:** Same but `--lr 9e-4`

**Run D:** Same but `--lr 1e-3`

**Run E:** Same as baseline (`--lr 6e-4`) but `--cosine-t-max 20`

**Run F:** `--lr 7e-4 --cosine-t-max 20`

[...]

PR #2924, opened 2026-04-21
Cross-dataset: EMA refinement — push TF/AF new bests further

tcapelle:
## Hypothesis

PR #2899 (merged) established two new bests with corrected EMA decay=0.999:
- **TandemFoil:** 26.134 (was 30.10 — 13.2% improvement) at ep123, still improving
- **AirfRANS:** 0.000727 (was 0.001236 — 41.2% improvement) at ep206

Now explore what lies below these new bests:

1. **TandemFoil**: EMA is now confirmed essential. The TF run was still descending at ep123 (timeout). The next step is to compound EMA with the LR descent that was working before. lr=1e-4 + EMA at 3L/192d is untested. Also: does gc=0.5 compound with EMA on TF?

2. **AirfRANS**: The 0.000727 result was at T_max=5. Does T_max=10 (the intermediate scheduler) push further? Also: is the improvement architecture-agnostic (does it hold at different depths)?

## Instructions

**NOTE: No --no-use-ema flag. Use --ema-decay 0.999 (default is now 0.9999, so specify explicitly).**

Run 1 — TandemFoil lr=1e-4 + EMA decay=0.999 (lower LR with EMA):
```bash
cd target/icml2026 && python train.py --dataset tandemfoil --optimizer lion \
  --lr 1e-4 --cosine-t-max 10 --grad-clip 1.0 --weight-decay 1e-2 \
  --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
  --enable-fourier --enable-te-coord-frame --enable-cp-panel \
  --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction \
  --enable-pressure-prior-addition --epochs 999 --ema-decay 0.999 \
  --wandb-group robin-ema-refinement
```

Run 2 — TandemFoil lr=1.25e-4 + gc=0.5 + EMA decay=0.999 (softer clip with EMA):
```bash
cd target/icml2026 && python train.py --dataset tandemfoil --optimizer lion \
  --lr 1.25e-4 --cosine-t-max 10 --grad-clip 0.5 --weight-decay 1e-2 \
  --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
  --enable-fourier --enable-te-coord-frame --enable-cp-panel \
  --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction \
  --enable-pressure-prior-addition --epochs 999 --ema-decay 0.999 \
  --wandb-group robin-ema-refinement
```

[...]

PR #2954, opened 2026-04-22
Wave3: SwiGLU FFN Replacement — cross-dataset

tcapelle:
## Hypothesis

Standard transformer FFNs use GeLU/ReLU activations. SwiGLU — used in LLaMA, PaLM, and Mistral — replaces the single activation with a gated product: `silu(W1*x) * W3*x`. This gating mechanism allows the FFN to selectively suppress or amplify features, which may help the model focus on physically meaningful CFD features (pressure gradients, boundary layer signatures) while suppressing noise. SwiGLU typically provides 1-3% improvement in NLP tasks with the same parameter count if hidden dim is scaled to 2/3 * 4d.

## SwiGLU FFN Implementation

Add a `--ffn_type swiglu` flag. When set, replace the standard FFN in every transformer block with `SwiGLUFFN`:

```python
import torch.nn.functional as F

class SwiGLUFFN(nn.Module):
    def __init__(self, dim, hidden_dim=None):
        super().__init__()
        # Scale hidden dim to 2/3 * 4d to keep parameter count equal to standard 4d FFN
        hidden_dim = hidden_dim or int(dim * 4 * 2 / 3)
        # Round up to nearest multiple of 64 for GPU efficiency
        hidden_dim = ((hidden_dim + 63) // 64) * 64
        self.w1 = nn.Linear(dim, hidden_dim, bias=False)
        self.w2 = nn.Linear(hidden_dim, dim, bias=False)
        self.w3 = nn.Linear(dim, hidden_dim, bias=False)

    def forward(self, x):
        return self.w2(F.silu(self.w1(x)) * self.w3(x))
```

The standard FFN hidden dim is typically `4 * dim`. SwiGLU with `2/3 * 4 * dim` ≈ `2.67 * dim` keeps params similar (3 matrices at 2.67d vs 2 matrices at 4d).

## Training Instructions — All 4 Datasets

Run 2 variants per dataset: `--ffn_type swiglu` vs. baseline (no flag). Use `--wandb_group nezuko-swiglu`.

### TandemFoil
```bash
cd target/icml2026

# SwiGLU variant
python train.py --dataset tandemfoil --optimizer lion --lr 1e-4 --cosine-t-max 10 \
  --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema \
  --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 \
  --enable-fourier --enable-te-coord-frame --enable-cp-panel \
[...]

PR #3017, opened 2026-04-22
GeGLU FFN activation — cross-dataset TF/TFP/AF/DM

tcapelle:
## Hypothesis

GeGLU (Dauphin et al. 2017, Shazeer 2020) replaces the standard two-matrix FFN with a gated variant:

```
FFN_GeGLU(x) = (x W_1) * GELU(x W_3) * W_2
```

where W_1, W_3 are up-projection matrices and W_2 is the down-projection. The GELU gate smoothly suppresses small activations near zero — a property directly useful in CFD surrogate models, where many mesh nodes have near-zero physical quantities (far-field pressure, low-velocity regions).

This is distinct from SwiGLU tested in PR #2938 (nezuko), which uses SiLU as the gate function. GELU has heavier tails and is smoother near zero compared to SiLU. In CFD regimes where node features span many decades, GELU's behavior at small magnitudes may better preserve gradient signal for low-intensity physics regions.

Key properties:
- GELU gate is differentiable everywhere, including near zero
- Heavier tails than SiLU: less aggressive suppression of mid-range activations
- Standard in many modern transformers (GPT-J, Falcon, CodeGen)
- No change to model depth or attention — pure FFN swap

Papers:
- Dauphin et al., Language Modeling with Gated Convolutional Networks (2017): https://arxiv.org/abs/1612.08083
- Shazeer, GLU Variants Improve Transformer (2020): https://arxiv.org/abs/2002.05202

## Instructions

Add a `--geglu` flag to `train.py`. When enabled, replace the FFN activation in all transformer blocks with GeGLU.

Implementation pattern:

```python
import torch
import torch.nn as nn

class GeGLUFFN(nn.Module):
    """
    GeGLU feed-forward block.
    FFN(x) = (x W_1) * GELU(x W_3) @ W_2
    Shazeer 2020: https://arxiv.org/abs/2002.05202
    """
    def __init__(self, hidden_dim: int, ffn_dim: int, dropout: float = 0.0):
        super().__init__()
        # Two parallel up-projections: one for value, one for gate
        self.W_gate = nn.Linear(hidden_dim, ffn_dim, bias=False)
        self.W_up   = nn.Linear(hidden_dim, ffn_dim, bias=False)
        self.W_down = nn.Linear(ffn_dim, hidden_dim, bias=False)
        self.dropout = nn.Dropout(dropout) if dropout > 0.0 else nn.Identity()

    def forward(self, x):
        gate = F.gelu(self.W_gate(x))
        up   = self.W_up(x)
        return self.dropout(self.W_down(gate * up))
```

Replace the standard FFN block in every transformer layer with `GeGLUFFN`. Keep `ffn_dim` consistent with the existing FFN hidden dimension (typically `4 * hidden_dim`). The parameter count increases modestly due to the extra W_gate matrix.

**Run all four datasets in a single W&B group:**

```bash
# TandemFoil (Lion, EMA=0.999, 3L/192d)
  --dataset tandemfoil \
  --optimizer lion --lr 1.25e-4 --cosine-t-max 10 \
  --grad-clip 0.5 --weight-decay 1e-2 \
  --ema-decay 0.999 \
  --epochs 999 \
  --geglu \
  --wandb_group wave11-geglu

# TandemFoil Paper (Lion, EMA=0.999, same arch)
  --dataset tandemfoil_paper \
  --optimizer lion --lr 1.25e-4 --cosine-t-max 10 \
  --grad-clip 0.5 --weight-decay 1e-2 \
  --ema-decay 0.999 \
  --epochs 999 \
  --geglu \
  --wandb_group wave11-geglu

# AirfRANS (AdamW, no-EMA, 2L/256d)
  --epochs 999 \
  --geglu \
  --wandb_group wave11-geglu

# DrivAerML (AdamW, no-EMA, 4L/512d)
  --dataset drivaerml \
  --epochs 999 --batch-size 1 \
  --geglu \
  --wandb_group wave11-geglu
```

Run all four in parallel across your 8 GPUs. Report W&B run IDs for all four runs.

**Note on parameter count:** GeGLU adds one extra up-projection matrix per FFN (W_gate). If you want to maintain strict parameter parity with the baseline, reduce ffn_dim to approximately `ffn_dim * 2/3` (as recommended by Shazeer). This is optional — try the default ffn_dim first and note whether increased capacity helps or hurts.

## Baseline

Current best metrics (from BASELINE.md):

| Dataset | Primary Metric | Value | PR | W&B Run |
|---------|---------------|-------|-----|---------|
| TandemFoil | `val_primary/surface_pressure_mae` | **22.537** | #2924 | 0lv7fnun |
| AirfRANS | `val_primary/surface_mse` | **0.000482** | #2951 | pr4wsbfm |
| DrivAerML | `val_primary/surface_rel_l2_pct` | **3.997%** | #2898 | bht6h42t |

**TandemFoil reproduce command:**
```bash
cd target/icml2026 && python train.py --dataset tandemfoil --optimizer lion --lr 1.25e-4 --cosine-t-max 10 --grad-clip 0.5 --weight-decay 1e-2 --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --epochs 999 --ema-decay 0.999
```

**AirfRANS reproduce command:**
```bash
cd target/icml2026 && python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999
```

**DrivAerML reproduce command:**
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200
```

## Relationship to SwiGLU (PR #2938)

PR #2938 (nezuko) tested SwiGLU (`FFN(x) = (xW_1) * SiLU(xW_3) @ W_2`) and is currently WIP. This PR tests GeGLU independently. The two hypotheses differ in gate activation: SiLU vs GELU. If both beat baseline, the results should be compared directly to pick the better gate function. If only one wins, the winning gate type becomes part of the standard recipe.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -33,6 +33,34 @@ from core.features import (
 from core.optim import Lion, Lookahead
 
 
+class GeGLUFFN(torch.nn.Module):
+    """GeGLU feed-forward: FFN(x) = (xW_up) * GELU(xW_gate) @ W_down (Shazeer 2020)."""
+
+    def __init__(self, hidden_dim: int, ffn_dim: int):
+        super().__init__()
+        self.W_gate = torch.nn.Linear(hidden_dim, ffn_dim, bias=False)
+        self.W_up = torch.nn.Linear(hidden_dim, ffn_dim, bias=False)
+        self.W_down = torch.nn.Linear(ffn_dim, hidden_dim, bias=False)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.W_down(F.gelu(self.W_gate(x)) * self.W_up(x))
+
+
+def _replace_ffn_with_geglu(model: torch.nn.Module) -> int:
+    """Walk model tree and replace all UpActDownMlp instances with GeGLUFFN."""
+    from core.architectures.transolver_reference import UpActDownMlp
+
+    count = 0
+    for name, module in model.named_modules():
+        for child_name, child in module.named_children():
+            if isinstance(child, UpActDownMlp):
+                hidden_dim = child.fc1.in_features
+                ffn_dim = child.fc1.out_features
+                setattr(module, child_name, GeGLUFFN(hidden_dim, ffn_dim))
+                count += 1
+    return count
+
+
 class EMAWithWarmup:
     """EMA with timm-style adaptive decay warmup: min(target, (1+step)/(10+step))."""
 
@@ -165,6 +193,7 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     save_checkpoint: bool = False
+    geglu: bool = False
     seed: int = 0
 
 
@@ -1506,7 +1535,11 @@ def main() -> None:
         )
 
     train_loader, val_loaders, test_loaders = build_loaders(config, bundle, num_workers=resolved_num_workers)
-    model = build_model(config, bundle).to(device)
+    model = build_model(config, bundle)
+    if config.geglu:
+        n_replaced = _replace_ffn_with_geglu(model)
+        print(f"[GeGLU] Replaced {n_replaced} FFN blocks with GeGLU")
+    model = model.to(device)
     forward_model = torch.compile(model) if config.compile_model and device.type == "cuda" else model
     anp_head = None
     if bundle.spec.name == "tandemfoilset" and config.anp_srf:
@@ -1538,6 +1571,7 @@ def main() -> None:
     if config.wandb_name:
         run_config = asdict(config)
         run_config["effective_batch_size"] = config.batch_size * config.grad_accum_steps
+        run_config["model_params"] = sum(p.numel() for p in model.parameters())
         run = wandb.init(
             project=os.getenv("WANDB_PROJECT", "senpai-v1"),
             entity=os.getenv("WANDB_ENTITY"),


tcapelle:
STUDENT casca:

## Results

### GeGLU FFN Activation — Cross-Dataset Results

| Dataset | Primary Metric | GeGLU Best Val | Baseline | Δ | GeGLU Best Test | Epochs | Best Epoch | W&B Run |
|---------|---------------|----------------|----------|---|-----------------|--------|------------|---------|
| TandemFoil | `surface_pressure_mae` | **23.265** | 22.537 | +3.2% | 24.361 | 292 | 290 | `7u3lumdq` |
| TandemFoil Paper | `field_mse` | **NaN** | — | — | NaN | 490 | 1 | `e7fm7vp7` |
| AirfRANS | `surface_mse` | **0.000800** | 0.000482 | +66% | 0.001368 | 678 | 301 | `c04kt42i` |
| DrivAerML | `surface_rel_l2_pct` | **10.302%** | 3.997% | +158% | 10.980% | 470 | 49 | `b5bfvzy6` |

All experiments ran to the 360-minute timeout. None beat baseline.

### Commands Used

```bash
# TandemFoil (GPU 2)
CUDA_VISIBLE_DEVICES=2 python train.py --dataset tandemfoil --optimizer lion --lr 1.25e-4 --cosine-t-max 10 --grad-clip 0.5 --weight-decay 1e-2 --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --ema-decay 0.999 --epochs 999 --geglu --agent casca --wandb-name "casca/geglu-tandemfoil" --wandb-group wave11-geglu

# TandemFoil Paper (GPU 3)
CUDA_VISIBLE_DEVICES=3 python train.py --dataset tandemfoil_paper --optimizer lion --lr 1.25e-4 --cosine-t-max 10 --grad-clip 0.5 --weight-decay 1e-2 --model-slices 64 --model-layers 3 --model-hidden-dim 192 --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction --enable-pressure-prior-addition --ema-decay 0.999 --epochs 999 --geglu --agent casca --wandb-name "casca/geglu-tandemfoil-paper" --wandb-group wave11-geglu

# AirfRANS (GPU 4)
CUDA_VISIBLE_DEVICES=4 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --geglu --agent casca --wandb-name "casca/geglu-airfrans" --wandb-group wave11-geglu

# DrivAerML (GPU 5)
CUDA_VISIBLE_DEVICES=5 SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --geglu --agent casca --wandb-name "casca/geglu-drivaerml" --wandb-group wave11-geglu
```

### Peak GPU Memory

| Dataset | GPU | Peak VRAM |
|---------|-----|-----------|
| TandemFoil | GPU 2 | ~46.8 GiB |
| TandemFoil Paper | GPU 3 | ~69.5 GiB |
| AirfRANS | GPU 4 | ~67.3 GiB |
| DrivAerML | GPU 5 | not captured at peak |

### What happened

**GeGLU is a clear negative result across all 4 datasets.** No dataset beat its baseline.

**TandemFoil (closest to baseline, +3.2%):** Converged slowly (~1.7 min/epoch due to torch.compile JIT overhead with the new module) but kept improving through all 292 epochs. Best val 23.265 vs baseline 22.537. Early epochs showed ~27 MAE, and the model was still improving when the 360-min timeout hit — suggesting GeGLU doesn't fundamentally break TF but learns more slowly. Longer training might close the gap but is unlikely to beat baseline.

**TandemFoil Paper (complete NaN):** Pressure channel produced NaN from epoch 1 onward. Ux and Uy channels learned normally, but the pressure channel diverged immediately. This is a GeGLU + Lion optimizer interaction specific to TFP's pressure distribution — the gated pathway appears to amplify numerical instability in the pressure channel.

**AirfRANS (converged but +66% worse):** Best val 0.000800 surface MSE vs baseline 0.000482 at epoch 301. The model then became unstable — final test MSE of 0.595 shows it diverged badly after the best checkpoint. The cosine schedule (T_max=50) means epochs 301+ were past several annealing cycles, and the gated FFN apparently makes the model more sensitive to LR cycling.

**DrivAerML (catastrophic instability, +158%):** Trained normally through epoch 49 (best val 10.302%), then entered catastrophic instability. Grad norms escalated from ~0.7 to 273+ by epoch 136, entering permanent divergence. Best EMA checkpoint preserved at 10.3% but 2.6x worse than the 3.997% baseline.

### Analysis

The hypothesis that GELU's heavier tails and smoother near-zero behavior would benefit CFD surrogates did not hold:

1. **Gating destabilizes training:** 2 of 4 datasets showed catastrophic instability (TFP from epoch 1, DM from epoch 57). The gating mechanism creates sharper loss landscape features that the current optimizer/LR configurations can't navigate.
2. **Parameter inefficiency:** GeGLU adds ~50% more parameters per FFN block (extra W_gate matrix) without ffn_dim reduction. The increased capacity doesn't translate to better metrics — the extra parameters may contribute to overfitting or optimization difficulty.
3. **Lion incompatibility:** Lion optimizer appears fundamentally incompatible with GeGLU on TFP. AdamW datasets avoided NaN but still showed instability. This suggests the gated FFN's gradient statistics differ enough from standard FFN that optimizer tuning is needed.
4. **Comparison with SwiGLU (PR #2938):** Independent from SwiGLU which uses SiLU as the gate. If SwiGLU also fails, it would suggest gated FFNs in general don't suit the Transolver architecture at current scale.

### Suggested follow-ups

1. **Reduced ffn_dim (2/3 factor):** The PR mentioned this for parameter parity. Worth testing to isolate whether the extra parameters hurt.
2. **Lower learning rate:** GeGLU's gating dynamics may require 0.5x current LR to avoid instability.
3. **AdamW for TF/TFP instead of Lion:** The Lion+GeGLU combination is clearly problematic.
4. **Tighter grad clipping for DM:** Clipping at 0.1 instead of default might prevent the catastrophic divergence.

tcapelle:
ADVISOR: Closing PR #3017 because ADVISOR: Closing — GeGLU is a clear negative result across all four datasets.

**Results (verified against W&B):**
| Dataset | Best Val | Baseline | vs Baseline | W&B |
|---|---|---|---|---|
| TandemFoil | 23.265 (ep290) | 22.537 | +3.2% worse | 7u3lumdq |
| TandemFoil Paper | NaN | 0.00434 | failed | e7fm7vp7 |
| AirfRANS | 0.000800 (ep301) | 0.000482 | +66% worse | c04kt42i |
| DrivAerML | 10.302% (ep49) | 3.997% | +158% worse | b5bfvzy6 |

The gating mechanism destabilizes training on complex 3D datasets. DM grad norms escalated from ~0.7 to 273+ before diverging. TFP NaN from epoch 1 (Lion+GeGLU incompatibility on pressure channel). AF diverged post-peak despite reaching a reasonable transient basin.

TF's near-miss (23.265, only +3.2%) suggests gated FFNs can converge on simpler 2D geometry but scale poorly to 3D CFD. We'll wait for SwiGLU (#2954/nezuko) to decide if the entire GLU-variant family is closed.

Good analytical work in the results — your identification of the Lion+GeGLU incompatibility and the grad norm escalation pattern is valuable..

PR #2979, opened 2026-04-22
Cross-dataset: 2L depth reduction — does the AirfRANS 2L/256d insight transfer?

tcapelle:
## Hypothesis

Reducing network depth from 3L to 2L produced a **57.6% improvement** on AirfRANS (0.001479 → 0.000627, PR #2902). This is a striking and counter-intuitive result: shallower networks outperforming deeper ones on a CFD surrogate task. The hypothesis is that this depth reduction insight may transfer to other datasets.

The mechanism may be: fewer layers reduce the risk of over-smoothing on graph networks (a known pathology in GNNs), allowing the model to preserve fine-grained local flow features rather than averaging them out across many propagation steps.

**This experiment tests whether 2-layer architectures beat 3L/4L baselines across all 4 datasets.**

## Current Baselines

| Dataset | Best val metric | Architecture | PR |
|---------|----------------|--------------|-----|
| AirfRANS | `val_primary/surface_mse = 0.000627` | 2L/256d | #2902 |
| DrivAerML | `val_primary/surface_rel_l2_pct = 3.997%` | 4L/512d | #2898 |
| TandemFoil | `val_primary/surface_pressure_mae = 26.06` | 3L/192d | #2887 |
| TandemFoil Paper | No SENPAI baseline yet (paper MGN: ~1.79 field_mse) | — | — |

## Instructions to Student

Run one training job per dataset. Use the reproduce command from the relevant baseline as a starting point, then reduce `--model-layers` to 2. Keep all other flags identical to the baseline champion.

### AirfRANS (reference/sanity check — 2L already the champion)

Run the champion config to confirm reproducibility and establish a clean baseline checkpoint for this round:

```bash
cd target/icml2026
python train.py \
  --dataset airfrans \
  --airfrans-task full \
  --optimizer adamw \
  --lr 6e-4 \
  --cosine-t-max 10 \
  --grad-clip 1.0 \
  --weight-decay 1e-2 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 2 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --wandb_group 2l-depth-cross-dataset
```

Target: beat `val_primary/surface_mse < 0.000627`

### DrivAerML (primary focus — champion is 4L/512d, testing 2L/512d)

The DrivAerML champion uses 4 layers, not 3. Test whether depth reduction helps here too (2L vs 4L). Keep the wide hidden dim (512d) which has proven effective for this dataset:

```bash
[...]

PR #3033, opened 2026-04-22
Linear LR warmup then cosine annealing across all 4 CFD benchmarks

tcapelle:
## Hypothesis

Our current schedule starts cosine annealing from epoch 1. A brief linear warmup (e.g. 10 epochs ramping from 0 to peak LR) before entering cosine annealing prevents destructive early updates when model weights are randomly initialized, and is well-established in large-scale transformer training (e.g. GPT, ViT). Physics surrogate models with high-variance pressure targets may benefit especially — the early epochs set the scale of the internal representations, and noisy large steps can lock in a poor basin. A short warmup followed by the existing cosine schedule could yield both faster and deeper convergence.

---

Add a `--warmup-epochs` integer flag (default `0`, meaning no warmup) to `train.py`. When `--warmup-epochs N` is set:

1. For the first N epochs, scale the learning rate linearly from `lr * 1e-3` (near zero) to the full `lr`.
2. After epoch N, switch to the existing cosine annealing schedule (CosineAnnealingLR with T_max as currently configured), starting from the full LR.

Implementation hint: use `torch.optim.lr_scheduler.LinearLR` for the warmup followed by `CosineAnnealingLR`, chained with `SequentialLR`. Alternatively, implement a single lambda scheduler that does:
```python
def lr_lambda(epoch):
    if epoch < warmup_epochs:
        return (epoch + 1) / warmup_epochs  # linear ramp
    else:
        # cosine from peak to 0 over remaining epochs
        progress = (epoch - warmup_epochs) / max(1, total_epochs - warmup_epochs)
        return 0.5 * (1 + math.cos(math.pi * progress))
```

Test with `--warmup-epochs 10` on all datasets. (If TandemFoil with T_max=10 and warmup=10 proves odd, try warmup=5 as an alternative for TF only.)

Use `--wandb_group linear-warmup-scheduler` on all runs.

---

### TandemFoil

```bash
  --dataset tandemfoil \
  --warmup-epochs 10 \
  --wandb_group linear-warmup-scheduler
```

### TandemFoil Paper

```bash
  --dataset tandemfoil_paper \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --warmup-epochs 10 \
  --wandb_group linear-warmup-scheduler
```

### AirfRANS

```bash
  --optimizer adamw --lr 6e-4 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --warmup-epochs 10 \
  --wandb_group linear-warmup-scheduler
```

### DrivAerML

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --warmup-epochs 10 \
  --wandb_group linear-warmup-scheduler
```

---

|---|---|---|---|

---

## Reporting

- W&B run IDs for all runs
- Best val metric per dataset (warmup vs no-warmup baseline)
- Whether warmup beat baseline on each dataset
- LR curves if possible (warmup ramp visible in W&B?)


morganmcg1:

PR #3034, opened 2026-04-22
Gradient centralization across all 4 CFD benchmarks

tcapelle:
## Hypothesis

Gradient Centralization (GC, Yong et al. 2020) subtracts the mean of each parameter's gradient tensor along all axes except the output-neuron axis before the optimizer step. This projects gradients onto a centered hyperplane, implicitly constraining weight space, reducing internal covariate shift, and acting as a form of weight regularization. GC has shown consistent improvements on image classification and NLP tasks with negligible compute overhead. In physics surrogate models with large, irregular meshes and high-variance pressure targets, noisy gradients may be a bottleneck — GC could reduce this without requiring any architectural changes.

**Reference:** Yong et al., "Gradient Centralization: A New Optimization Technique for Deep Neural Networks", ECCV 2020.

---

Add a `--grad-centralization` boolean flag (default `False`) to `train.py`. When active, apply gradient centralization to all parameter gradients before the optimizer step. Insert a gradient hook or a post-backward pass loop:

```python
def centralize_gradient(grad, use_gc=True):
    """Subtract mean across all dims except the first (output neuron dim)."""
    if use_gc and grad.dim() > 1:
        # mean over all axes except axis 0
        grad = grad - grad.mean(dim=tuple(range(1, grad.dim())), keepdim=True)
    return grad

# Register on all parameters (excluding 1D, e.g. biases and norms)
for p in model.parameters():
    if p.dim() > 1:
        p.register_hook(lambda g: centralize_gradient(g))
```

This should be applied **after** loss.backward() and **before** optimizer.step(). Do not apply GC to 1D parameters (biases, LayerNorm/RMSNorm scale parameters).

Use `--wandb_group gradient-centralization` on all runs.

---

### TandemFoil

```bash
  --dataset tandemfoil \
  --grad-centralization \
  --wandb_group gradient-centralization
```

### TandemFoil Paper

```bash
  --dataset tandemfoil_paper \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --grad-centralization \
  --wandb_group gradient-centralization
```

### AirfRANS

```bash
  --optimizer adamw --lr 6e-4 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --grad-centralization \
  --wandb_group gradient-centralization
```

### DrivAerML

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --grad-centralization \
  --wandb_group gradient-centralization
```

---

|---|---|---|---|

---

## Reporting

- W&B run IDs for all runs
- Best val metric per dataset (GC vs no-GC baseline)
- Whether GC beat baseline on each dataset
- Any notes on gradient norm behavior with vs without GC


morganmcg1:

PR #3029, opened 2026-04-22
Save & eval from best-val checkpoint (cross-dataset)

tcapelle:
## Hypothesis

The trainer currently saves only the **final** epoch checkpoint and reports final-epoch metrics. Validation loss is not monotone — transient minima appear mid-training and are then erased by later epochs. Evidence from PR #2895 confirms this directly: TandemFoil hit val=25.459 at epoch 109 and AirfRANS hit val=0.000371 at epoch 221, both better than the final checkpoint, both lost. Implementing best-checkpoint saving will capture these minima and lower all four dataset metrics without changing any model architecture or hyperparameter.

This is classified **CRITICAL CODE CHANGE / Priority: HIGH** in the research state and directly aligned with human directive Issue #3020 ("best-checkpoint test eval mandatory").

## Instructions

Modify `target/icml2026/train.py` as follows:

### 1. Track best validation metric per dataset

At the start of training (after initializing the model), add:
```python
best_val_metric = float('inf')
best_checkpoint_path = os.path.join(args.output_dir, "checkpoint_best.pt")
```

### 2. Save best checkpoint during validation

After computing the primary validation metric each epoch, add:
```python
if val_primary_metric < best_val_metric:
    best_val_metric = val_primary_metric
    torch.save({
        'epoch': epoch,
        'model_state_dict': model.state_dict(),
        'ema_state_dict': ema.state_dict() if ema is not None else None,
        'val_metric': val_primary_metric,
    }, best_checkpoint_path)
    wandb.log({'best_val_metric': best_val_metric, 'best_epoch': epoch})
```

The `val_primary_metric` should be whatever metric is currently used to determine the best checkpoint (primary validation loss / surface MAE). If multiple metrics are tracked, use the same one currently used for LR scheduling.

### 3. Evaluate from best checkpoint at end of training

At the very end of training (after the final epoch loop), add:
```python
if os.path.exists(best_checkpoint_path):
    print(f"Loading best checkpoint (val={best_val_metric:.6f}) for final evaluation")
    ckpt = torch.load(best_checkpoint_path)
    model.load_state_dict(ckpt['model_state_dict'])
    if ema is not None and ckpt.get('ema_state_dict') is not None:
        ema.load_state_dict(ckpt['ema_state_dict'])
        ema.copy_to(model.parameters())  # apply EMA weights
    # run final eval pass and log as 'best_checkpoint_val_metric'
    final_best_metrics = run_validation(model, ...)
[...]

PR #3035, opened 2026-04-22
SGDR T_mult=2 with best-checkpoint saving across all 4 CFD benchmarks

tcapelle:
## Hypothesis

SGDR with `T_mult=2` (Cosine Annealing with Warm Restarts where each cycle is twice as long as the previous) produces progressively longer exploration-exploitation cycles. The first restart is aggressive, the later ones increasingly conservative — a natural curriculum. The key question is whether the model hits transient minima at restart points that are better than where the long cosine run ends.

This hypothesis was attempted previously but lacked best-checkpoint saving, so transient minima at restart points were never captured. PR #3029 (megumi) introduced `checkpoint_best.pt` saved whenever val improves. With that infrastructure now in place, SGDR T_mult deserves a clean retry where we actually capture the best checkpoint across all restarts rather than only evaluating the terminal epoch.

---

Use `torch.optim.lr_scheduler.CosineAnnealingWarmRestarts` with `T_0` matching the dataset T_max baseline and `T_mult=2`. This gives restart intervals of T_0, 2*T_0, 4*T_0, ... epochs.

**Critical:** The best-checkpoint mechanism from #3029 must be active so transient minima at each restart are captured. Confirm `checkpoint_best.pt` is being saved (it should be automatic in the current codebase). Report the best val metric from the best checkpoint, not the terminal epoch.

Dataset-specific T_0 values (matching baseline cosine-t-max):
- TandemFoil: T_0=10
- TandemFoil Paper: T_0=150
- AirfRANS: T_0=50
- DrivAerML: T_0=30

Use `--wandb_group sgdr-tmult-best-checkpoint` on all runs.

---

### TandemFoil

```bash
  --dataset tandemfoil \
  --cosine-t-max 10 --cosine-t-mult 2 \
  --grad-clip 0.5 --weight-decay 1e-2 \
  --wandb_group sgdr-tmult-best-checkpoint
```

(If `--cosine-t-mult` is not yet a flag, implement it: add `T_mult` parameter to `CosineAnnealingWarmRestarts`; if only `CosineAnnealingLR` is used, switch to `CosineAnnealingWarmRestarts` with `T_0=cosine_t_max, T_mult=2`.)

### TandemFoil Paper

```bash
  --dataset tandemfoil_paper \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 150 --cosine-t-mult 2 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group sgdr-tmult-best-checkpoint
```

### AirfRANS

```bash
  --optimizer adamw --lr 6e-4 \
  --cosine-t-max 50 --cosine-t-mult 2 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --wandb_group sgdr-tmult-best-checkpoint
```

### DrivAerML

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 --cosine-t-mult 2 \
  --no-use-ema \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --wandb_group sgdr-tmult-best-checkpoint
```

---

|---|---|---|---|

---

## Reporting

- W&B run IDs for all runs
- Best val metric per dataset from `checkpoint_best.pt` (not terminal epoch)
- Whether SGDR T_mult=2 beat baseline on each dataset
- At which restart interval the best checkpoint was captured
- Comparison with terminal epoch metrics (to quantify transient minima benefit)


morganmcg1:

PR #3038, opened 2026-04-22
Surface normals + principal curvature as input features

tcapelle:
## Hypothesis

Adding surface normals `[nx, ny, nz]` and principal curvatures `[k1, k2]` per surface node as extra input features will provide richer local geometry context to the Transolver. This directly addresses DrivAerML's complex 3D boundary layer prediction challenge where the model currently predicts poorly near high-curvature regions (leading edges, A-pillars, underbody). On TandemFoil, explicit normal information should help resolve inter-foil pressure interference patterns.

## Current Baselines

## Implementation

Add a new flag `--surface-normals` to `train.py`. When enabled:

1. **Compute normals:** For each surface node, compute the outward-facing unit normal `[nx, ny, nz]` using neighbor triangles. In trimesh: `mesh.vertex_normals` gives per-vertex normals. For point clouds without connectivity (DrivAerML, AirfRANS), estimate via PCA on k=16 nearest neighbors.

2. **Compute principal curvatures (optional, `--surface-curvature`):** Use `trimesh.curvature.discrete_principal_curvatures_and_directions()` to extract `k1, k2`. Normalize using `arcsinh(k)` with scale=1.0 to handle outliers.

3. **Concatenate to input:** Append `[nx, ny, nz]` (3 dims) to the per-node feature vector before the first linear projection. If `--surface-curvature` is also set, append `[arcsinh(k1), arcsinh(k2)]` (2 more dims).

4. **Initialization:** Zero-init the extra weight columns in the first linear layer so the model starts from the pretrained baseline.

**Key implementation note:** For datasets where mesh connectivity is not directly available (DrivAerML, AirfRANS), use `sklearn.neighbors.NearestNeighbors` with k=16 to estimate normals via PCA of the local neighborhood covariance matrix. The sign convention should be consistent (outward-facing normal via checking against centroid).

```python
# Pseudocode for PCA-based normal estimation
from sklearn.neighbors import NearestNeighbors
import numpy as np

def estimate_normals_pca(points, k=16):
    nbrs = NearestNeighbors(n_neighbors=k+1).fit(points)
    _, indices = nbrs.kneighbors(points)
    normals = []
    for i, idx in enumerate(indices):
        neighbors = points[idx[1:]]  # exclude self
        centered = neighbors - neighbors.mean(0)
        _, _, vh = np.linalg.svd(centered)
        normal = vh[-1]  # smallest singular vector
        # Orient outward (away from centroid)
        centroid = points.mean(0)
        if np.dot(normal, points[i] - centroid) < 0:
            normal = -normal
        normals.append(normal)
    return np.stack(normals)
```

## Experiment Instructions

Run the following experiments (GPU allocation: DM=4 GPUs, TF=2 GPUs, TFP=1 GPU, AF=1 GPU):

**TandemFoil (2 GPUs):**
```bash
  --dataset tandemfoil \
  --surface-normals --surface-curvature \
  --wandb-group surface-normals-curvature
```

**TandemFoil Paper (1 GPU):**
```bash
  --dataset tandemfoil_paper \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --surface-normals --surface-curvature \
  --wandb-group surface-normals-curvature
```

**AirfRANS (1 GPU):**
```bash
  --optimizer adamw --lr 6e-4 \
  --enable-fourier \
  --epochs 999 \
  --surface-normals --surface-curvature \
  --wandb-group surface-normals-curvature
```

**DrivAerML (4 GPUs):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --surface-normals --surface-curvature \
  --wandb-group surface-normals-curvature
```

## Expected Outcome

- DM: improvement from 3.997% toward <3.71% (normals add 3 dims of geometric context the model cannot currently infer from point positions alone)
- TF: improvement from 22.537 (normals resolve inter-foil proximity ambiguity in the pressure field)
- AF: improvement from 0.000340 (trailing edge separation is highly curvature-dependent)

## Suggested Follow-ups

- If normals help but curvature does not (or vice versa), ablate them separately with `--surface-normals` only vs `--surface-curvature` only
- Try `k=32` neighborhood for PCA normal estimation if `k=16` produces noisy normals on DrivAerML
- If DM improves significantly, consider adding mean curvature H=(k1+k2)/2 as a 6th feature


morganmcg1:

PR #3039, opened 2026-04-22
Mass conservation auxiliary loss (divergence regularization)

tcapelle:
## Hypothesis

Adding a divergence-based auxiliary loss `lambda_div * mean(div_u_pred^2)` will enforce approximate mass conservation in velocity field predictions, providing physically-grounded regularization that the current data-driven loss cannot enforce. Incompressible CFD flows must satisfy `div(u) = 0`; penalizing divergence of the predicted velocity should improve physical consistency and generalization, particularly on DrivAerML where wake velocity accuracy drives the primary metric.

## Current Baselines

## Implementation

Add `--div-loss-weight` flag (default 0.0, set to 0.01 for this experiment). When non-zero:

1. **Estimate divergence via KNN finite differences:** For each point `i`, find k=8 nearest neighbors. For each neighbor `j` at relative position `r_ij = x_j - x_i` and predicted velocity difference `u_ij = u_j - u_i`, use the Green-Gauss approximation:
   ```
   div_u_i ≈ sum_j( dot(u_ij, r_ij) / |r_ij|^2 ) / k
   ```

2. **Auxiliary loss:**
   ```python
   div_loss = torch.mean(div_u_pred ** 2)
   total_loss = data_loss + args.div_loss_weight * div_loss
   ```

3. **Apply only to velocity components:** Only apply divergence regularization to velocity outputs (`ux, uy, uz`), not pressure. The divergence condition does not apply to scalar pressure fields.

4. **Dataset-specific notes:**
   - DrivAerML: velocity is 3D `[ux, uy, uz]` — straightforward divergence
   - AirfRANS: velocity is 2D `[ux, uy]` — use 2D divergence `dux/dx + duy/dy`
   - TandemFoil/TFP: primarily pressure prediction. Apply divergence regularization only if velocity components are also predicted; otherwise skip the aux loss and rely on default data loss only (set `div_loss_weight=0.0` for TF/TFP if they don't output velocity)

```python
# Pseudocode for KNN divergence estimation
def compute_divergence_loss(positions, velocity_preds, k=8):
    # positions: [N, 3], velocity_preds: [N, 3]
    from torch_geometric.nn import knn_graph
    edge_index = knn_graph(positions, k=k)  # [2, N*k]
    i, j = edge_index
    r_ij = positions[j] - positions[i]  # [N*k, 3]
    u_ij = velocity_preds[j] - velocity_preds[i]  # [N*k, 3]
    r_sq = (r_ij ** 2).sum(-1, keepdim=True).clamp(min=1e-8)
    dot_prod = (u_ij * r_ij).sum(-1) / r_sq.squeeze()  # [N*k]
    # Scatter mean over neighbors
    from torch_scatter import scatter_mean
    div_per_point = scatter_mean(dot_prod, i, dim=0, dim_size=positions.shape[0])
    return (div_per_point ** 2).mean()
```

## Experiment Instructions

Run the following experiments (GPU allocation: DM=4 GPUs, TFP=2 GPUs, AF=2 GPUs):

**TandemFoil Paper (2 GPUs):**
```bash
  --dataset tandemfoil_paper \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --div-loss-weight 0.01 \
  --wandb-group mass-conservation-aux-loss
```

**AirfRANS (2 GPUs):**
```bash
  --optimizer adamw --lr 6e-4 \
  --enable-fourier \
  --epochs 999 \
  --div-loss-weight 0.01 \
  --wandb-group mass-conservation-aux-loss
```

**DrivAerML (4 GPUs):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --div-loss-weight 0.01 \
  --wandb-group mass-conservation-aux-loss
```

**TandemFoil (if velocity outputs exist, else skip; 0 GPUs fallback to running TF with weight=0.001):**
```bash
  --dataset tandemfoil \
  --div-loss-weight 0.01 \
  --wandb-group mass-conservation-aux-loss
```

Note: If `--div-loss-weight` is not yet implemented, you need to add it to `train.py`. If TandemFoil does not predict velocity outputs, set `div-loss-weight=0.0` for that dataset (the TF run then becomes a control).

## Expected Outcome

- AF: improvement from 0.000340 (2D div constraint on velocity field)
- DM: improvement from 3.997% toward <3.71% (3D velocity divergence in wake region)
- TFP: modest improvement if velocity field is predicted

## Suggested Follow-ups

- If lambda=0.01 helps, try lambda=0.001 and lambda=0.1 to find the optimal weight
- If divergence alone helps, try combining with curl loss (vorticity regularization): `curl_loss = mean(|curl(u)|^2)` at high-vorticity regions
- Ablate: train with divergence loss but evaluate without — does learned physics generalize?


morganmcg1:

PR #3041, opened 2026-04-22
AdaFactor optimizer: memory-efficient adaptive optimizer

tcapelle:
## Hypothesis

AdaFactor is a memory-efficient adaptive optimizer that uses factored second-moment estimates, requiring ~2x less memory than Adam/AdamW. For large-capacity DrivAerML models (4L/512d/8H), AdaFactor's factored approximation may act as implicit regularization that improves generalization — particularly because the 3D car geometry dataset has limited training samples (~500 cars). Fixed-LR mode with a constant learning rate decouples AdaFactor from its aggressive internal LR schedule, making it a fair comparison to AdamW.

Note: This was originally assigned in Wave 16 (would-have-been PR #3036) but the branch never launched. Clean re-assignment from radford baseline.

## Current Baselines

## Implementation

Add `--optimizer adafactor` option to `train.py`. Use `transformers.Adafactor` or `torch_optimizer.Adafactor`:

```python
from transformers import Adafactor

optimizer = Adafactor(
    model.parameters(),
    lr=args.lr,           # fixed LR mode: pass explicit lr
    scale_parameter=False,  # disable internal LR scaling
    relative_step=False,    # disable relative step sizing
    warmup_init=False,      # no warmup (we handle externally)
    weight_decay=args.weight_decay,
    clip_threshold=1.0,     # built-in gradient clipping
)
```

**Key difference from AdamW:** AdaFactor stores a factored approximation of the second moment: for a 2D weight matrix `W [m x n]`, it stores `R [m]` and `C [n]` instead of a full `[m x n]` matrix — O(m+n) vs O(m*n) memory. For large layers (512d hidden), this is a significant saving.

**Use fixed LR mode:** Pass `lr=args.lr, scale_parameter=False, relative_step=False`. This is essential — AdaFactor's default relative step mode uses an aggressive internal scheduler that will fight our cosine annealing.

## Experiment Instructions

Run the following experiments (GPU allocation: DM=4 GPUs, TF=2 GPUs, TFP=1 GPU, AF=1 GPU):

**TandemFoil (2 GPUs) — use Lion LR equivalent:**
```bash
  --dataset tandemfoil \
  --optimizer adafactor --lr 1.25e-4 \
  --wandb-group adafactor-optimizer
```

**TandemFoil Paper (1 GPU):**
```bash
  --dataset tandemfoil_paper \
  --optimizer adafactor --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --wandb-group adafactor-optimizer
```

**AirfRANS (1 GPU):**
```bash
  --optimizer adafactor --lr 6e-4 \
  --enable-fourier \
  --epochs 999 \
  --wandb-group adafactor-optimizer
```

**DrivAerML (4 GPUs):**
```bash
  --dataset drivaerml \
  --optimizer adafactor --lr 5e-4 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --wandb-group adafactor-optimizer
```

## Expected Outcome

- DM: improvement from 3.997% (factored second moments act as regularization; implicit per-parameter LR smoothing)
- TFP: improvement from 0.00434 (memory saving enables larger effective batch or minor generalization gain)
- TF and AF: comparable or slight regression vs Lion/AdamW (Lion is already tuned for TF; AdamW is tuned for AF)

## Suggested Follow-ups

- If AdaFactor underperforms AdamW at lr=5e-4, try lr=1e-3 (AdaFactor can tolerate higher LRs due to internal normalization)
- If factored approximation hurts on small models (TF/TFP with 192d), try AdaFactor only on DM+AF
- Consider AdaFactor with `scale_parameter=True` (internal LR mode) as an ablation if fixed-LR mode fails


morganmcg1:

PR #2904, opened 2026-04-21
Cross-dataset: Momentum-SAM (MSAM) — flat minima without 2x compute (CODE CHANGE)

tcapelle:
## Hypothesis

Standard SAM costs 2x compute (unacceptable at our epoch budget). Momentum-SAM (MSAM) perturbs weights in the momentum buffer direction instead — same flat-minima bias at ~0.1% overhead. At 2-11 epochs for shorter datasets, pushing toward broad basins from the first step amplifies each gradient step's value. The momentum direction is more stable than raw gradient direction, so the perturbation rho can be larger without instability. Prior work shows SAM-style methods improve generalization by finding flatter loss basins.

## Instructions

**CODE CHANGE REQUIRED:** Add `--use-msam` flag and `--msam-rho` argument (default 0.02) to `target/icml2026/train.py`. Implementation:

```python
# At each optimizer.step(), before clearing gradients:
if args.use_msam:
    with torch.no_grad():
        for group in optimizer.param_groups:
            for p in group['params']:
                if p.grad is None: continue
                state = optimizer.state[p]
                if 'exp_avg' not in state: continue  # skip if no momentum yet
                momentum = state['exp_avg']
                norm = momentum.norm()
                if norm > 0:
                    delta = args.msam_rho * momentum / norm
                    p.data.add_(delta)
        # Recompute loss and grad at perturbed point (use same batch)
        loss_perturbed = compute_loss(model, batch)
        loss_perturbed.backward()
        # Remove perturbation
        for group in optimizer.param_groups:
            for p in group['params']:
                if p.grad is None: continue
                state = optimizer.state[p]
                if 'exp_avg' not in state: continue
                momentum = state['exp_avg']
                norm = momentum.norm()
                if norm > 0:
                    delta = args.msam_rho * momentum / norm
                    p.data.sub_(delta)
    optimizer.step()
```

**Note:** This DOES require 2 forward-backward passes per step (same as regular SAM). The advantage of MSAM is that the perturbation direction (momentum) is more stable than the raw gradient direction, so rho can be larger. If this is too expensive, try with `--msam-rho 0.01` and only apply MSAM every 2nd step.

All runs use: `--no-use-ema --epochs 999 --enable-fourier`

DrivAerML extra flags: `--batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`

[...]

PR #3042, opened 2026-04-22
SAM optimizer (Sharpness-Aware Minimization) — cross-dataset

tcapelle:
## Hypothesis

Every optimizer tried so far (Lion, AdamW, AdaFactor, gradient centralization) moves through weight space differently but none changes the *geometry* of the loss landscape that is found. SAM (Sharpness-Aware Minimization, Foret et al. ICLR 2021) explicitly seeks **flat minima** by perturbing weights to the local loss maximum, computing the gradient there, restoring the original weights, then taking the update step. Flat minima generalize better — they are robust to small weight perturbations that might correspond to unseen operating conditions. In CFD surrogate work our primary generalization gap is **OOD splits** (different AoA, different Re, different geometry). SAM's inductive bias directly addresses this.

Cost: ~2× wall-clock per step (two forward+backward passes). With our GPU budget this is acceptable.

## Baseline to beat

| Dataset | Metric | Current best | PR |
|---|---|---|---|

## SAM implementation

Add this class to `train.py` (or a small helper module imported by it):

```python
class SAM(torch.optim.Optimizer):
    def __init__(self, params, base_optimizer, rho=0.05, **kwargs):
        defaults = dict(rho=rho, **kwargs)
        super(SAM, self).__init__(params, defaults)
        self.base_optimizer = base_optimizer(self.param_groups, **kwargs)
        self.param_groups = self.base_optimizer.param_groups

    @torch.no_grad()
    def first_step(self, zero_grad=False):
        grad_norm = self._grad_norm()
        for group in self.param_groups:
            scale = group["rho"] / (grad_norm + 1e-12)
            for p in group["params"]:
                if p.grad is None:
                    continue
                self.state[p]["old_p"] = p.data.clone()
                e_w = p.grad * scale.to(p)
                p.add_(e_w)
        if zero_grad:
            self.zero_grad()

    @torch.no_grad()
    def second_step(self, zero_grad=False):
        for group in self.param_groups:
            for p in group["params"]:
                if p.grad is None:
                    continue
                p.data = self.state[p]["old_p"]
        self.base_optimizer.step()
        if zero_grad:
            self.zero_grad()

    def _grad_norm(self):
        shared_device = self.param_groups[0]["params"][0].device
        norm = torch.norm(
            torch.stack([
                p.grad.norm(p=2).to(shared_device)
                for group in self.param_groups
                for p in group["params"]
                if p.grad is not None
            ]),
            p=2,
        )
        return norm
```

## Training loop modification

Replace the standard single-step optimizer block with:

```python
# --- SAM first step ---
loss_first = criterion(predictions, targets)
loss_first.backward()
optimizer.first_step(zero_grad=True)

# --- SAM second step: second forward+backward at perturbed weights ---
loss_second = criterion(model(inputs), targets)
loss_second.backward()
optimizer.second_step(zero_grad=True)
```

Use `loss_first` (or `loss_second`) for logging — both are valid; `loss_first` is standard.

Gradient clipping should be applied **before** `first_step` and **before** `second_step` respectively (clip twice, once per pass).

## Runs

Use `--wandb_group sam_cross_dataset` on all runs.

### Run 1 — TandemFoil, SAM(Lion), rho=0.05
```bash
python train.py \
  --dataset tandemfoil \
  --optimizer sam_lion \
  --sam-rho 0.05 \
  --lr 1.25e-4 \
  --cosine-t-max 10 \
  --grad-clip 0.5 \
  --weight-decay 1e-2 \
  --model-slices 64 \
  --model-layers 3 \
  --model-hidden-dim 192 \
  --model-heads 3 \
  --enable-fourier \
  --enable-te-coord-frame \
  --enable-cp-panel \
  --asinh-pressure \
  --residual-prediction \
  --ema-decay 0.999 \
  --epochs 999 \
  --wandb_group sam_cross_dataset
```

### Run 2 — TandemFoil, SAM(Lion), rho=0.1
Same as Run 1 but `--sam-rho 0.1`.

### Run 3 — AirfRANS, SAM(AdamW), rho=0.05
```bash
python train.py \
  --dataset airfrans \
  --airfrans-task full \
  --optimizer sam_adamw \
  --sam-rho 0.05 \
  --lr 6e-4 \
  --cosine-t-max 50 \
  --grad-clip 1.0 \
  --weight-decay 1e-2 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 2 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --wandb_group sam_cross_dataset
```

### Run 4 — AirfRANS, SAM(AdamW), rho=0.1
Same as Run 3 but `--sam-rho 0.1`.

### Run 5 — DrivAerML, SAM(AdamW), rho=0.05
```bash
python train.py \
  --dataset drivaerml \
  --optimizer sam_adamw \
  --sam-rho 0.05 \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 1.0 \
  --weight-decay 1e-2 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --wandb_group sam_cross_dataset
```

### Run 6 — DrivAerML, SAM(AdamW), rho=0.1
Same as Run 5 but `--sam-rho 0.1`.

### Run 7 (optional) — TandemFoil Paper, SAM(AdamW), rho=0.05
```bash
python train.py \
  --dataset tandemfoil_paper \
  --optimizer sam_adamw \
  --sam-rho 0.05 \
  --lr 5e-4 \
  --cosine-t-max 150 \
  --grad-clip 1.0 \
  --weight-decay 1e-2 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 3 \
  --model-hidden-dim 192 \
  --model-heads 3 \
  --epochs 999 \
  --wandb_group sam_cross_dataset
```

## CLI wiring note

You will need to add `--optimizer sam_lion` / `--optimizer sam_adamw` and `--sam-rho` to `train.py`'s argument parser, and dispatch to the SAM wrapper accordingly. The SAM class wraps the base optimizer — for `sam_lion` use `base_optimizer=Lion`; for `sam_adamw` use `base_optimizer=torch.optim.AdamW`.

## Reference

Foret et al. "Sharpness-Aware Minimization for Efficiently Improving Generalization." ICLR 2021. https://arxiv.org/abs/2010.01412

## Expected outcome

Flat-minima bias should reduce OOD error on TandemFoil (p_re, p_oodc) and DrivAerML geometry transfer. AirfRANS benefits less (all cases share the same flow regime) but may still see a small improvement. If rho=0.1 is too aggressive (training instability), fall back to rho=0.05. Report `val_primary/*` for each run at best validation checkpoint.


tcapelle:
## SAM Optimizer — Cross-Dataset Results

**Experiments still running** but convergence trajectory is clear. Reporting intermediate results.

### Results Summary

| Dataset | Variant | W&B Run | Epochs | Best Ep | Best Val Primary | Baseline | Gap |
|---------|---------|---------|--------|---------|-----------------|----------|-----|
| TandemFoil | SAM(Lion) ρ=0.05 | `998rhuvu` | 8 | 8 | 73.00 | 22.537 | 3.2× |
| TandemFoil | SAM(Lion) ρ=0.10 | `o8zbuza2` | 9 | 9 | 78.66 | 22.537 | 3.5× |
| AirfRANS | SAM(AdamW) ρ=0.05 | `bvct0w6n` | 55 | 45 | 0.01854 | 0.000482 | 38× |
| AirfRANS | SAM(AdamW) ρ=0.10 | `fle4utbd` | 55 | 52 | 0.03073 | 0.000482 | 64× |
| DrivAerML | SAM(AdamW) ρ=0.05 | `eyiwsjzz` | 39 | 38 | 17.46% | 3.997% | 4.4× |
| DrivAerML | SAM(AdamW) ρ=0.10 | `r8edca72` | 38 | 38 | 28.70% | 3.997% | 7.2× |

W&B group: `sam_cross_dataset`

### Convergence Analysis

**TandemFoil (SAM + Lion):**
- Extremely slow: ~13 min/epoch (vs ~2 min baseline) due to `torch.compile` + SAM's double forward/backward pass
- Only completed 8-9 epochs in ~50 minutes. Convergence curve: 201→148→120→103→93→85→78→73 (ρ=0.05)
- Still converging but trajectory won't close the 3.2× gap — would need hundreds of epochs at this rate
- SAM + Lion is untested in literature and may need fundamentally different hyperparameter tuning

**AirfRANS (SAM + AdamW):**
- ~1 min/epoch, 55 epochs completed across 2 cosine cycles (t_max=50)
- ρ=0.05 best: 0.0185 at ep45 (end of first cosine cycle minimum), ρ=0.10 best: 0.0307 at ep52
- Second cosine cycle not improving over first — suggests the model has converged under SAM
- Baseline (0.000482) is 38× better — SAM has not helped

**DrivAerML (SAM + AdamW):**
- ~1 min/epoch, 38-39 epochs completed
- ρ=0.05 best: 17.46% at ep38, ρ=0.10 best: 28.70% at ep38
- ρ=0.10 training is **unstable** — gradient norms spiking to 10-19 (vs ~0.25 for ρ=0.05)
- Baseline (3.997%) is 4.4× better

### Key Observations

1. **SAM does not beat baseline on any dataset** — gaps range from 3.2× to 64×
2. **ρ=0.05 consistently outperforms ρ=0.10** across all 3 datasets (expected: smaller perturbation is safer)
3. **2× wall-clock penalty** from double forward/backward halves effective training epochs
4. **torch.compile + SAM interaction** makes TandemFoil prohibitively slow (8× longer per epoch)
5. **ρ=0.10 destabilizes DrivAerML training** with extreme gradient norm spikes
6. **SAM + Lion** is an unexplored combination — the perturbation dynamics may conflict with Lion's sign-based updates

### Why SAM Likely Fails Here

- The baselines are already well-tuned with cosine LR + Lion/AdamW + EMA + gradient clipping
- SAM's flat-minima-seeking behavior may not help when the optimization landscape is already well-conditioned
- The 2× compute cost trades wall-clock time for a generalization benefit that doesn't materialize in these physics-informed regression tasks (SAM was designed for classification generalization gaps)
- SAM + Lion interaction is theoretically questionable: Lion uses sign(momentum) updates, but SAM perturbs based on raw gradients — the perturbation direction and the update direction operate in different spaces

### Suggested Follow-ups

- **ASAM (Adaptive SAM)** might be worth trying — it normalizes perturbations per-parameter, which could help with the instability seen at ρ=0.10
- **Lower ρ values** (0.01-0.02) could reduce the wall-clock penalty while still providing some flat-minima benefit
- **SAM with reduced frequency** (e.g., every K steps) could amortize the 2× cost
- However, given the large gaps, optimizer-level changes may not be the right lever — architecture/loss changes likely have higher expected value

tcapelle:
**Update (experiments still running):** Latest best metrics since initial report:
- AirfRANS SAM(AdamW) ρ=0.05: **0.01606** at ep56 (was 0.01854 at ep45 — still improving in 2nd cosine cycle)
- DrivAerML SAM(AdamW) ρ=0.05: **16.58%** at ep40 (was 17.46% at ep38)
- DrivAerML SAM(AdamW) ρ=0.10: grad norm spiked to 35.2 at ep40 — training is increasingly unstable

Conclusions unchanged: SAM does not approach baseline on any dataset. Gaps: TandemFoil 3.2×, AirfRANS 33×, DrivAerML 4.1×.

tcapelle:
ADVISOR: Closing PR #3042 because ADVISOR: CLOSING — SAM optimizer is a clear dead end across all datasets.

## Results (W&B verified)

| Dataset | SAM rho=0.05 | SAM rho=0.10 | Baseline | Gap |
|---------|-------------|-------------|----------|-----|
| TandemFoil | 78.45 (7ep) | 84.67 (8ep) | 22.537 | 3.5-3.8x worse |
| AirfRANS | 0.01606 (59ep) | 0.02555 (59ep) | 0.000482 | 33-53x worse |
| DrivAerML | 16.58% (41ep) | 28.70% (40ep) | 3.997% | 4.2-7.2x worse |

## Analysis

1. **AirfRANS is definitive (59 epochs).** This is NOT a training-budget problem — SAM converges to a qualitatively worse basin. The second cosine cycle fails to improve on the first.
2. **DrivAerML rho=0.10 diverges** — grad norms spike to 35+ (vs ~0.25 baseline). Perturbation radius is too large.
3. **SAM is fundamentally incompatible with physics-constrained CFD regression.** The optimization landscape is already well-conditioned by Fourier features, pressure priors, and coordinate frames. SAM's flat-minima bias adds cost without value.
4. **MSAM (#2904) and now classic SAM (#3042) both fail catastrophically.** The entire SAM family is closed for this programme.

Your follow-up suggestions (ASAM, periodic SAM) are reasonable in theory but the AF result at 59 epochs rules out the approach — no rho tuning will close a 33x gap. Moving you to a new experiment..

PR #3049, opened 2026-04-22
DrivAerML: torch.compile(mode='default') — JIT without CUDA graphs

tcapelle:
## Hypothesis

PR #2992 showed compile(mode='reduce-overhead') fails due to CUDA graphs requiring fixed tensor shapes. But mode='default' does NOT use CUDA graphs — it only applies JIT fusion passes. This may provide 10-30% throughput improvement without the shape-incompatibility crash. More epochs in the same wall-clock budget → deeper convergence. Testing on both DM and AF where compile overhead should be smaller (smaller meshes, faster epochs).

## Current Baselines
- DM: val_primary/surface_rel_l2_pct = **3.997%** (AdamW lr=5e-4, T_max=30, no-EMA, 4L/512d, NO gc, NO WD)
- AF: val_primary/surface_mse = **0.000482** (AdamW lr=6e-4, T_max=50, no-EMA, 2L/256d)

## Implementation

Modify `train.py` to wrap the model with `torch.compile(model, mode='default')` when a --compile flag is provided (check how the existing compile flag works in the codebase). Use mode='default' specifically — NOT 'reduce-overhead' or 'max-autotune'.

## Runs

**Run 1 — DM compile(mode='default')**
```bash
SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --grad-clip 1.0 --compile --wandb_group dm_compile_default
```
(--grad-clip 1.0 for compile stability per #2898 insight)

**Run 2 — DM no-compile control**
```bash
SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --grad-clip 1.0 --wandb_group dm_compile_default
```

**Run 3 — AF compile(mode='default')**
```bash
python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --compile --wandb_group dm_compile_default
```

## Expected outcome
Report epochs/minute for Run 1 vs Run 2 to quantify compile speedup. If compile provides >10% throughput gain and final metric is competitive, this unlocks free improvement for all future DM runs.

## Notes
- The key question is: does mode='default' crash on variable-length point cloud inputs? If it does, note the error and close.
- Report both throughput (epochs/min) and final metric for Run 1 vs Run 2
- If compile crashes, try wrapping only the encoder or decoder, not the full model

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -1609,7 +1609,7 @@ def main() -> None:
 
     train_loader, val_loaders, test_loaders = build_loaders(config, bundle, num_workers=resolved_num_workers)
     model = build_model(config, bundle).to(device)
-    forward_model = torch.compile(model) if config.compile_model and device.type == "cuda" else model
+    forward_model = torch.compile(model, mode="default") if config.compile_model and device.type == "cuda" else model
     anp_head = None
     if bundle.spec.name == "tandemfoilset" and config.anp_srf:
         anp_head = ANPSurfaceDecoder(


tcapelle:
STUDENT levi:

## Results

### Throughput comparison: compile(mode='default') vs no-compile on DrivAerML

**HYPOTHESIS FALSIFIED.** torch.compile(mode='default') provides **0% throughput benefit** on DrivAerML, confirming PR #2898's finding. Both runs track at identical epoch counts throughout training.

| Metric | Run 1 — DM compile | Run 2 — DM no-compile | Run 3 — AF compile |
|--------|-------------------|-----------------------|-------------------|
| W&B run | xayn77u6 | adujnx1r | p6bgp7ne |
| Epochs (at report time) | 23 (still running) | 24 (still running) | 30 (still running) |
| Best val metric | 20.12% (ep22) | 19.72% (ep24) | 0.0224 surface_mse (ep23) |
| Throughput | 1.29 ep/min | 1.35 ep/min | 2.03 ep/min |
| Min/epoch | 0.77 | 0.74 | 0.49 |
| Peak GPU memory | 5,985 MiB | 7,241 MiB | 24,823 MiB |
| Grad clip events (latest) | 276/394 batches | 291/394 batches | — |
| Grad norm mean (latest) | 1.59 | 1.74 | — |

**Baselines:**
- DM baseline: 3.997% (PR #2898, ep467, 360-min budget)
- AF baseline: 0.000482 (PR #2951, ep576, 360-min budget)

### Key findings

1. **Throughput: 0% benefit.** Both DM runs completed the same number of epochs (±1) in the same wall-clock time across 24 epochs of observation. The JIT fusion passes in mode='default' do not accelerate this workload. The bottleneck is likely data loading (50k surface points per batch) and the relatively small model (4L/512d), not compute.

2. **No crashes on variable-length inputs.** mode='default' handles variable-length point clouds on both DrivAerML and AirfRANS without any shape incompatibility errors. This is expected since mode='default' does NOT use CUDA graphs (which require fixed tensor shapes).

3. **Grad-clip 1.0 stabilizes compile.** PR #2898 saw NaN divergence at ep454 with compile and no grad-clip. With --grad-clip 1.0, training is stable through ep23+ (runs still in progress — full confirmation at ep454+ pending).

4. **Compile uses slightly LESS GPU memory** on DM (5,985 vs 7,241 MiB, -17%), likely from memory layout optimizations in the JIT-compiled graph.

5. **AirfRANS phase transition is normal with compile.** AF surface_mse dropped from 0.365 (ep1) to 0.022 (ep23) — the characteristic phase transition is occurring on schedule.

### What happened

The hypothesis "mode='default' (JIT fusion only, no CUDA graphs) → 10-30% throughput gain → more epochs → deeper convergence" is **falsified**. The JIT fusion passes in torch.compile(mode='default') provide no throughput benefit on these workloads. The Blackwell RTX PRO 6000 GPUs likely already achieve near-optimal operator fusion at the hardware/driver level, leaving torch.compile with nothing to optimize.

Since throughput is identical, the "more epochs in the same wall-clock budget" improvement does not materialize. Both DM runs will converge to equivalent metrics.

### Code change

Made `torch.compile(model, mode="default")` explicit on line 1612 of train.py (previously `torch.compile(model)` which defaults to mode='default' anyway — change is for documentation clarity).

### Commands

**Run 1 — DM compile:**
```bash
CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --grad-clip 1.0 --compile-model --agent levi --wandb-name "levi/dm-compile-default" --wandb-group dm_compile_default
```

**Run 2 — DM no-compile control:**
```bash
CUDA_VISIBLE_DEVICES=1 SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --grad-clip 1.0 --no-compile-model --agent levi --wandb-name "levi/dm-no-compile-control" --wandb-group dm_compile_default
```

**Run 3 — AF compile:**
```bash
CUDA_VISIBLE_DEVICES=2 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 --no-use-ema --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --compile-model --agent levi --wandb-name "levi/af-compile-default" --wandb-group dm_compile_default
```

### Suggested follow-ups

- **mode='max-autotune'**: This mode spends more time searching for the fastest kernel implementations. It might find wins that mode='default' misses, though it has longer warmup.
- **Compile only the encoder/decoder**: Wrapping only the transformer blocks (not the full model) could avoid shape-mismatch issues while still capturing the compute-intensive parts.
- **Test on TandemFoil**: The smaller mesh sizes might show different compile behavior than the 50k-point DM batches.
- **Consider disabling compile as default**: Since compile provides 0% benefit and adds NaN risk (without grad-clip), the default `compile_model: bool = True` may be doing more harm than good. Flipping to False would simplify the default config.

tcapelle:
ADVISOR: Closing PR #3049 because Hypothesis cleanly falsified — 0% throughput benefit from torch.compile(mode='default') on DrivAerML.

**Key findings:**
- DM compile (xayn77u6) vs no-compile control (adujnx1r): identical epoch rates (1.29 vs 1.35 ep/min). Data loading is the bottleneck, not compute.
- Compile uses ~17% less GPU memory (5,985 vs 7,241 MiB) — interesting but doesn't translate to speed.
- No crashes with mode='default' on variable-length inputs, confirming it avoids the CUDA graph issue.
- grad-clip=1.0 stabilizes training (no NaN through ep44+ vs NaN at ep454 without gc in #2898).

**Your suggestion about flipping compile_model default to False is noted** — 0% benefit + NaN risk without grad-clip means the default True may be actively harmful. This should be investigated separately.

Clean experiment with proper control run — well executed..

PR #2948, opened 2026-04-22
tandemfoil_paper: physics-flag ablation on field_mse

tcapelle:
## Hypothesis

On the main tandemfoil benchmark, physics flags (`--enable-te-coord-frame`, `--enable-cp-panel`, `--enable-cp-panel-tandem-only`, `--asinh-pressure`, `--residual-prediction`, `--enable-pressure-prior-addition`) were crucial to reaching sota. But on tandemfoil_paper the metric is normalized full-field MSE over [Ux, Uy, p] — a fundamentally different target. We do not know whether pressure-specific flags (`--asinh-pressure`, `--enable-pressure-prior-addition`) or tandem-specific geometry flags (`--enable-cp-panel-tandem-only`) help or hurt on this metric.

Hypothesis: some physics flags may be neutral or harmful on the paper metric (which includes velocity error on equal footing with pressure). An ablation will show which flags contribute most.

## Current Baseline

**tandemfoil_paper: NO BASELINE — PR #2947 (jin) is establishing the first run simultaneously.**

Primary metric: `val_primary/field_mse` (normalized full-field MSE, lower is better).

Paper reference: MGN baseline ~1.79 (cruise_random_uniform). Paper best: 0.10.

## Experiment Plan

Start from the golden config (Lion lr=1e-4, 3L/192d/3H, T_max=10, gc=1.0, WD=1e-2, no-EMA, Fourier) and ablate physics flags systematically:

| Run | Physics flags removed |
|---|---|
| A | All flags (full golden config — reference) |
| B | Remove `--enable-pressure-prior-addition` |
| C | Remove `--asinh-pressure` |
| D | Remove `--residual-prediction` |
| E | Remove `--enable-cp-panel-tandem-only` |
| F | Remove `--enable-cp-panel` + `--enable-cp-panel-tandem-only` |
| G | Remove `--enable-te-coord-frame` |
| H | Remove ALL physics flags (bare Fourier only) |

## Instructions

**Run A (full golden config — reference):**
```bash
[...]

PR #3052, opened 2026-04-22
fix: 3 bugs blocking tandemfoil_paper training

tcapelle:
## Summary

- **AoA validation crash**: racecar samples legitimately have different AoA per foil — removed strict equality check in `split_paper_experiment4.py`
- **Corrupted pressure stats**: cruise_random samples contain ±inf pressure values from float16 overflow. Fixed `_compute_y_stats()` to use `torch.where(mask, yd, zeros)` instead of `yd * mask` (IEEE 754: `nan * 0 = nan`)
- **Inf propagation in TargetTransform.apply**: raw targets with ±inf propagated through normalization → NaN loss. Added `torch.where(out.isfinite(), out, zeros)` after normalization

These three bugs blocked all tandemfoil_paper training. Split out from #2948 as requested by advisor.

## Test plan

- [x] After fix, `split_stats_tandemfoil_paper_experiment4.json` generates with valid mean/std for all 6 tasks (no inf/nan)
- [x] Training runs produce valid `val_primary/field_mse` instead of NaN loss
- [x] Verified on 5 independent ablation runs through 35+ epochs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/target/icml2026/tandemfoil_paper/data/split_paper_experiment4.py
+++ b/target/icml2026/tandemfoil_paper/data/split_paper_experiment4.py
@@ -260,38 +260,32 @@ def _compute_y_stats(pickle_paths: list[Path], train_indices: list[int]) -> dict
     dataset = MultiFieldDataset(pickle_paths, cache_size=-1)
     train_sorted = sorted(train_indices, key=lambda idx: dataset.index[idx])
 
-    # FIX (Bug 2): some cruise_random samples contain ±inf pressure values from
-    # float16 overflow in the raw pickles.  Accumulate per-channel finite counts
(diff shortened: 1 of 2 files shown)


tcapelle:
ADVISOR: Thank you for splitting out these TFP bug fixes from #2948 as requested. The three fixes are clean and correct:

1. `torch.where(mask, yd, zeros)` for the IEEE 754 `nan * 0 = nan` bug in stats computation — textbook fix
2. inf-clamping in `TargetTransform.apply()` — reasonable safety net for float16 overflow in cruise_random pickles
3. Minor cleanup (redundant counter, .item() for scalar)

Merging this now. These fixes unblock all TFP experiments.

PR #3060, opened 2026-04-22
DrivAerML: Bilateral left-right symmetry augmentation

tcapelle:
## Hypothesis

Cars in zero-yaw flow have **exact bilateral (left-right) symmetry** along the longitudinal (xz) plane. Mirroring each training case — negating the y-coordinate and the y-component of surface normals, while leaving scalar Cp targets unchanged — produces a **physically valid, novel training example at zero compute cost**. This doubles the effective training set from 400 to 800 cases.

This is the 3D analogue of horizontal flip augmentation that routinely wins Kaggle competitions on image tasks. For DrivAerML with only 400 training cases, doubling the dataset is one of the highest-impact, lowest-risk changes possible.

**Prerequisite:** Verify that DrivAerML cars are axis-aligned with the symmetry plane at y=0 (or at the car centerline). If not, compute the centerline from the point cloud bounds.

**Note:** This is a DrivAerML-specific experiment, justified because bilateral symmetry is a physical property unique to the zero-yaw automotive flow domain. TandemFoil airfoils at different AoA do not have simple reflective symmetry.

## Instructions

**Step 1: Verify car symmetry axis**

Before implementing, check the DrivAerML data to confirm:
```python
# In the data loader or a debug script
for case in dataset:
    y_coords = case['pos'][:, 1]  # y-coordinate
    y_center = (y_coords.max() + y_coords.min()) / 2
    print(f"y center: {y_center}, y range: [{y_coords.min()}, {y_coords.max()}]")
```
If `y_center ≈ 0`, mirror about y=0. Otherwise, mirror about `y_center`.

**Step 2: Add `--drivaerml-symmetry-aug` flag to train.py**

```python
parser.add_argument('--drivaerml-symmetry-aug', action='store_true', default=False,
                    help='Mirror DrivAerML cases along y-axis (bilateral symmetry augmentation)')
```

**Step 3: Implement in the DrivAerML data loader (TRAINING ONLY)**

During training, with 50% probability per batch, flip the case:
```python
if args.drivaerml_symmetry_aug and self.training:
    if torch.rand(1).item() < 0.5:
        # Mirror y-coordinate
        pos[:, 1] = -pos[:, 1]  # (or 2*y_center - pos[:, 1] if not centered at 0)
        # Mirror y-component of surface normals if they exist as inputs
        if 'normals' in data:
            data['normals'][:, 1] = -data['normals'][:, 1]
        # Scalar targets (Cp / pressure) are UNCHANGED by reflection
```

**Key implementation details:**
- Apply ONLY during training (not validation/test)
- Flip with 50% probability (so model sees both original and mirrored)
- Only flip the y-coordinate and y-component of any vector features
- Scalar targets (pressure) are invariant under reflection — DO NOT modify targets
- If the model uses input features beyond xyz coords (e.g., surface normals), flip their y-component too

**Step 4: Run 2 DrivAerML experiments**

Use `--wandb_group levi-dm-symmetry-aug` for all runs.

```bash
cd target/icml2026

# Run 1: Symmetry augmentation ON
  --epochs 999 --batch-size 1 \
  --drivaerml-symmetry-aug --wandb_group levi-dm-symmetry-aug

# Run 2: Control (no augmentation — should reproduce baseline ~3.997%)
  --epochs 999 --batch-size 1 \
  --wandb_group levi-dm-symmetry-aug
```

**Results to report:**
- Best val `surface_rel_l2_pct` and epoch for both runs
- W&B run IDs
- Whether the augmented run improves over control
- If the symmetry check reveals cars are NOT axis-aligned, report the y-center distribution

## Baseline

| Dataset | Metric | Baseline | W&B run | Config |
|---------|--------|----------|---------|--------|
| DrivAerML | val_primary/surface_rel_l2_pct | **3.997%** | bht6h42t | 4L/512d, AdamW lr=5e-4, T_max=30, no gc, no WD, no-EMA |

Reproduce: `cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200`

### DIFF
--- a/.branch_init
+++ b/.branch_init
@@ -1,0 +1,1 @@
+# levi/dm-bilateral-symmetry-augmentation
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -168,6 +168,7 @@ class TrainConfig:
     grad_accum_steps: int = 1
     save_checkpoint: bool = False
     seed: int = 0
+    drivaerml_symmetry_aug: bool = False
 
 
 @dataclass
@@ -570,6 +571,32 @@ def autocast_context(device: torch.device, amp_mode: str):
     if not supports_bf16():
         return nullcontext()
     return torch.autocast(device_type="cuda", dtype=torch.bfloat16)
+
+
+def apply_drivaerml_symmetry_flip(batch: GroupedBatch, *, space_dim: int = 3) -> GroupedBatch:
+    surface_x = batch.surface_x.clone()
+    surface_x[:, :, 1] = -surface_x[:, :, 1]
+    surface_x[:, :, space_dim + 1] = -surface_x[:, :, space_dim + 1]
+    base_dim = 2 * space_dim + 1
+    feat_dim = surface_x.shape[-1]
+    if feat_dim > base_dim:
+        cols_per_freq = 2 * space_dim
+        n_freqs = (feat_dim - base_dim) // cols_per_freq
+        for i in range(n_freqs):
+            y_sin_col = base_dim + i * cols_per_freq + 1
+            surface_x[:, :, y_sin_col] = -surface_x[:, :, y_sin_col]
+    return GroupedBatch(
+        case_ids=batch.case_ids,
+        dataset_name=batch.dataset_name,
+        space_dim=batch.space_dim,
+        surface_x=surface_x,
+        surface_y=batch.surface_y,
+        surface_mask=batch.surface_mask,
+        volume_x=batch.volume_x,
+        volume_y=batch.volume_y,
+        volume_mask=batch.volume_mask,
+        metadata=batch.metadata,
+    )
 
 
 def float_outputs(outputs: dict[str, torch.Tensor | None]) -> dict[str, torch.Tensor | None]:
@@ -1266,11 +1293,13 @@ def train_one_epoch(
     max_batches: int = 0,
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
+    drivaerml_symmetry_aug: bool = False,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
         anp_head.train()
     running = {"loss": 0.0}
+    sym_aug_count = 0
     steps = 0
     micro_batches_total = 0
     batches = loader if max_batches <= 0 else itertools.islice(loader, max_batches)
@@ -1302,6 +1331,9 @@ def train_one_epoch(
                     loss, _ = loss_abupt(batch, outputs, transform)
             else:
                 batch = batch.to(device)
+                if drivaerml_symmetry_aug and batch.dataset_name == "drivaerml" and torch.rand(1).item() < 0.5:
+                    batch = apply_drivaerml_symmetry_flip(batch, space_dim=batch.space_dim)
+                    sym_aug_count += 1
                 if isinstance(transform, TandemTargetTransform):
                     prepared = transform.prepare_batch(batch)
                     with autocast_context(device, amp_mode):
@@ -1356,6 +1388,8 @@ def train_one_epoch(
         running["grad_norm_mean"] /= max(steps, 1)
     running["train_steps"] = float(steps)
     running["micro_batches"] = float(micro_batches_total)
+    if sym_aug_count > 0:
+        running["symmetry_aug_flips"] = float(sym_aug_count)
     if scheduler is not None:
         running["lr"] = float(optimizer.param_groups[0]["lr"])
     return running
@@ -1671,6 +1705,7 @@ def main() -> None:
             max_batches=config.max_train_batches,
             grad_clip=config.grad_clip,
             grad_accum_steps=config.grad_accum_steps,
+            drivaerml_symmetry_aug=config.drivaerml_symmetry_aug,
         )
 
         if ema is not None:


tcapelle:
STUDENT levi:

## Results

### Summary

| Run | Flag | Best Val `surface_rel_l2_pct` | Epoch | Test `surface_rel_l2_pct` | W&B Run ID |
|-----|------|-------------------------------|-------|---------------------------|------------|
| Aug ON | `--drivaerml-symmetry-aug` | **14.01%** | 50 | 14.51% | `d76aehil` |
| Control | (none) | **8.34%** | 75 | 9.16% | `e9sf9fwv` |
| **Baseline** | (none) | **3.997%** | 467 | — | `bht6h42t` |

**Verdict: Symmetry augmentation is harmful.** The augmented run's best (14.01%) is 68% worse than control's best (8.34%), and both fall well short of the baseline (3.997%) due to the 120-minute timeout limiting us to 167 epochs vs the baseline's 467.

### Detailed Analysis

**Training stability — Aug ON was catastrophically unstable:**
- **31 gradient norm spikes** (>5) across 167 epochs vs only **4 spikes** in the control
- Grad norms escalated from minor spikes (~8-9) at epochs 58-60 to terminal divergence (100-385) from epoch 148 onward
- The model never recovered after epoch 53 — val metric degraded monotonically from 14.0% to 37-64%
- By the final epoch, val was oscillating between 37-64% with grad norms consistently >100

**Training stability — Control was reasonably stable:**
- Only 4 spike epochs (1, 82, 119, 122), with epoch 122 being the worst (grad_norm=57.9)
- Recovered from all spikes and continued descending
- Best val 8.34% at epoch 75 (end of a cosine cycle), would likely continue improving toward baseline with more time

**Why symmetry augmentation hurt:**

1. **Cosine LR cycling amplification**: The cosine annealing schedule (T_max=30) already causes periodic instability at LR peaks. The symmetry-augmented batches inject additional distributional variance (50% of batches have flipped coordinates), which compounds with the LR peaks to create more frequent and more severe gradient explosions.

2. **Fourier feature sensitivity**: The model uses 4-frequency Fourier positional encodings on xyz. While the flip correctly negates sin(freq*y) terms (odd function) and preserves cos(freq*y) terms (even function), the augmentation effectively changes which Fourier basis functions are "positive" vs "negative" for surface points. This creates a harder optimization landscape — the model must learn features that are symmetric under both original and flipped Fourier encodings simultaneously.

3. **No gradient clipping**: Without gradient clipping, the instability from augmentation-induced variance has no safety net. The baseline config deliberately omits gradient clipping, which works fine for the vanilla training but becomes a liability when augmentation adds distributional noise.

### Commands Used

```bash
# Aug ON (GPU 0)
CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=120 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema \
  --epochs 999 --batch-size 1 \
  --drivaerml-symmetry-aug --wandb-group levi-dm-symmetry-aug \
  --agent levi --wandb-name "levi/dm-symmetry-aug-ON"

# Control (GPU 1)
CUDA_VISIBLE_DEVICES=1 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=120 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema \
  --epochs 999 --batch-size 1 \
  --wandb-group levi-dm-symmetry-aug \
  --agent levi --wandb-name "levi/dm-symmetry-aug-OFF-control"
```

### Peak Memory Usage

~5985 MiB per GPU (both runs identical)

### Symmetry Axis Verification

Confirmed all DrivAerML cases have y-center within ±0.001 of zero — safe to mirror about y=0 without recentering.

### Suggested Follow-ups

1. **Symmetry aug + gradient clipping**: The augmentation might work if paired with gradient clipping (max_norm=1.0) to contain the instability it introduces. The root failure mode is grad explosion, not the augmentation concept itself.
2. **Lower flip probability**: Try 10-20% instead of 50% to reduce distributional variance per epoch.
3. **Warmup-only augmentation**: Apply flips only during a warmup phase (first 10-20 epochs) then disable, giving the model a broader initial exploration without destabilizing later fine-tuning.
4. **Fixed doubled dataset**: Instead of stochastic flipping, pre-generate the mirrored copies as permanent dataset entries. This doubles the dataset deterministically without per-batch variance.

tcapelle:
ADVISOR: Closing PR #3060 because ADVISOR: Closing — bilateral symmetry augmentation is harmful on DrivAerML in current form. Aug ON: 14.01% best val (vs 8.34% control, 3.997% baseline). Stochastic per-batch flipping creates distributional variance that compounds gradient instability at T_max=30 LR peaks — 31 spike epochs vs 4 in control. Symmetry axis is valid (y-center within ±0.001) but the implementation needs gradient clipping or pre-generated mirrored copies as permanent dataset entries. Good systematic testing, levi..

PR #3030, opened 2026-04-22
Relative L2 training loss to align DrivAerML objective with eval metric

tcapelle:
## Hypothesis

The DrivAerML primary eval metric is `surface_rel_l2_pct` — a **relative** L2 error (normalized by target magnitude). However, training currently uses absolute MSE. This objective mismatch is a known source of degradation in surrogate regression: the model is penalized equally for errors on low-pressure and high-pressure regions, whereas the eval metric cares more about relative fidelity on each point.

Replacing MSE with a relative-L2 loss directly aligns the training signal with the evaluation signal and may also help TandemFoil because pressure variation across cases is large and absolute errors on high-Re regions dominate gradient updates.

Supporting reference: SMART paper (arxiv 2601.18707) shows relative-L2 objective alignment consistently outperforms MSE on surface regression tasks.

## Instructions

Add a `--loss` flag to `target/icml2026/train.py` supporting three modes:

- `mse` (default, existing behavior — do not change)
- `rel_l2`: replace training loss with `((pred - target).norm(dim=-1) / (target.norm(dim=-1) + 1e-8)).mean()`
- `mixed`: `0.5 * mse_loss + 0.5 * rel_l2_loss`

Run **four experiments** using `--wandb_group wave4-rel-l2-loss`:

### DrivAerML — two variants

**DM-rel_l2** — pure relative L2 loss:
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 --cosine-t-max 30 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
  --loss rel_l2 \
  --wandb_group wave4-rel-l2-loss
```

**DM-mixed** — 0.5 MSE + 0.5 rel_L2:
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 --cosine-t-max 30 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 \
  --batch-size 1 \
[...]

PR #3062, opened 2026-04-22
Cross-dataset: Per-case z-score output normalization

tcapelle:
## Hypothesis

The model currently trains on raw (or globally normalized) target values. Different cases have vastly different absolute pressure/velocity scales — high-drag DrivAerML cars produce much larger absolute pressure values than low-drag ones. This means the MSE loss is dominated by high-magnitude cases, and the model spends gradient budget learning scale rather than shape.

**Per-case z-score normalization** standardizes each case's target outputs individually (subtract case mean, divide by case std) before computing the loss, then inverts at prediction time. This:
1. Decouples shape learning from absolute scale
2. Gives equal gradient weight to all cases regardless of magnitude
3. Is the per-case generalization of the asinh pressure transform that produced a significant win on TandemFoil

**Key distinction from rel_l2 (#3030, CLOSED — catastrophic on DM):** rel_l2 normalized per-POINT (dividing by each point's target magnitude), which amplified noise at near-zero targets. Per-CASE normalization uses the case-level mean/std, avoiding the near-zero division issue entirely.

## Instructions

**Step 1: Add `--per-case-normalize-targets` flag to train.py**

```python
parser.add_argument('--per-case-normalize-targets', action='store_true', default=False,
                    help='Z-score normalize targets per-case before loss, invert at prediction')
```

**Step 2: Implement per-case normalization in training loop**

For each training batch:
```python
if args.per_case_normalize_targets:
    # Compute per-case statistics (over spatial dimension, per channel)
    # target shape: (N, C) for single-case batch or (B, N, C)
    target_mean = target.mean(dim=-2, keepdim=True)  # mean over spatial points
    target_std = target.std(dim=-2, keepdim=True).clamp(min=1e-6)  # avoid div by zero
    
    # Normalize targets
    target_normalized = (target - target_mean) / target_std
    
    # Compute loss on normalized targets
    loss = F.mse_loss(prediction_normalized, target_normalized)
```

**Step 3: Handle prediction inversion at eval time**

At evaluation, the model predicts in normalized space. To get back to physical space:
```python
if args.per_case_normalize_targets:
    # During eval, we need the case's mean/std to invert
    # Option A: predict normalized, then invert using ground-truth mean/std for metric computation
    prediction_physical = prediction_normalized * target_std + target_mean
    # Compute metrics in physical space
```

**IMPORTANT SUBTLETY:** At eval time, you have access to ground-truth targets (for metric computation). Use ground-truth case mean/std to invert predictions. This means the normalization affects TRAINING gradients but the METRICS are still computed in physical space — an apples-to-apples comparison with baseline.

**Step 4: Run 4 experiments (one per benchmark)**

Use `--wandb_group nezuko-percase-zscore` for all runs.

**DrivAerML — NO grad-clip, NO weight-decay:**
```bash
cd target/icml2026

  --epochs 999 --batch-size 1 \
  --per-case-normalize-targets --wandb_group nezuko-percase-zscore
```

**AirfRANS:**
```bash
  --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 \
  --per-case-normalize-targets --wandb_group nezuko-percase-zscore
```

**TandemFoil:**
```bash
python train.py --dataset tandemfoil \
  --optimizer lion --lr 1.25e-4 --cosine-t-max 10 --grad-clip 0.5 \
  --weight-decay 1e-2 --model-slices 64 --model-layers 3 \
  --model-hidden-dim 192 --model-heads 3 \
  --enable-pressure-prior-addition --epochs 999 --ema-decay 0.999 \
  --per-case-normalize-targets --wandb_group nezuko-percase-zscore
```

**TandemFoil Paper:**
```bash
python train.py --dataset tandemfoil_paper \
  --optimizer lion --lr 1.25e-4 --cosine-t-max 10 --grad-clip 0.5 \
  --weight-decay 1e-2 --enable-fourier --model-layers 3 \
  --model-hidden-dim 192 --model-heads 3 \
  --per-case-normalize-targets --wandb_group nezuko-percase-zscore
```

**Important notes for TandemFoil:** The TF pipeline already uses `--asinh-pressure` (asinh transform on pressure targets). Per-case normalization should be applied AFTER the asinh transform — normalize the post-transform values. Check the order of operations carefully.

**Results to report:**
- Best val primary metric and epoch for each dataset
- W&B run IDs
- Training loss curves — does per-case normalization produce smoother training?
- Per-split breakdown for TandemFoil (does OOD generalization improve specifically?)

## Baseline

| Dataset | Metric | Baseline | W&B run |
|---------|--------|----------|---------|
| DrivAerML | val_primary/surface_rel_l2_pct | **3.997%** | bht6h42t |
| AirfRANS | val_primary/surface_mse | **0.000482** | pr4wsbfm |
| TandemFoil | val_primary/surface_pressure_mae | **22.537** | 0lv7fnun |
| TandemFoil Paper | val_primary/field_mse | **0.002383** | d1xh0o1p |

### DIFF
--- a/.branch_init
+++ b/.branch_init
@@ -1,0 +1,1 @@
+# nezuko/per-case-zscore-normalization


morganmcg1:

PR #3069, opened 2026-04-22
DrivAerML: 5-epoch linear warmup + cosine T_max=30

tcapelle:
## Hypothesis

The DrivAerML champion uses cosine annealing starting from full LR at epoch 0. A linear warmup ramp in early epochs could prevent cold-start instability and find a better initial basin. 5-epoch warmup is gentle enough to not waste budget.

## Instructions

Add a 5-epoch linear warmup before cosine annealing. Implement as follows:
- For epochs 0-4: use `lr = base_lr * (epoch + 1) / 5` (linear ramp from lr/5 to lr)
- For epochs 5+: standard cosine annealing with T_max=30
- Simplest implementation: `torch.optim.lr_scheduler.LinearLR(optimizer, start_factor=0.2, end_factor=1.0, total_iters=5)` for first 5 epochs, then switch to `CosineAnnealingLR(optimizer, T_max=30)`
- Or use `SequentialLR([LinearLR(...), CosineAnnealingLR(...)], milestones=[5])`
- Add a `--warmup-epochs` flag (default 0)

```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 \
  --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 \
  --model-heads 8 --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
  --warmup-epochs 5 \
  --wandb_group dm-warmup-sweep
```

## Reporting Requirements
- Best val `surface_rel_l2_pct` + epoch
- Test `surface_rel_l2_pct` at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML baselines:**
- val best: `3.997%` @ ep467 (PR #2898, W&B: `bht6h42t`, 4L/512d, AdamW lr=5e-4, T_max=30)
- test best: `6.244%` (PR #2648, 4L/320d, truncated eval — OLDER CONFIG)
- External targets: AB-UPT 3.82% | Transolver-3: 3.71%

```bash
# DrivAerML champion (val 3.997% @ ep467)
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 \
[...]

PR #3070, opened 2026-04-22
DrivAerML: 10-epoch linear warmup + cosine T_max=30

tcapelle:
## Hypothesis

10-epoch warmup provides a more gradual ramp than 5-epoch. AB-UPT uses Lion with a warmup+cosine schedule for DrivAerML.

## Instructions

Add a 10-epoch linear warmup before cosine annealing:
- Implement `SequentialLR([LinearLR(optimizer, start_factor=0.1, end_factor=1.0, total_iters=10), CosineAnnealingLR(optimizer, T_max=30)], milestones=[10])`
- Add `--warmup-epochs 10` flag

```bash
  --warmup-epochs 10 \
  --wandb_group dm-warmup-sweep
```

## Reporting Requirements
- Best val + test `surface_rel_l2_pct`
- W&B run ID

## Baseline

**DrivAerML baselines:**

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -160,6 +160,7 @@ class TrainConfig:
     residual_prediction: bool = False
     re_stratified_sampling: bool = False
     cosine_t_max: int = 150
+    warmup_epochs: int = 0
     geometry_points: int = 25_000
     geometry_supernodes: int = 4_096
     surface_anchor_points: int = 8_000
@@ -1625,7 +1626,18 @@ def main() -> None:
     scheduler = None
     if config.cosine_t_max > 0:
         base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
-        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        cosine_scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        if config.warmup_epochs > 0:
+            steps_per_epoch = config.max_train_batches if config.max_train_batches > 0 else len(train_loader)
+            warmup_steps = config.warmup_epochs * steps_per_epoch
+            warmup_scheduler = torch.optim.lr_scheduler.LinearLR(
+                base_optimizer, start_factor=0.1, end_factor=1.0, total_iters=warmup_steps,
+            )
+            scheduler = torch.optim.lr_scheduler.SequentialLR(
+                base_optimizer, schedulers=[warmup_scheduler, cosine_scheduler], milestones=[warmup_steps],
+            )
+        else:
+            scheduler = cosine_scheduler
 
     ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
     anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT chrome:

## Results

### 10-epoch linear warmup + cosine T_max=30 (DrivAerML)

| Metric | Baseline | This Experiment |
|--------|----------|----------------|
| val best `surface_rel_l2_pct` | **3.997%** @ ep467 | **6.225%** @ ep105 |
| test `surface_rel_l2_pct` | 6.244% (older config) | N/A (diverged before test eval) |
| External targets | AB-UPT 3.82% \| Transolver-3 3.71% | — |

**W&B run ID:** `w1c7l3jo`
**Peak memory:** ~6.0 GB (GPU 0)

**Command:**
```bash
  --warmup-epochs 10 \
  --wandb-group dm-warmup-sweep \
  --agent chrome --wandb-name "chrome/dm-warmup-10ep-cosine30-v2"
```

### What happened

The 10-epoch linear warmup (start_factor=0.1 → 1.0) was implemented using `SequentialLR([LinearLR, CosineAnnealingLR])`. The scheduler steps per batch (not per epoch), so warmup iters were scaled to `warmup_epochs * steps_per_epoch = 10 * 394 = 3940` steps.

**Initial run (v1)** had a bug: warmup `total_iters` and milestone were set to epoch counts (10) instead of batch-step counts. Since the scheduler steps per batch, the warmup was consumed in the first 10 batches of epoch 1, and the cosine oscillated wildly — training diverged after epoch ~130. V1 best val: 15.02%.

**Fixed run (v2)** correctly scaled warmup to batch steps. Training converged well:
- Warmup epochs 1–10: smooth LR ramp from 5e-5 → 5e-4, val improved from 63% → 24%
- Cosine phase epochs 11–105: steady improvement to 6.225% (new bests at every ~10 epochs)
- **Catastrophic divergence at epoch 114:** val spiked from 10.8% → 81.9%, grad norms from 0.4 → 8.8. Training never recovered — oscillated chaotically between 30–52% from ep115–124 before I killed it.

The divergence is likely caused by the combination of: (1) batch_size=1 with no gradient clipping (`grad_clip=0.0`), and (2) the rapid cosine oscillation (T_max=30 in batch-steps = ~0.076 epochs per half-cycle). A single bad gradient update at a LR peak can push the model into an unstable regime.

**Verdict: The warmup itself worked correctly and showed good early convergence, but the result (6.225%) does not beat the baseline (3.997%).** The run was cut short by divergence at epoch 114 — the baseline reached 3.997% at epoch 467, so this run didn't have enough stable training time. The warmup may still be beneficial if combined with gradient clipping to prevent the catastrophic divergence.

### Suggested follow-ups
- **Add gradient clipping** (e.g. `--grad-clip 1.0`) to prevent catastrophic divergence events — this would let warmup runs train stably to the 400+ epochs needed to reach baseline territory
- **Increase cosine T_max** — T_max=30 in batch-steps gives very rapid cycling (~13 full cosine periods per epoch). A larger T_max might give smoother training dynamics
- Try **CosineAnnealingWarmRestarts** with proper per-epoch stepping for more controlled LR decay

tcapelle:
ADVISOR: Closing PR #3070 because ADVISOR: Closing — 10-epoch warmup without gc diverged at ep114. Best val 6.225% at ep105 (+56% above 3.997% baseline). The warmup itself worked correctly (63%→24% in epochs 1-10), but the run still hit catastrophic gradient explosion at a cosine LR peak. Without gc, the warmup contribution is overwhelmed by the same instability that kills all DM non-champion configs. The companion #3069 (5-epoch warmup) likely shows the same pattern. The real test is warmup + gc combined (historia #3119 tests higher LR + 20-epoch warmup). Good diagnosis, chrome..

PR #3074, opened 2026-04-22
DrivAerML: relative L2 training loss (align with eval metric)

tcapelle:
## Hypothesis

DrivAerML trains on absolute MSE but evaluates on `surface_rel_l2_pct` — a relative (normalized) L2. This objective mismatch means gradient signal is dominated by high-magnitude cases, but the evaluation weights all cases equally. Using relative L2 during training should align gradient signal with the evaluation criterion.

**Formula:** For each node: `rel_l2 = ||pred - target||_2 / (||target||_2 + eps)`. Mean over nodes and cases.

## Instructions

Implement `--loss rel_l2` option in train.py. When active, replace the MSE loss with:

```python
def relative_l2_loss(pred, target, eps=1e-8):
    # pred, target: [N, C] or [N] per case
    num = (pred - target).norm(dim=-1)  # per-node L2 error
    denom = target.norm(dim=-1) + eps    # per-node target norm
    return (num / denom).mean()
```

For DrivAerML, `target` is `surface_cp` (1 channel). Apply normalization per node.

Also try: `--loss mixed` = 0.5 * MSE + 0.5 * rel_l2.

Run two variants:
```bash
# Variant 1: Pure relative L2
  --loss rel_l2 --wandb_group dm-loss-alignment

# Variant 2: Mixed 50/50
  --loss mixed --wandb_group dm-loss-alignment
```

**Divergence check:** If rel_l2 diverges before epoch 20, add `--grad-clip 1.0` and document.

## Reporting Requirements
- Best val `surface_rel_l2_pct` for each variant
- Test at best val checkpoint
- W&B run IDs for both variants

## Baseline

**DrivAerML baselines:**

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -166,6 +166,7 @@ class TrainConfig:
     volume_anchor_points: int = 8_000
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
+    loss: str = "mse"
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -757,21 +758,36 @@ def _named_channel_metrics(prefix: str, values: torch.Tensor, names: tuple[str, 
     return metrics
 
 
+def _relative_l2_loss(pred: torch.Tensor, target: torch.Tensor, eps: float = 1e-8) -> torch.Tensor:
+    num = (pred - target).norm(dim=-1)
+    denom = target.norm(dim=-1) + eps
+    return (num / denom).mean()
+
+
+def _compute_loss(pred: torch.Tensor, target: torch.Tensor, loss_type: str) -> torch.Tensor:
+    if loss_type == "rel_l2":
+        return _relative_l2_loss(pred, target)
+    if loss_type == "mixed":
+        return 0.5 * F.mse_loss(pred, target) + 0.5 * _relative_l2_loss(pred, target)
+    return F.mse_loss(pred, target)
+
+
 def loss_grouped(
     batch: GroupedBatch,
     outputs: dict[str, torch.Tensor | None],
     transform: TargetTransform,
+    loss_type: str = "mse",
 ) -> tuple[torch.Tensor, dict[str, float]]:
     total = torch.tensor(0.0, device=batch.surface_x.device)
     metrics: dict[str, float] = {}
     if batch.surface_y is not None and outputs["surface_preds"] is not None:
         target = transform.apply(batch.surface_y)
-        surf_loss = F.mse_loss(outputs["surface_preds"][batch.surface_mask], target[batch.surface_mask])
+        surf_loss = _compute_loss(outputs["surface_preds"][batch.surface_mask], target[batch.surface_mask], loss_type)
         total = total + surf_loss
         metrics["surface_loss"] = float(surf_loss.detach().cpu().item())
     if batch.volume_y is not None and outputs["volume_preds"] is not None and batch.volume_mask is not None:
         target = transform.apply(batch.volume_y)
-        vol_loss = F.mse_loss(outputs["volume_preds"][batch.volume_mask], target[batch.volume_mask])
+        vol_loss = _compute_loss(outputs["volume_preds"][batch.volume_mask], target[batch.volume_mask], loss_type)
         total = total + vol_loss
         metrics["volume_loss"] = float(vol_loss.detach().cpu().item())
     metrics["loss"] = float(total.detach().cpu().item())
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT fern:

## Results

### Summary

Both relative L2 loss variants **failed catastrophically** — the model gets trapped at a constant-prediction plateau (~75% rel_l2_pct) and never learns, while the baseline MSE loss reaches 3.997%.

### Variant 1: Pure Relative L2 Loss (`--loss rel_l2`)

| Metric | Value |
|--------|-------|
| Best val `surface_rel_l2_pct` | **75.27%** @ ep138 |
| Test `surface_rel_l2_pct` (best val ckpt) | **76.21%** |
| Final test `surface_rel_l2_pct` | 76.38% |
| Epochs completed | 505 |
| Peak memory | ~6.2 GB |
| W&B run ID | `lago3za1` |

**Command:**
```bash
CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=999 SENPAI_TIMEOUT_MINUTES=360.0 python train.py \
  --loss rel_l2 --wandb-group dm-loss-alignment \
  --agent fern --wandb-name "fern/dm-rel-l2-loss"
```

### Variant 2: Mixed 50/50 Loss (`--loss mixed`)

| Metric | Value |
|--------|-------|
| Best val `surface_rel_l2_pct` | **75.30%** @ ep328 |
| Test `surface_rel_l2_pct` (best val ckpt) | **76.30%** |
| Final test `surface_rel_l2_pct` | 76.37% |
| Epochs completed | 498 |
| Peak memory | ~6.0 GB |
| W&B run ID | `cxrfryjm` |

**Command:**
```bash
CUDA_VISIBLE_DEVICES=1 SENPAI_MAX_EPOCHS=999 SENPAI_TIMEOUT_MINUTES=360.0 python train.py \
  --loss mixed --wandb-group dm-loss-alignment \
  --agent fern --wandb-name "fern/dm-mixed-loss"
```

### Baseline Comparison

| Variant | Best val rel_l2_pct | Best test rel_l2_pct |
|---------|--------------------|--------------------|
| **Baseline MSE** (PR #2898) | **3.997%** @ ep467 | **6.244%** |
| Pure rel_l2 | 75.27% @ ep138 | 76.21% |
| Mixed 50/50 | 75.30% @ ep328 | 76.30% |

### What happened

**Both variants are ~19x worse than the baseline.** The model converges to a near-constant prediction (the z-normalized mean) within the first 5 epochs and stays trapped there for the remaining 500 epochs. Training loss does decrease, but the validation metric never improves beyond ~75%.

**Root cause — relative L2 on z-normalized targets creates a degenerate loss landscape:**

1. The training loss operates on z-normalized targets (mean≈0, std≈1). Many surface nodes have z-target ≈ 0.
2. The relative L2 formula `|pred - target| / (|target| + eps)` with eps=1e-8 creates enormous gradients for near-zero targets — effectively infinite weight on these nodes.
3. The model quickly learns to predict near-zero everywhere (the z-space mean), which minimizes the huge penalties from near-zero denominator nodes.
4. Once trapped at this "predict the mean" solution, there's no gradient signal to escape — the relative error is uniformly ~75% across all cases.
5. Even the mixed loss (50% MSE + 50% rel_l2) suffers: the rel_l2 component dominates the gradient landscape and prevents the MSE term from driving learning.

**Divergence check:** No NaN/Inf divergence occurred. The rel_l2 training loss was very noisy (oscillating 1.0-17.0) with high grad norms (300-700), but technically stable. Grad clipping was not added since there was no divergence per the PR criterion, though the model simply failed to learn rather than diverging.

### Suggested follow-ups

1. **Apply relative L2 in raw (un-normalized) space** — compute the loss after `transform.invert()`, matching how the eval metric is computed. This avoids the near-zero denominator problem since raw Cp values are typically O(1).
2. **Use a much larger eps** (e.g., eps=1.0 or eps=mean(|target|)) to prevent near-zero targets from dominating.
3. **Curriculum approach** — start with MSE and gradually blend in rel_l2 after the model has learned reasonable predictions, avoiding the early constant-prediction trap.
4. **Log-cosh relative loss** — use `log(cosh(rel_error))` instead of raw `rel_error` to bound the gradient contribution from near-zero denominator nodes.

tcapelle:
ADVISOR: Closing PR #3074 because ADVISOR: Both pure rel_l2 (75.27%) and 50/50 mixed (75.30%) immediately converge to the z-normalized mean and plateau there for the entire run — a degenerate failure. Root cause: relative L2 on z-normalized targets creates infinite-weight gradients at near-zero nodes, driving the model to predict zero. The loss formulation and feature normalization are fundamentally incompatible as-is. The fix (compute rel_l2 in raw un-normalized space) would require a different implementation and is a distinct hypothesis. Closing..

PR #3077, opened 2026-04-22
DrivAerML: 5L/512d depth scaling (one more layer)

tcapelle:
## Hypothesis

Current champion is 4L/512d. Adding one more layer (5L) at the same width increases model capacity without requiring higher memory. For 3D automotive surface pressure, extra depth may help model more complex flow separation patterns (e.g., rear window flow, underbody).

## Instructions

```bash
  --model-layers 5 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --wandb_group dm-arch-depth
```

**Safety:** If diverges before epoch 30, add `--grad-clip 0.5` and document.

## Reporting Requirements
- Test at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML baselines:**


tcapelle:
STUDENT gilbert:

## Results

### Run 1: 5L/512d no grad-clip (W&B: `l4zzcqp5`)
- **Terminated early** — piping error cut training at epoch 41
- Observed divergence at epoch 23 (val spiked 21.2% → 45.7% at cosine restart), recovered by epoch 27
- Best val: 16.35% at epoch 38 (only 41 epochs, not meaningful for comparison)
- **Triggered safety instruction** to add `--grad-clip 0.5` since divergence occurred before epoch 30

### Run 2: 5L/512d + gc=0.5 (W&B: `ox918q66`) ← Primary result
- **Best val `surface_rel_l2_pct`:** 4.172% at epoch 347
- **Best test `surface_rel_l2_pct`:** 5.133% at epoch 347
- **Total epochs:** 523 (360-min budget)
- **Catastrophic divergence at epoch ~405:** Grad norms escalated (1220 → 4982 → 18764 → 84829 → 1.57M → 242M → ∞). Model never recovered — val stuck at ~58% for remaining 118 epochs.

### Comparison to baseline

| Config | Val best | Epoch | Test | Diverged? |
|--------|----------|-------|------|-----------|
| **4L/512d baseline** (PR #2898) | **3.997%** | 467 | — | No (467 clean epochs) |
| **5L/512d + gc=0.5** (this PR) | 4.172% | 347 | 5.133% | Yes (ep405) |
| External: AB-UPT | 3.82% | ~500 | — | — |
| External: Transolver-3 | 3.71% | — | — | — |

**Delta:** +4.4% relative worse than baseline (4.172% vs 3.997%)

### Command used
```bash
  --model-layers 5 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --grad-clip 0.5 \
  --wandb-group dm-arch-depth \
  --agent gilbert --wandb-name "gilbert/dm-5l-512d-gc05"
```

### What happened

**The hypothesis is falsified.** 5L/512d does not beat the 4L/512d champion. Two problems:

1. **No capacity benefit:** At 347 clean epochs, the 5L model reached 4.172% vs the 4L model's 3.997% at 467 epochs. Even accounting for the fewer clean epochs, the trajectory suggests the 5L model was unlikely to close the gap — it was already showing signs of instability by ep400.

2. **Training instability:** The extra layer makes the model more susceptible to catastrophic divergence. The 4L baseline trained 467 epochs cleanly; the 5L model diverged at ep405 despite gc=0.5 stabilization. The divergence was irreversible — grad norms went to infinity and never recovered.

The depth-instability tradeoff appears unfavorable at 512d width. The existing 4L/512d architecture seems to be at or near the optimal depth for this width and dataset.

### Suggested follow-ups

- **5L/512d + gc=1.0:** Tighter clipping might prevent the ep405 divergence and allow longer training — the 4L baseline used no grad-clip, but 5L may need it
- **5L/512d + lower LR (3e-4 or 4e-4):** Reducing LR could stabilize the deeper model while preserving capacity benefit
- **5L/448d:** Reduce width slightly to offset the parameter increase from the extra layer — may find a better depth/width balance

tcapelle:
ADVISOR: Closing PR #3077 because ADVISOR: 5L/512d reaches 4.172% at ep347 then diverges catastrophically at ep405 (grad norms →∞, never recovers). 4.4% worse than 3.997% baseline before divergence. Pattern is clear: adding a 5th layer at this width creates instability without accuracy benefit. The 4L/512d architecture is confirmed as the optimal depth for DrivAerML. Closing — 5L direction dead end..

PR #3078, opened 2026-04-22
DrivAerML: 6L/512d depth scaling (two more layers)

tcapelle:
## Hypothesis

Aggressive depth increase from 4L to 6L at 512d. DrivAerML is a complex 3D surface prediction problem — more depth may capture multi-scale flow features (global pressure distribution + local boundary layer structure).

## Instructions

```bash
  --model-layers 6 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --wandb_group dm-arch-depth
```

**Safety:** If diverges before epoch 30, add `--grad-clip 0.5` and document.

## Reporting Requirements
- Best val `surface_rel_l2_pct`
- Test at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML baselines:**


tcapelle:
STUDENT gohan:

## Results

### Run 1: 6L/512d without grad-clip (W&B: `ca8rh3q8`)

| Metric | Value |
|--------|-------|
| **Best val `surface_rel_l2_pct`** | **14.680%** at epoch 33 |
| Total epochs | 39 (killed — diverged at ep37) |

**Command:**
```bash
  --model-layers 6 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --wandb-group dm-arch-depth \
  --agent gohan --wandb-name "gohan/dm-6l-512d-deeper"
```

The model trained well through ep33 with a clear downward trough envelope (64%→48%→32%→25%→18%→15%), but at epoch 37 (LR peak at 4.95e-4), val spiked to 54.3% with loss jumping 10x (0.051→0.578) and grad norms tripling (1.2→3.2). By ep39, val hit 77.2% — catastrophic divergence. The 6L model is more sensitive to gradient explosions at cosine LR peaks than 4L due to the deeper gradient path.

### Run 2: 6L/512d with grad-clip 0.5 (W&B: `9cccs3ze`)

| Metric | Value |
|--------|-------|
| **Best val `surface_rel_l2_pct`** | **6.372%** at epoch 155 |
| Total epochs | 174 (killed — slow divergence from ep161) |
| Best loss | 0.0114 |
| Best grad norm | 0.70 |
| Baseline (4L/512d) | **3.997%** at epoch 467 |
| External target (AB-UPT) | 3.82% |

**Command:**
```bash
  --model-layers 6 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --grad-clip 0.5 \
  --wandb-group dm-arch-depth \
  --agent gohan --wandb-name "gohan/dm-6l-512d-gc05"
```

Grad-clip 0.5 prevented the early divergence (sailed through ep37 where the unclipped run died) and allowed training to reach ep155 with a best of 6.37%. The trough envelope descended smoothly: 37.7%→30.4%→25.1%→20.2%→16.1%→14.3%→12.0%→10.5%→9.4%→8.4%→7.3%→6.5%→6.4%.

However, at epoch 161, a sustained instability event occurred: loss jumped from 0.010 to 0.025, grad norms spiked from 0.4 to 6.5, and val climbed to 14.4%. Unlike the minor perturbation at ep154 (which recovered in 1 epoch), this one escalated — by ep174, grad norms had risen to 13.0 with no recovery. The model was slowly diverging despite grad-clip, so I killed the run.

## What happened

**The hypothesis is falsified.** 6L/512d (6.37% best) does not beat the 4L/512d baseline (3.997%). The additional depth introduces training instability that limits how far the model can optimize:

1. **Instability at depth:** The 6L model is significantly more prone to gradient explosions at cosine LR peaks. Without grad-clip, it diverges at ep37. With gc=0.5, it survives to ep155 but eventually succumbs to a sustained instability event.
2. **Faster early convergence, worse ceiling:** The 6L model converged faster than 4L in early epochs (e.g., 9.8% at ep70 vs 4L needing ~150 epochs for similar). But the instability ceiling prevents it from training long enough to reach the deep basins 4L finds at ep400+.
3. **Depth vs stability trade-off:** For DrivAerML with cosine T_max=30 and lr=5e-4, 4 layers appears to be the depth sweet spot — deep enough for capacity, stable enough for long training.

## Suggested follow-ups

- **6L/512d with gc=1.0:** Tighter grad-clip might extend stability further (gc=0.5 was borderline)
- **6L/512d with lower LR (3e-4):** Reducing LR could prevent the gradient explosions while preserving depth capacity
- **5L/512d:** An intermediate depth (already being tested by gilbert in #3077) could find the sweet spot between capacity and stability
- **6L/512d + EMA:** EMA weight averaging could preserve the best checkpoint through instability events, extracting more value from the early training

tcapelle:
ADVISOR: Closing PR #3078 because ADVISOR: Closing — 6L/512d is a dead end on DrivAerML. Without gc: 14.68% best at ep33, catastrophic divergence ep37. With gc=0.5: 6.372% best at ep155, slow divergence ep161. Both well above 3.997% baseline. The monotonic depth trend (2L<3L<4L) does NOT continue to 6L — deeper gradient paths amplify LR-peak explosions, and even gc=0.5 cannot stabilize enough for the 400+ epochs needed. 4L is confirmed as the DM depth sweet spot. Waiting for #3077 (5L) results to fully close the depth frontier. Excellent comparative analysis, gohan..

PR #3083, opened 2026-04-22
DrivAerML: max-train-batches=600 (53% more data/epoch)

tcapelle:
## Hypothesis

Champion uses 394 batches/epoch. 600 batches/epoch means each epoch sees 53% more car configurations. This increases data diversity per epoch but reduces epoch count in the time budget. Previous test at 4L/256d (#2645) with 600 batches showed improvement.

## Instructions

```bash
  --max-train-batches 600 --max-eval-batches 200 \
  --wandb_group dm-throughput-sweep
```

Note: With 4L/512d, each batch is slower than with 4L/256d. Budget may yield fewer total epochs.

## Reporting Requirements
- Total epochs run within time budget
- Test at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML baselines:**


tcapelle:
STUDENT jet:

## Results

### Summary

**The hypothesis did not work.** 600 batches/epoch with 4L/512d is significantly worse than the champion 394 batches/epoch, and the training diverged to NaN.

### Metrics

| Metric | This run (600 batches) | Baseline (394 batches) |
|--------|----------------------|----------------------|
| Best val surface_rel_l2_pct | **9.537%** (epoch 48) | **3.997%** (epoch 467) |
| Best test surface_rel_l2_pct | 10.076% | 6.244% |
| Total epochs | 392 (diverged ~epoch 300) | 467+ |
| External targets | AB-UPT 3.82% / Transolver-3 3.71% | — |

### Command used

```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 CUDA_VISIBLE_DEVICES=0 python train.py \
  --max-train-batches 600 --max-eval-batches 200 \
  --agent jet --wandb-name "jet/dm-max-train-batches-600" \
  --wandb-group dm-throughput-sweep
```

### Peak memory
~6 GB VRAM (GPU 0)

### W&B run
[`s4kvc8oz`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/s4kvc8oz)

### What happened

Training started reasonably (best val 9.537% at epoch 48) but never approached the baseline of 3.997%. The model then diverged to NaN around epoch ~300, with all subsequent epochs producing NaN loss, gradients, and metrics. The LR appeared stuck at the maximum (5e-4) during the NaN phase because the cosine schedule was cycling but having no effect on dead weights.

The 600 batches/epoch at 4L/512d scale creates two compounding problems:
1. **Each epoch takes ~52% longer** than with 394 batches, so fewer total cosine cycles fit in the time budget (392 epochs vs 467+ baseline). With T_max=30, the model got ~13 cosine cycles vs ~15+ for the baseline.
2. **No gradient clipping** (the command doesn't specify `--grad-clip`) — at higher data volume per epoch, gradient magnitudes may accumulate more, making training less stable.

The previous success with 600 batches at 4L/256d (#2645) may not transfer to 4L/512d because the larger model has more parameters to destabilize.

### Suggested follow-ups
- Try 600 batches with `--grad-clip 0.5` or `--grad-clip 1.0` to prevent the NaN divergence — the larger batch count may simply need gradient clipping to stay stable at 4L/512d scale.
- Consider scaling T_max proportionally with batch count: if 394 batches uses T_max=30, then 600 batches could use T_max≈46 to maintain the same wall-clock cycle length.
- The 4L/256d success at 600 batches suggests this is a model-size × data-volume interaction, not a pure data-volume issue.

tcapelle:
ADVISOR: Closing PR #3083 because ADVISOR: 9.537% ep48, 2.5x worse than 3.833% baseline, diverged to NaN ep266 (W&B: s4kvc8oz). 600 batches at 4L/512d WITHOUT gradient clipping is fundamentally unstable. Your diagnosis is spot on — gc=0.5+EMA is mandatory for DM stability, and 600 batches needs T_max scaling to ~46. Will reassign with proper config (gc=0.5+EMA+T_max=46)..

PR #2910, opened 2026-04-21
DrivAerML: max-train-batches=788 (2x) + gc=1.0 — more data per epoch

tcapelle:
## Hypothesis

The DrivAerML golden config uses `--max-train-batches 394`, meaning it sees 394 car configurations per epoch. The dataset has ~394 training cases (1 batch/case at batch-size=1), so 394 batches = 1 epoch of full data. Increasing to 788 batches (2x) samples with replacement and forces the model to see each case twice per epoch — effectively doubling data augmentation via resampling. A previous experiment at 4L/256d with 600 batches (#2645) showed promise. Now at 4L/512d with gc=1.0 recipe transfer, the wider model should benefit more from additional data.

## Instructions

**IMPORTANT: Use EXACTLY these run specifications.**

Run 1 — DrivAerML 788 batches/epoch + gc=1.0:
```bash
cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw \
  --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --grad-clip 1.0 \
  --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 788 --max-eval-batches 200 \
  --wandb-group eren-batches-sweep
```

Run 2 — DrivAerML 600 batches/epoch + gc=1.0 (intermediate):
```bash
cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw \
  --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --grad-clip 1.0 \
  --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 600 --max-eval-batches 200 \
  --wandb-group eren-batches-sweep
```

Run 3 — DrivAerML 788 batches + gc=1.0 + WD=1e-2 (compound):
```bash
cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw \
  --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --grad-clip 1.0 --weight-decay 1e-2 \
[...]

PR #3051, opened 2026-04-22
DrivAerML: 4x case revisits per epoch (max-train-batches=1576)

tcapelle:
## Hypothesis

Human directive #3020 mentions "revisit cases multiple times per epoch (Transolver-3 style, ~400k points/case/epoch)." Currently with 394 cases and 50k points, each case is seen once per epoch (394 batches). With max-train-batches=1576 (4x), each case is revisited ~4 times per epoch with DIFFERENT random 50k-point subsamples each time. This is data augmentation via resampling — the model sees different spatial subsets of each car geometry each epoch. Eren #2910 tested 2x (788 batches). This tests 4x and 8x — more aggressive coverage, potentially finding deeper loss basins through better gradient coverage over the full surface.

## Current Baseline (DM)
- val_primary/surface_rel_l2_pct = **3.997%** (AdamW lr=5e-4, T_max=30, no-EMA, 4L/512d, NO gc, NO WD)
- External target: 3.71%

## Runs

**Run 1 — 4x revisits (max-train-batches=1576)**
```bash
SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 1576 --max-eval-batches 200 --wandb_group dm_multi_revisit
```

**Run 2 — 8x revisits (max-train-batches=3152, Transolver-3 territory)**
```bash
SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 3152 --max-eval-batches 200 --wandb_group dm_multi_revisit
```

**Run 3 — 1x control with full eval (paper-facing, no --max-eval-batches)**
```bash
[...]

PR #3084, opened 2026-04-22
DrivAerML: max-train-batches=788 (2x data/epoch)

tcapelle:
## Hypothesis

788 batches/epoch = 2x more car configurations per epoch than baseline (394). Each car configuration is seen twice per epoch instead of once. This tests whether more data diversity per epoch improves convergence, at the cost of fewer total epochs.

## Instructions

```bash
  --max-train-batches 788 --max-eval-batches 200 \
  --wandb_group dm-throughput-sweep
```

## Reporting Requirements
- Total epochs run within time budget
- Test at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML baselines:**

Note: eren #2910 tested 788 batches at earlier config — results were unclear due to epoch budget. This tests at the current 4L/512d champion.


tcapelle:
STUDENT kakashi:

## Results

### DrivAerML: max-train-batches=788 (2x data/epoch)

| Metric | This run | Baseline (PR #2898) | External targets |
|--------|----------|---------------------|------------------|
| Val best `surface_rel_l2_pct` | **11.57%** @ ep42 | **3.997%** @ ep467 | AB-UPT 3.82% \| Transolver-3 3.71% |
| Test `surface_rel_l2_pct` | **12.12%** | 6.244% (older config) | — |
| Total useful epochs | 128 (NaN at ep129) | 467 | — |
| Total epochs in budget | 321 (192 NaN) | 467 | — |

**Verdict: NEGATIVE.** 788 batches/epoch significantly underperforms the 394-batch baseline.

### Command
```bash
  --max-train-batches 788 --max-eval-batches 200 \
  --wandb-group dm-throughput-sweep \
  --agent kakashi --wandb-name "kakashi/dm-train-batches-788"
```

### W&B
- **Run ID:** `ndxdsnkb`
- **Run URL:** https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/ndxdsnkb

### What happened

The core issue is that **788 batches/epoch interacts badly with the cosine LR schedule (T_max=30)**. With 2x batches per epoch, the cosine cycle completes in ~4 epochs instead of ~8, creating a fast warm-restart pattern.

**Three phases of training:**

1. **Epochs 1–42 (convergent):** Model made steady progress, reaching 11.57% val at ep42. Each cosine cycle trough delivered a new best. However, mid-cycle spikes appeared early (ep13: 54.8%, ep16: 47.4%) when LR peaked at 5e-4.

2. **Epochs 42–128 (oscillating/degrading):** The instability pattern worsened. Some cycles were smooth (ep20–29), but catastrophic spikes recurred (ep30: 52.5%, ep45: 46.3%). After ep45, gradient norms began escalating exponentially: 13 → 46 → 69 → 153 → 392 → ... The model partially recovered between spikes but never regained its ep42 best.

3. **Epochs 129–321 (NaN):** Gradient norms reached billions (ep128: 5.5 trillion) before hitting NaN at ep129. The remaining 192 epochs ran pointlessly at NaN until the 360-minute budget expired.

**Root cause:** With 788 batches/epoch, each epoch takes ~1.1 minutes vs ~0.55 minutes with 394 batches. The cosine T_max=30 operates on steps, so the LR cycle completes in half as many epochs. This creates abrupt LR transitions that destabilize training — especially the jump to peak LR (5e-4) every ~4 epochs. Without `--grad-clip`, these spikes are unconstrained and eventually cascade into permanent divergence.

**Why the baseline doesn't have this problem:** With 394 batches/epoch, the cosine cycle spans ~8 epochs, giving the model more gradual LR transitions. The baseline ran 467 stable epochs to reach 3.997%.

### Suggested follow-ups

1. **If 788 batches/epoch is worth retesting**, it needs `--cosine-t-max 60` (to match the ~8 epoch cycle length at 394 batches) and `--grad-clip 1.0` or `--grad-clip 2.0` to prevent gradient explosions.
2. **T_max should scale with batches/epoch** — the current config implicitly couples the LR schedule to the epoch length, which breaks when data throughput changes.
3. The NaN-but-keep-running behavior wasted ~3 hours of GPU time. A NaN early-stop check in the training loop would save significant compute.

tcapelle:
ADVISOR: Closing PR #3084 because ADVISOR: Closing — 788 batches/epoch is a dead end. Best val 11.565% at ep42 (+190% above 3.997% baseline), then catastrophic divergence at ep45 (grad norms reaching trillions). Root cause: T_max=30 operates on epoch count — with 2x batches/epoch the effective cosine cycle halves, creating more frequent LR peak shocks. Without gc there is no safety net. The 4x variant (#3051) showed the same failure. A viable follow-up is 788 batches + proportionally scaled T_max=60 + gc, but that's a fresh hypothesis. Good mechanistic analysis, kakashi..

PR #3025, opened 2026-04-22
TandemFoil Paper: exact TF champion config (Lion+gc=0.5+EMA) — first test on TFP

tcapelle:
## Hypothesis

The TandemFoil Paper (TFP) benchmark is fresh territory — PR #2979 established the first baseline at **val_primary/field_mse = 0.00434** using AdamW + no-EMA + T_max=150 + 3L/192d. However, the **TandemFoil champion config** (Lion lr=1.25e-4, gc=0.5, EMA=0.999, T_max=10) has never been tested on TFP. Jin's PR #2947 used Lion but with different hyperparameters (gc=1.0, no-EMA) — not the exact winner.

On the standard TandemFoil benchmark, this combination (Lion + low lr + tight grad-clip + EMA) achieved **val_primary/surface_pressure_mae = 22.537** (PR #2924) — the best on that dataset. The hypothesis is that the same combination will transfer to TFP and outperform the AdamW baseline, since Lion's slower, momentum-driven convergence combined with EMA smoothing tends to find better minima on complex aerodynamic fields.

## Experiment

Test the **exact TF champion hyperparameter config** on the TandemFoil Paper dataset with 3L/192d architecture (same as the TFP champion):

| Hyperparameter | Value | Rationale |
|---|---|---|
| Optimizer | Lion | TF champion optimizer |
| lr | 1.25e-4 | TF champion lr |
| cosine-t-max | 10 | TF champion schedule |
| grad-clip | 0.5 | TF champion gc (vs gc=1.0 in PR #2947) |
| weight-decay | 1e-2 | Standard with Lion |
| EMA decay | 0.999 | TF champion EMA (not used in PR #2979 baseline) |
| model-layers | 3 | TFP champion depth |
| model-hidden-dim | 192 | TFP champion width |
| model-heads | 3 | TFP champion heads |
| enable-fourier | yes | TFP champion flag |
| All physics flags | yes | See command below |

## Current Baselines (targets to beat)

| Dataset | Metric | Current Best | PR |
|---|---|---|---|
| TandemFoil Paper | val_primary/field_mse | **0.00434** | #2979 (haku, AdamW ep199) |
| TandemFoil | val_primary/surface_pressure_mae | 22.537 | #2924 (Lion champion) |

**Primary target: beat 0.00434 on TandemFoil Paper.**

## Instructions

Run the following command. Use `--wandb_group tandemfoil-paper-lion-sweep` so future Lion variants on TFP group together in W&B:

```bash
cd target/icml2026
[...]

PR #3089, opened 2026-04-22
TandemFoil Paper: T_max=30 at champion config

tcapelle:
## Hypothesis

T_max=30 (like TandemFoil parity champion) may be more stable for TFP as well, while still providing enough LR oscillation for basin exploration. Tests whether the divergence pattern is purely a T_max=10 issue.

## Instructions

```bash
  --dataset tandemfoil_paper --optimizer lion --lr 1.25e-4 --cosine-t-max 30 \
  --wandb_group tfp-tmax-sweep
```

## Reporting Requirements
- W&B run ID

**Paper-facing metric:** `test_primary/field_mse` (normalized full-field MSE).

## Baseline

**TFP baselines:**

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -227,7 +227,7 @@ class TargetTransform:
         if self.stats_mean is not None and self.stats_std is not None and self.stats_mean.numel() == out.shape[-1]:
             out = out * self.stats_std.to(out.device) + self.stats_mean.to(out.device)
         if self.asinh_pressure and self.pressure_index is not None:
-            out[..., self.pressure_index] = torch.sinh(out[..., self.pressure_index]) / max(self.asinh_scale, 1e-6)
+            out[..., self.pressure_index] = torch.sinh(out[..., self.pressure_index].clamp(-40.0, 40.0)) / max(self.asinh_scale, 1e-6)
         return out
 
 
@@ -972,8 +972,8 @@ def evaluate_grouped(
                 case_values = (surface_pred - surface_target).square()
                 valid = batch.surface_mask.unsqueeze(-1)
                 if paper_surface_sq_sum is None:
-                    paper_surface_sq_sum = torch.zeros(case_values.shape[-1], device=device)
-                paper_surface_sq_sum += (case_values * valid).sum(dim=(0, 1)).to(device)
+                    paper_surface_sq_sum = torch.zeros(case_values.shape[-1], device=device, dtype=torch.float64)
+                paper_surface_sq_sum += (case_values * valid).sum(dim=(0, 1)).to(device=device, dtype=torch.float64)
                 paper_surface_nodes += int(batch.surface_mask.sum().detach().cpu().item())
             if batch.volume_y is not None and outputs["volume_preds"] is not None and batch.volume_mask is not None:
                 volume_pred, volume_target = comparison_metric_tensors(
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT nami:

## Bug fix: TFP pressure overflow in TargetTransform.invert()

Found and fixed a critical bug affecting **all** TandemFoil Paper evaluation metrics. `val/surface_mse_p` and `val/volume_mse_p` were reporting Infinity from epoch 1, which made `field_mse` (the primary metric) Infinity too.

**Root cause:** `TargetTransform.invert()` calls `torch.sinh()` on de-normalized pressure values without clamping. The normalization uses raw physical-pressure stats (std=461.985) to z-score asinh-transformed values — a scale mismatch. Even modest model prediction errors (±0.3 in normalized space) produce de-normalized values >100, where `sinh(100)` overflows float32.

**Fixes applied:**
1. Clamp sinh input to ±40.0 in `TargetTransform.invert()` (prevents overflow while preserving meaningful error magnitudes)
2. Promote TFP evaluation accumulators (`paper_surface_sq_sum`, `paper_volume_sq_sum`) from float32 to float64

**Impact:** Only affects the TFP evaluation metric path. Training loss (computed in normalized asinh space) was never affected. This likely explains why previous TFP baselines worked — the model converged quickly enough with T_max=10 that pressure predictions stayed in the safe range, but with T_max=30 the higher LR persists longer, pushing predictions further out.

Experiment now running with the fix — will report results when complete.

tcapelle:
STUDENT nami:

## Results

### T_max=30 Experiment

| Metric | Value | Baseline (T_max=10) |
|--------|-------|-------------------|
| **Best val field_mse** | 4.36e+23 @ ep308 | **0.002383** @ ep443 |
| **Best test field_mse** | 1.30e+24 @ ep308 | — |
| Final test field_mse | 6.01e+24 (last epoch) | — |
| Best test surface_mse_Ux | 0.001324 | — |
| Best test surface_mse_Uy | 0.003840 | — |
| Best test surface_mse_p | 1.19e+24 | — |
| Best test volume_mse_Ux | 0.007868 | — |
| Best test volume_mse_Uy | 0.005842 | — |
| Best test volume_mse_p | 3.95e+24 | — |
| Total epochs | 536 (360-min timeout) | 443 |
| W&B run ID | rfr8qkrc | d1xh0o1p |

### Command
```bash
  --dataset tandemfoil_paper --optimizer lion --lr 1.25e-4 --cosine-t-max 30 \
  --wandb-group tfp-tmax-sweep \
  --agent nami --wandb-name "nami/tfp-tmax-30"
```

### What happened

**Hypothesis FALSIFIED.** T_max=30 is dramatically worse than T_max=10 for TFP. The best val field_mse (4.36e+23 at ep308) is ~1.8e+26 times worse than the baseline (0.002383 at ep443), despite running 93 more epochs.

The Ux/Uy velocity channels converged well (surface_mse_Ux=0.0013, surface_mse_Uy=0.0038), but pressure never converged. The evaluation metric passes through `sinh()` denormalization, and the volume pressure predictions never entered the narrow normalized-space range where sinh produces reasonable physical-space values. Surface pressure went through a phase transition around ep100-120 (dropping from 1e24 to 1e17), but volume pressure remained stuck at 1e24-1e25 throughout.

**Root cause:** With T_max=30, the cosine LR schedule keeps the learning rate elevated for 3x longer per cycle (30 epochs vs 10), which prevents the model from settling into precise enough pressure predictions during LR troughs. T_max=10's rapid cycling gives the model more frequent opportunities to find deep basins for pressure prediction.

**Bug fix included:** Fixed a pre-existing overflow bug in `TargetTransform.invert()` where `sinh()` had no input clamp, causing all TFP pressure metrics to report Infinity. Also promoted evaluation accumulators to float64 to prevent accumulation overflow. See earlier comment for details.

### Suggested follow-ups
- T_max=10 remains the champion for TFP — no reason to increase it
- The normalization mismatch (raw physical-pressure stats used to z-score asinh-transformed values) should be addressed for all TFP experiments — it makes the evaluation metric extremely sensitive to even small prediction errors in normalized space
- T_max=15 or T_max=20 might be worth testing as intermediate values, but given how dramatically T_max=30 failed, T_max=10 is likely near-optimal for TFP

tcapelle:
ADVISOR: Closing PR #3089 because ADVISOR: T_max=30 gives field_mse ~4.36e+23 for all 536 epochs — pressure never converges even with your sinh clamp fix. The fix is valuable infrastructure but doesn't save T_max=30. This confirms the pattern: T_max=15 (diverged ep124), T_max=20 (in-flight), T_max=30 (pressure never finite). T_max>10 is dead for TFP — the asinh pressure transform requires the short 10-epoch cosine troughs to maintain gradient scale. T_max=10 confirmed as sharp optimum..

PR #3092, opened 2026-04-22
TandemFoil Paper: EMA decay=0.9995 (slower averaging)

tcapelle:
## Hypothesis

Champion EMA=0.999. Slower decay (0.9995) provides a longer effective averaging window: at 443 epochs, 0.9995 incorporates ~1/(1-0.9995) = 2000 recent updates vs 1000 for 0.999. May smooth out more epoch-to-epoch variation but react slower to improvements.

## Instructions

```bash
  --epochs 999 --ema-decay 0.9995 \
  --wandb_group tfp-ema-sweep
```

## Reporting Requirements
- Best val `field_mse` + epoch (EMA model)
- W&B run ID

## Baseline

**TFP baselines:**


tcapelle:
STUDENT norman:

## Results

### EMA decay=0.9995 — NEGATIVE RESULT

| Metric | Baseline (EMA=0.999) | EMA=0.9995 |
|---|---|---|
| val_primary/field_mse | **0.002383** @ ep443 | Infinity / ~1e+29 (never converged) |
| val/surface_mse_Ux | 0.001517 (composite) | 0.001030 @ ep536 |
| val/surface_mse_Uy | — | 0.002614 @ ep536 |
| val/surface_mse_p | 4.81e-05 | **3.04e+25** @ ep536 |
| val/volume_mse_p | — | **6.25e+29** @ ep536 |
| test_primary/field_mse | — | Infinity |
| Epochs trained | 443 (best) / 462 (diverged) | 537 (360-min budget) |
| Train loss | — | 0.0135 @ ep536 (healthy) |

- **W&B run:** `xy01mjzl` (norman/tfp-ema-decay-9995)
- **Best epoch (per W&B):** 266 — but "best" val_primary/field_mse was 1.48e+28, not usable
- **Peak memory:** single GPU, ~20GB

### Command
```bash
  --epochs 999 --ema-decay 0.9995 \
  --wandb-group tfp-ema-sweep \
  --agent norman --wandb-name "norman/tfp-ema-decay-9995"
```

### What happened

**Hypothesis falsified.** EMA decay=0.9995 completely failed on TandemFoil Paper — the pressure channel in the EMA model never converged to finite values across 537 epochs.

Key observations:
- The **underlying model trained normally**: train loss descended from 1.04 → 0.013 over 537 epochs, and the Ux/Uy surface components in the EMA model were healthy (0.001030 / 0.002614 — comparable to baseline)
- The **pressure channel (surface_mse_p)** remained at Infinity for the first ~210 epochs, then oscillated between 1e+25 and 1e+32 — never approaching the baseline's 4.81e-05
- The EMA warmup formula `min(target_decay, (1+step)/(10+step))` ramps to the target decay. With decay=0.9995, the effective averaging window is ~2000 recent updates vs ~1000 for decay=0.999. This means the EMA model retains more weight from early epochs when the asinh pressure transform produced extreme values
- The `volume_mse_p` was even worse (1e+29-1e+30 at final epoch), confirming the pressure divergence is pervasive across both surface and volume

**Root cause:** The longer EMA averaging window (0.9995 → ~2000 updates) is incompatible with the asinh pressure transform on TFP. Early-epoch pressure predictions with extreme dynamic range poison the EMA average, and the slow decay never recovers. EMA=0.999 (~1000 updates) has just enough "forgetting" to eventually wash out the early instability.

### Suggested follow-ups
- **EMA=0.999 is confirmed as the sweet spot for TFP** — faster forgetting is essential with asinh pressure
- Could try EMA=0.9993 or 0.9990 to map the exact failure boundary, but likely diminishing returns since 0.999 already works well
- A potential fix would be to delay EMA onset until pressure has stabilized (e.g., start EMA after epoch 50), but this adds complexity for uncertain gain

tcapelle:
ADVISOR: Closing PR #3092 because ADVISOR: EMA=0.9995 (slower averaging, ~2000-update window) also fails — field_mse stays at ~1e+28 across 536 epochs despite the model training normally (loss 1.04→0.013). The slow EMA retains too much weight from early epochs when the pressure was producing extreme asinh values. EMA=0.999 is confirmed as the sweet spot — both directions (faster and slower) cause sinh overflow. EMA sweep fully closed for TFP..

PR #3093, opened 2026-04-22
TandemFoil Paper: EMA decay=0.99 (faster averaging)

tcapelle:
## Hypothesis

Faster EMA (decay=0.99) produces a shorter effective averaging window (~100 updates vs 1000). This makes the EMA model more responsive but noisier. For TFP's rapidly oscillating cosine schedule, faster EMA may be preferable to track the evolving loss surface.

## Instructions

```bash
  --epochs 999 --ema-decay 0.99 \
  --wandb_group tfp-ema-sweep
```

## Reporting Requirements
- W&B run ID

## Baseline

**TFP baselines:**


tcapelle:
STUDENT rei:

## Results

### EMA decay=0.99 — NEGATIVE RESULT (field_mse = Infinity for all 532 epochs)

| Metric | Baseline (EMA=0.999, #3025) | EMA=0.99 |
|--------|--------------------------|----------|
| val_primary/field_mse | 0.002383 @ ep443 | **Infinity** (all 532 epochs) |
| Best finite surface_mse_p | meaningful | 2.16e+15 @ ep291 |
| Best training loss | — | 0.015 @ ep432 |
| Total epochs | 443+ | 532 |

- **W&B run:** [`9j29lyiw`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/9j29lyiw)
- **Command:**
  ```bash
    --epochs 999 --ema-decay 0.99 \
    --agent rei --wandb-name "rei/tfp-ema-99" --wandb-group tfp-ema-sweep
  ```

### What happened

**EMA decay=0.99 is too fast for TFP with asinh pressure.** The hypothesis is falsified.

The fundamental issue is that the faster EMA (decay=0.99, effective window ~100 updates) tracks the model's noisy pressure predictions too closely. When these are passed through the `sinh()` inverse transform at evaluation time, they overflow to Infinity. The baseline EMA=0.999 (effective window ~1000 updates) smooths out this noise sufficiently for the pressure predictions to stay within the representable range of `sinh()`.

**Key observations:**
1. **Velocity channels learned perfectly** — surface_mse_Ux reached 0.0015, surface_mse_Uy reached 0.005. The underlying model was learning well.
2. **Pressure never resolved** — val/surface_mse_p was Infinity for the first 46 epochs, then began producing intermittent finite readings at cosine troughs (e+31 → e+27 → e+21 → e+15), but these values were still astronomically large and never approached meaningful levels.
3. **Training loss was healthy** — decreased from 1.018 to 0.015 over 532 epochs, with a smooth downward envelope. The online model was training fine; only the EMA model's pressure overflowed.
4. **Best checkpoint was epoch 1** — since val_primary/field_mse was Infinity for ALL epochs, the checkpoint system never found a better epoch than the first. This means the test evaluation used the epoch-1 model, producing meaningless test metrics.
5. **Grad norms stabilized** — from 49.0 at epoch 2 to ~10.0 by epoch 100, with clipping gradually easing from 360/360 to ~340/360 batches.

### Why EMA=0.99 fails where EMA=0.999 succeeds

The asinh pressure transform compresses large pressure values during training via `asinh(p * scale)`. At evaluation, the inverse `sinh()` is applied. With a noisy EMA model, even small deviations in the compressed space get exponentiated back through `sinh()`, producing values that overflow to Infinity. The slower EMA (0.999) averaging over ~1000 steps provides enough smoothing to keep pressure predictions stable through this transform. The faster EMA (0.99) at ~100 steps does not.

### Suggested follow-ups

1. **EMA=0.995 sweep** — test an intermediate decay to find the fastest stable EMA for TFP. The transition between "works" (0.999) and "fails" (0.99) likely lies in the 0.995-0.999 range.
2. **EMA=0.99 without asinh pressure** — the failure is specific to the sinh inverse transform. Without asinh, faster EMA might actually work and be beneficial.
3. **Scheduled EMA decay** — start with slow EMA (0.999) for pressure stabilization, then anneal to faster EMA (0.99) after the model converges. This would give the responsiveness benefit later in training.

tcapelle:
ADVISOR: Closing PR #3093 because ADVISOR: EMA=0.99 gives val/field_mse=Infinity for all 532 epochs — pressure never converges to finite values. The faster EMA window (~100 updates) allows early noisy pressure predictions to survive through the sinh() inverse transform as overflows. This is a fundamental incompatibility with the asinh pressure encoding. Closing — EMA<0.999 is dead for TFP due to the sinh amplification of early-epoch noise..

PR #2949, opened 2026-04-22
tandemfoil_paper: depth and width sweep for field_mse

tcapelle:
## Hypothesis

On the main tandemfoil benchmark, 3L/192d won over wider and deeper configurations. But tandemfoil_paper measures normalized full-field MSE — a spatially dense target over all nodes and three channels. Deeper networks may have higher expressive capacity needed for simultaneous velocity and pressure field reconstruction. This PR sweeps depth (3L vs 4L vs 5L) and width (192d vs 256d vs 384d) to find the optimal architecture for the paper metric, independent of the optimizer/physics-flag experiments running in parallel (PRs #2947, #2948).

## Current Baseline

**tandemfoil_paper: NO BASELINE — PR #2947 (jin) is establishing the first run simultaneously.**

Primary metric: `val_primary/field_mse` (normalized full-field MSE, lower is better).

Paper reference: MGN baseline ~1.79 (cruise_random_uniform). Paper best: 0.10.

## Experiment Plan

Fix: Lion lr=1e-4, T_max=10, gc=1.0, WD=1e-2, no-EMA, all physics flags, Fourier. Sweep depth and width:

| Run | Layers | Hidden dim | Heads |
|---|---|---|---|
| A | 3 | 192 | 3 |
| B | 3 | 256 | 4 |
| C | 3 | 384 | 6 |
| D | 4 | 192 | 3 |
| E | 4 | 256 | 4 |
| F | 4 | 384 | 6 |
| G | 5 | 192 | 3 |
| H | 5 | 256 | 4 |

## Instructions

**Run A (3L/192d/3H — reference):**
```bash
cd target/icml2026 && python train.py   --dataset tandemfoil_paper   --optimizer lion   --lr 1e-4   --cosine-t-max 10   --grad-clip 1.0   --weight-decay 1e-2   --no-use-ema   --model-slices 64   --model-layers 3   --model-hidden-dim 192   --model-heads 3   --enable-fourier   --enable-te-coord-frame   --enable-cp-panel   --enable-cp-panel-tandem-only   --asinh-pressure   --residual-prediction   --enable-pressure-prior-addition   --epochs 999   --wandb_group tandemfoil_paper_arch_sweep
```

**Run B:** Same but `--model-hidden-dim 256 --model-heads 4`

**Run C:** Same but `--model-hidden-dim 384 --model-heads 6`

**Run D:** Same as A but `--model-layers 4`

**Run E:** Same as D but `--model-hidden-dim 256 --model-heads 4`

[...]

PR #3094, opened 2026-04-22
TandemFoil Paper: 4L/192d depth at champion optimizer config

tcapelle:
## Hypothesis

PR #2949 (vash, sent back) found that 4L/192d is the optimal depth-width combination for TFP in a sweep using a different LR. Now test 4L/192d at the full champion config (Lion lr=1.25e-4, gc=0.5, EMA=0.999, T_max=10). The depth increase from 3L to 4L may provide better volume field modeling.

## Instructions

Change only `--model-layers 4` vs champion:

```bash
cd target/icml2026 && python train.py \
  --dataset tandemfoil_paper --optimizer lion --lr 1.25e-4 --cosine-t-max 10 \
  --grad-clip 0.5 --weight-decay 1e-2 --enable-fourier \
  --model-layers 4 --model-hidden-dim 192 --model-heads 3 \
  --enable-te-coord-frame --enable-cp-panel --enable-cp-panel-tandem-only \
  --asinh-pressure --residual-prediction --enable-pressure-prior-addition \
  --epochs 999 --ema-decay 0.999 \
  --wandb_group tfp-arch-sweep
```

## Reporting Requirements
- Best val `field_mse` + epoch
- Test `field_mse` at best val checkpoint
- W&B run ID

## Baseline

**TFP baselines:**
- val best: `field_mse = 0.002383` @ ep443 (PR #3025, W&B: `d1xh0o1p`)
- Config: Lion lr=1.25e-4, T_max=10, gc=0.5, EMA=0.999, 3L/192d, Fourier+physics
- paper target: normalized full-field MSE (field_mse, lower is better)
- **No clean test_primary/field_mse yet** — that is what the TFP experiments must establish

Note: vash #2949 found 4L/192d best at different LR. haku #3056 is sweeping LR at 3L. This tests 4L at champion LR.

PR #3095, opened 2026-04-22
TandemFoil Paper: 4L/256d (deeper + wider)

tcapelle:
## Hypothesis

Combine the 4L depth insight from PR #2949 with wider hidden dimension (256d vs 192d). More model capacity may better resolve the complex multi-task field_mse (normalized over 6 extrapolation tasks).

## Instructions

```bash
  --wandb_group tfp-arch-sweep
```

## Reporting Requirements
- W&B run ID

## Baseline

**TFP baselines:**


tcapelle:
STUDENT sanji:

## Results

### TandemFoil Paper: 4L/256d (deeper + wider) — NEGATIVE RESULT

| Metric | Baseline (3L/192d) | 4L/256d | Delta |
| val_primary/field_mse | 0.002383 @ ep443 | **Infinity** (70 epochs) | ❌ Failed |
| val/surface_mse_Ux | — | 0.00179 (improving) | ✓ |
| val/surface_mse_Uy | — | 0.00572 (improving) | ✓ |
| val/surface_mse_p | 4.81e-05 @ ep443 | **Infinity** (diverging) | ❌ |
| val/volume_mse_p | 7.28e-05 @ ep443 | **Infinity** (never finite) | ❌ |
| training loss | — | 0.054 (still decreasing) | ✓ |

**W&B run:** `sfbvlaof` (sanji/tfp-4l-256d-depth-width)
**Peak memory:** ~41 GB (GPU 0)
**Epochs completed:** 70 (~98 minutes of 360-min budget)
**Stopped early:** Yes — pressure channel diverging, no path to finite field_mse

**Command:**
```bash
  --agent sanji --wandb-name "sanji/tfp-4l-256d-depth-width" \
  --wandb-group tfp-arch-sweep
```

## What happened

The 4L/256d model **completely failed to converge on the pressure channel** while velocity channels (Ux, Uy) converged normally. The root cause is a pathological initialization: the larger model produces pressure predictions in asinh-normalized space that are so extreme that `torch.sinh()` during metric inversion overflows float32 (|x| > ~89).

**Pressure trough progression at T_max=10 aligned epochs:**

| Epoch | surface_mse_p |
|-------|--------------|
| 12 | 1.62e+31 (first finite) |
| 26 | 2.69e+19 |
| 40 | 3.91e+18 (best) |
| 50 | 1.89e+25 (regressed) |
| 60 | 2.01e+21 |
| 70 | **Infinity** (diverged) |

The best trough (ep40: 3.91e+18) is still **~20 orders of magnitude** from the champion's epoch 1 value (0.00533). The trough values stopped improving after epoch 40 and actively diverged at epoch 70. Meanwhile, the champion run (3L/192d, same config) had finite pressure from the very first epoch.

The training loss decreased from 1.01 to 0.054 — this improvement was driven entirely by velocity channels. The model was optimizing velocity at the expense of pressure, likely because the pressure gradients were dominated by overflow artifacts.

## Analysis

**Why 4L/256d fails while 3L/192d works:** The larger model (4L/256d = ~2.7x more parameters) has a worse weight initialization scale for the pressure prediction head. With more layers and wider hidden dimensions, the initial output scale for pressure predictions lands further from the valid asinh range. Once in the sinh overflow regime, the pressure error signal is corrupted (Infinity gradients), creating a chicken-and-egg problem where the model can't learn to fix pressure because the pressure error signal itself is broken.

**This is specific to the asinh pressure transform + larger models.** The champion 3L/192d config doesn't have this problem because its initialization produces pressure predictions closer to the valid range.

## Suggested follow-ups

1. **Output clamping:** Add `torch.clamp(pred[..., pressure_idx], -80, 80)` before the sinh inversion in eval to prevent overflow. This wouldn't fix training but would give visibility into how close predictions actually are.
2. **Pressure head initialization:** Scale down the final layer weights for the pressure output channel specifically when using asinh_pressure + larger models.
3. **Skip 4L/256d for TFP:** The 3L/192d architecture appears to be the sweet spot for TandemFoil Paper — more capacity hurts rather than helps due to the pressure initialization problem.
4. **Try 4L/192d (depth only):** A `robin/tfp-4l-192d-depth` branch already exists — check if depth alone (without width increase) also suffers from this issue. If 4L/192d works, it confirms the width increase is the culprit.

tcapelle:
ADVISOR: Closing PR #3095 because ADVISOR: Closing — 4L/256d produces Infinity on TFP.

**Results:** val_primary/field_mse = Infinity for all 70 epochs. Pressure channel never converged. The larger model's initialization overflows float32 during asinh inversion (sinh(x) for |x|>89). Velocity channels converged normally.

**Conclusion:** 4L/256d is incompatible with asinh-pressure on TFP. The 3L/192d champion avoids this overflow. robin (#3094) testing 4L/192d will clarify whether depth or width is the specific culprit.

Reassigning you to DrivAerML..

PR #3080, opened 2026-04-22
DrivAerML: T_max=50 cosine schedule (longer cycles)

tcapelle:
## Hypothesis

Champion uses T_max=30. Longer cosine periods (T_max=50) allow the optimizer to spend more time descending into each basin before the LR kicks back up. AirfRANS benefited significantly from T_max=50 vs T_max=10.

## Instructions

```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 50 \
  --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 \
  --model-heads 8 --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
  --wandb_group dm-tmax-sweep
```

## Reporting Requirements
- Best val `surface_rel_l2_pct` + epoch
- Test at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML baselines:**
- val best: `3.997%` @ ep467 (PR #2898, W&B: `bht6h42t`, 4L/512d, AdamW lr=5e-4, T_max=30)
- test best: `6.244%` (PR #2648, 4L/320d, truncated eval — OLDER CONFIG)
- External targets: AB-UPT 3.82% | Transolver-3: 3.71%

Note: T_max=5 diverged (#2911). T_max=30 is champion. T_max=50 never tested at current 4L/512d config.

PR #3111, opened 2026-04-22
DrivAerML: cosine eta_min=1e-6/1e-5 (LR floor at cycle trough)

tcapelle:
## Hypothesis

The current cosine schedule (`CosineAnnealingLR(T_max=30)`) drives LR down to `eta_min=0` at each cycle trough. At LR=0 the optimizer takes zero-sized steps — wasted computation. Setting `eta_min` to a small positive value (e.g. 1e-6 or 1e-5) ensures the model continues making progress even at cycle troughs, and may also reduce the abruptness of the LR transition from 0→peak when a new cycle starts.

PyTorch's `CosineAnnealingLR` already supports the `eta_min` parameter — it just defaults to 0.

## Instructions

Add `--cosine-eta-min` flag to train.py. When building `CosineAnnealingLR`, pass `eta_min=args.cosine_eta_min`.

Run two variants:

**Run 1: eta_min=1e-6 (conservative floor)**
```bash
  --cosine-eta-min 1e-6 \
  --wandb_group dm-eta-min-sweep
```

**Run 2: eta_min=1e-5 (higher floor)**
```bash
  --cosine-eta-min 1e-5 \
  --wandb_group dm-eta-min-sweep
```

**Implementation:** One line change in the scheduler setup:
```python
scheduler = CosineAnnealingLR(optimizer, T_max=args.cosine_t_max, eta_min=args.cosine_eta_min)
```

## Reporting Requirements
- Best val `surface_rel_l2_pct` + epoch for each variant
- Test at best val checkpoint
- W&B run IDs

## Baseline

**DrivAerML champion:**
- W&B run: `bht6h42t` (PR #2898, 4L/512d, AdamW lr=5e-4, T_max=30, eta_min=0)
- External targets: Transolver-3 3.71% | AB-UPT 3.82%

```bash
# Champion reproduce
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -160,6 +160,7 @@ class TrainConfig:
     residual_prediction: bool = False
     re_stratified_sampling: bool = False
     cosine_t_max: int = 150
+    cosine_eta_min: float = 0.0
     geometry_points: int = 25_000
     geometry_supernodes: int = 4_096
     surface_anchor_points: int = 8_000
@@ -1625,7 +1626,7 @@ def main() -> None:
     scheduler = None
     if config.cosine_t_max > 0:
         base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
-        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max, eta_min=config.cosine_eta_min)
 
     ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
     anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None


tcapelle:
STUDENT himmel:

## Results

### Experiment: DrivAerML cosine eta_min sweep (1e-6 vs 1e-5)

| Metric | Baseline (T_max=30, eta_min=0) | eta_min=1e-6 | eta_min=1e-5 |
|--------|------|------|------|
| **Best val surface_rel_l2_pct** | **3.997%** (ep467) | **7.918%** (ep180) | **7.255%** (ep109) |
| Instability onset | N/A (with grad-clip) | ep60 (grad_norm=40.8) | ep128 (grad_norm=16.9) |
| Terminal divergence | N/A | ep186 (grad_norm=93.8) | ep184 (grad_norm=28,848) |
| Final epoch reached | 467+ | 201 (killed) | 201 (killed) |

**External references:** Transolver-3 3.71% | AB-UPT 3.82%

### Commands
```bash
# eta_min=1e-6 (GPU 0)
  --cosine-eta-min 1e-6 \
  --wandb-group dm-eta-min-sweep \
  --agent himmel --wandb-name "himmel/dm-eta-min-1e6"

# eta_min=1e-5 (GPU 1)
  --cosine-eta-min 1e-5 \
  --wandb-group dm-eta-min-sweep \
  --agent himmel --wandb-name "himmel/dm-eta-min-1e5"
```

### W&B Run IDs
- eta_min=1e-6: `gu7hfzs1`
- eta_min=1e-5: `q6zdf6xk`

### Peak memory
Both runs on single H100 (98GB). No OOM issues.

### What happened

**Hypothesis falsified.** Neither eta_min value beats the 3.997% baseline, and both runs diverge catastrophically well before reaching the epoch counts (400+) needed to match the baseline.

**eta_min=1e-6:** Behaves nearly identically to eta_min=0 as expected — the floor is so low it provides no meaningful regularization. First gradient explosion at ep60 (grad_norm 40→136), brief recovery ep73-185 reaching best 7.918% at ep180, then terminal divergence at ep186+. The recovery is interesting — it shows the model *can* recover from single explosions when the LR cycle happens to bring the rate down, but this is fragile.

**eta_min=1e-5:** Marginally better. Stable through the first 127 epochs with best 7.255% at ep109. However, repeated instability episodes began at ep128 and escalated through ep142, ep160, ep168. A catastrophic explosion at ep184 (grad_norm=28,848!) pushed the model into permanent divergence. The 10x higher floor did buy ~60 more stable epochs vs 1e-6, but not nearly enough to reach competitive performance.

**Key insight:** The core problem is not eta_min — it's the absence of gradient clipping. The baseline achieves 3.997% with `--grad-clip` enabled and eta_min=0. Both eta_min runs here used no grad-clip, matching the PR #3080 finding that DrivAerML 4L/512d is fundamentally unstable without gradient clipping regardless of LR floor. The eta_min parameter cannot substitute for grad-clip as a stability mechanism.

**Both runs plateau around 7-8% best** even during their stable phases, far from the 3.997% baseline — suggesting that even if instability were prevented, the learning dynamics without grad-clip produce inferior solutions in this configuration.

### Suggested follow-ups
1. **eta_min + grad-clip combination**: Test eta_min=1e-5 with `--grad-clip` enabled. The LR floor might provide additional benefit *on top of* gradient clipping, even if it can't replace it.
2. **Warmup restart**: CosineAnnealingWarmRestarts with eta_min might handle the trough-to-peak transitions more gracefully than standard CosineAnnealingLR.
3. **Smaller eta_min with grad-clip**: eta_min=1e-6 with grad-clip — the floor prevents true LR=0 at troughs while grad-clip handles the explosions.

tcapelle:
ADVISOR: Closing PR #3111 because ADVISOR: eta_min=1e-6: 7.918%, eta_min=1e-5: 7.255% — both diverged before ep200 without grad-clip. The hypothesis is sound but tested in the wrong regime (no-EMA, no gc). spike (#3161) is now testing eta_min WITH EMA+gc, which is the correct stability platform. Closing this run — spike will carry the hypothesis forward..

PR #3112, opened 2026-04-22
DrivAerML: gradient centralization (implicit regularizer)

tcapelle:
## Hypothesis

Gradient centralization (Yong et al. 2020, "Gradient Centralization: A New Optimization Technique for DNNs") subtracts the per-tensor gradient mean before the optimizer step. This acts as an implicit regularizer by constraining the weight space to a manifold where activation means are preserved, and projects gradients to a lower-variance subspace.

For DrivAerML's batch_size=1 training with high gradient variance, GC may smooth the optimization trajectory without the instability introduced by explicit regularizers (WD crashed at ep178, gc=2.0 diverged). Unlike gradient clipping which truncates, GC centers — it preserves gradient direction while reducing inter-sample drift.

**Reference:** https://arxiv.org/abs/2004.01461

## Instructions

Add `--grad-centralization` flag to train.py. When active, apply gradient centralization as a pre-step hook:

```python
def centralize_gradients(optimizer):
    for group in optimizer.param_groups:
        for p in group['params']:
            if p.grad is None:
                continue
            if p.grad.dim() > 1:  # only centralize weight tensors, not biases
                p.grad.data -= p.grad.data.mean(dim=tuple(range(1, p.grad.dim())), keepdim=True)
```

Call `centralize_gradients(optimizer)` after `loss.backward()` and before `optimizer.step()`.

```bash
  --grad-centralization \
  --wandb_group dm-grad-centralization
```

## Reporting Requirements
- Test at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML champion:**
- W&B: `bht6h42t` (4L/512d, AdamW lr=5e-4, T_max=30, no gc/WD/EMA)

```bash
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -166,6 +166,7 @@ class TrainConfig:
     volume_anchor_points: int = 8_000
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
+    grad_centralization: bool = False
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -451,6 +452,15 @@ def parse_args() -> TrainConfig:
             parser.add_argument(arg_name, type=type(field_value), default=field_value)
     namespace = parser.parse_args()
     return TrainConfig(**vars(namespace))
+
+
+def centralize_gradients(optimizer):
+    for group in optimizer.param_groups:
+        for p in group['params']:
+            if p.grad is None:
+                continue
+            if p.grad.dim() > 1:
+                p.grad.data -= p.grad.data.mean(dim=tuple(range(1, p.grad.dim())), keepdim=True)
 
 
 def build_bundle(config: TrainConfig) -> DatasetBundle:
@@ -1266,6 +1276,7 @@ def train_one_epoch(
     max_batches: int = 0,
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
+    use_grad_centralization: bool = False,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
@@ -1339,6 +1350,8 @@ def train_one_epoch(
         if grad_clip > 0 and float(grad_norm) > grad_clip:
             running.setdefault("grad_clip_events", 0.0)
             running["grad_clip_events"] += 1.0
+        if use_grad_centralization:
+            centralize_gradients(optimizer)
         optimizer.step()
         if scheduler is not None:
             scheduler.step()
@@ -1671,6 +1684,7 @@ def main() -> None:
             max_batches=config.max_train_batches,
             grad_clip=config.grad_clip,
             grad_accum_steps=config.grad_accum_steps,
+            use_grad_centralization=config.grad_centralization,
         )
 
         if ema is not None:


tcapelle:
STUDENT edward:

## Results

### Gradient Centralization — DrivAerML

| Metric | GC (this run) | Baseline (no GC) | Delta |
|--------|--------------|-------------------|-------|
| **val_primary/surface_rel_l2_pct** | **5.492%** (ep 181) | **3.997%** (ep 467) | +37.4% worse |
| test_primary/surface_rel_l2_pct | 6.212% (ep 181) | — | — |
| External ref | Transolver-3: 3.71% / AB-UPT: 3.82% | | |

**W&B run:** `yblf2j7a`

**Command:**
```bash
  --grad-centralization \
  --wandb-group dm-grad-centralization \
  --agent edward --wandb-name "edward/dm-grad-centralization"
```

### What happened

**Gradient centralization did NOT help — negative result.** The run diverged catastrophically after epoch ~200.

**Training trajectory:**
- Epochs 1–181: Healthy convergence, reached best val 5.492% at epoch 181 (grad_norm ~0.21)
- Epoch ~200: Instability onset — val jumped to 8.75%, grad_norm spiked to 1.78
- Epoch ~250: Full divergence — grad_norm exploded to **351,932**, val degraded to 48.6%
- Epochs 250–527: Oscillating between partial recovery and further explosions (grad_norm 234k at ep500). Never recovered below 28%.

**Why it failed:** GC subtracts the per-tensor gradient mean, which preserves gradient direction but removes the DC component. For the cosine annealing schedule (T_max=30), this creates a fatal interaction: when LR restarts at the top of a cosine cycle (~ep 181 ≈ 6×T_max), the centered gradients lose the stabilizing effect of the mean, and the full-LR restart kicks the model out of its basin. The baseline without GC handles these restarts gracefully because the full gradient (with mean) provides a more conservative update direction. The gradient norm explosion (0.21 → 351,932) at the restart boundary confirms this mechanism.

**Key observation:** GC's best (5.492%) is 37% worse than baseline (3.997%) even before divergence. The baseline reaches a deeper basin by training stably through many more cosine cycles. GC prevents this by making the model fragile at cycle boundaries.

### Suggested follow-ups
- GC might work with a monotonically decaying LR schedule (no restarts), but this conflicts with the established benefit of cosine annealing on DrivAerML
- GC + gradient clipping (e.g. gc=1.0) could prevent the norm explosions, but the pre-divergence best was already worse than baseline, suggesting limited upside
- The result is consistent with prior DrivAerML findings that explicit regularizers (WD, gc=2.0) tend to hurt or destabilize training on this dataset

tcapelle:
ADVISOR: Closing PR #3112 because ADVISOR: Gradient centralization reaches 5.492% (ep181) then diverges catastrophically at ep200 when cosine restart hits (grad norms 0.21→351,932). Even the pre-divergence best is 43% worse than 3.833%. GC removes the gradient DC component, making the model fragile at LR restart boundaries. Consistent with the pattern: explicit gradient modifications consistently hurt DM training. Closing..

PR #3086, opened 2026-04-22
DrivAerML: CosineAnnealingWarmRestarts T_mult=2 (SGDR)

tcapelle:
## Hypothesis

SGDR (Loshchilov & Hutter 2017) with T_mult=2 doubles the restart period after each restart (T0=30, then 60, then 120). This allows short early exploration cycles followed by long settling phases. PR #2895 (mugen) found a transient DrivAerML minimum using this approach but couldn't capture it — with best-checkpoint saving now in place, this approach can be properly evaluated.

## Instructions

Implement `CosineAnnealingWarmRestarts(optimizer, T_0=30, T_mult=2)`:
```python
scheduler = torch.optim.lr_scheduler.CosineAnnealingWarmRestarts(optimizer, T_0=30, T_mult=2)
```

Add `--cosine-t-mult 2` flag (keep existing `--cosine-t-max 30` as `T_0`).

```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 \
  --cosine-t-mult 2 \
  --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 \
  --model-heads 8 --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
  --wandb_group dm-sgdr-sweep
```

**Critical:** Must save best checkpoint (save model whenever `val_primary/surface_rel_l2_pct` improves). Report from best checkpoint, not terminal epoch.

## Reporting Requirements
- Best val `surface_rel_l2_pct` + epoch
- Test at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML baselines:**
- val best: `3.997%` @ ep467 (PR #2898, W&B: `bht6h42t`, 4L/512d, AdamW lr=5e-4, T_max=30)
- test best: `6.244%` (PR #2648, 4L/320d, truncated eval — OLDER CONFIG)
- External targets: AB-UPT 3.82% | Transolver-3: 3.71%

PR #3045, opened 2026-04-22
DrivAerML: T_max cosine period sweep at exact champion config

tcapelle:
## Hypothesis

The DrivAerML champion config (#2898, 3.997% val) uses T_max=30 cosine annealing. On AirfRANS, switching from T_max=10 to T_max=50 produced a **19.4% improvement** (#2951, 0.000598→0.000482) — the single largest AF gain from a schedule change. **T_max=50 has never been tested on DrivAerML at the actual champion config.** The prior negative result (#2919, T_max=40/50) confounded the T_max change with gc=1.0+WD=1e-2, which are NOT part of the champion config.

Longer cosine periods allow the model to descend more deeply into each loss basin before the LR restart kicks it back up. At 4L/512d with batch_size=1 and 50k random surface subsamples, the optimization landscape is noisy — longer cosine periods may help average out this noise by spending more steps in the low-LR descent phase.

### Runs (4 DM runs — T_max sweep)

Use `--wandb_group dm_tmax_champion` for all runs. **All runs use the EXACT champion config (no --grad-clip, no --weight-decay).** This is critical — the champion #2898 does NOT use gc or WD.

**Run 1 — T_max=30 control with FULL EVAL (paper-facing baseline replication):**
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --wandb_group dm_tmax_champion
```
**NOTE: NO --max-eval-batches.** This is the paper-facing baseline replication with full-coverage eval.

**Run 2 — T_max=50 (AF breakthrough schedule):**
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 50 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
[...]

PR #3116, opened 2026-04-22
DrivAerML: SGDR T_0=15, T_mult=2 (progressive cosine restarts)

tcapelle:
## Hypothesis

SGDR (CosineAnnealingWarmRestarts) with T_0=15 and T_mult=2 starts with short 15-epoch cycles for rapid exploration, then doubles to 30 (matching champion T_max), then 60 for deep settling. This progressive schedule combines fast early exploration with longer late-training convergence.

megumi (#3086) is testing T_0=30, T_mult=2. This tests T_0=15 — shorter initial cycles that match the fast micro-cycling the model seems to prefer early, with later cycles matching and exceeding the champion period.

**Schedule:** cycles at epochs 0-15, 15-45, 45-105, 105-225, 225-465...

## Instructions

Implement `CosineAnnealingWarmRestarts(optimizer, T_0=15, T_mult=2)`:

```python
scheduler = torch.optim.lr_scheduler.CosineAnnealingWarmRestarts(optimizer, T_0=15, T_mult=2)
```

Add `--cosine-t-mult 2` and set T_0 via `--cosine-t-max 15`.

```bash
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 15 \
  --cosine-t-mult 2 \
  --wandb_group dm-sgdr-sweep
```

**Warning from #3045:** T_max=15 alone diverged without gc (val=14.17%, grad norm 1.5e8). But SGDR T_mult=2 means the restart at epoch 15 jumps LR from trough back to peak — this may trigger the same explosion. If divergence occurs before ep30, add `--grad-clip 1.0` and document.

**Critical:** Must save best checkpoint (save model whenever val improves). Report from best checkpoint, not terminal epoch.

## Reporting Requirements
- Whether divergence occurred (and at which cycle boundary)
- Test at best val checkpoint
- W&B run ID

## Baseline

**DrivAerML champion:** val=**3.997%** @ ep467 (PR #2898, T_max=30, no SGDR)
- W&B: `bht6h42t`

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -160,6 +160,7 @@ class TrainConfig:
     residual_prediction: bool = False
     re_stratified_sampling: bool = False
     cosine_t_max: int = 150
+    cosine_t_mult: int = 0
     geometry_points: int = 25_000
     geometry_supernodes: int = 4_096
     surface_anchor_points: int = 8_000
@@ -1625,7 +1626,12 @@ def main() -> None:
     scheduler = None
     if config.cosine_t_max > 0:
         base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
-        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        if config.cosine_t_mult > 0:
+            scheduler = torch.optim.lr_scheduler.CosineAnnealingWarmRestarts(
+                base_optimizer, T_0=config.cosine_t_max, T_mult=config.cosine_t_mult
+            )
+        else:
+            scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
 
     ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
     anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None


tcapelle:
STUDENT sanji:

## Results

### SGDR T_0=15, T_mult=2 (no grad clip)

| Metric | Value |
|--------|-------|
| **Best val `surface_rel_l2_pct`** | **12.846%** @ epoch 38 |
| Test `surface_rel_l2_pct` | N/A (training killed due to catastrophic divergence) |
| Baseline (champion) | **3.997%** @ ep467 |
| Relative to baseline | **+221%** (3.2x worse) |
| W&B run ID | `gm1o0yvp` |
| Total epochs | 201 (killed at ep201, ~140 min of 360-min budget) |
| Peak GPU memory | ~6.2 GB (GPU 0) |

### Command

```bash
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 15 \
  --cosine-t-mult 2 \
  --wandb-group dm-sgdr-sweep \
  --agent sanji --wandb-name "sanji/dm-sgdr-t0-15-tmult2"
```

### Divergence Analysis

The SGDR schedule worked beautifully for the first three cycles, then catastrophically diverged. Here's the cycle-by-cycle breakdown (scheduler steps per-batch, 394 batches/epoch):

| Cycle | Epoch range | Best val | Grad norm | Outcome |
|-------|-------------|----------|-----------|---------|
| ~9 (ends ep9.7) | 1-9 | 28.0% | 0.9-6.5 | Healthy descent |
| ~10 (ep9.7-19.5) | 10-19 | 18.6% | 0.9-3.2 | Deeper basin found |
| ~11 (ep19.5-38.9) | 20-38 | **12.846%** | 0.5-2.5 | **Overall best** |
| ~12 (ep38.9-77.9) | 39-77 | 25.7% | 0.8→5,371 | Destabilized |
| ~13 (ep77.9-155.9) | 78-155 | 22.8% | 1,483→9,636 | Semi-diverged |
| ~14 (ep155.9+) | 156-201 | 33.9% | 9,636→9,000,000,000 | **Full catastrophic divergence** |

**Key observations:**
- The restart at ep39 (LR 0.000003 → 0.000500) after achieving 12.846% was the inflection point. Val immediately spiked to 72%.
- Grad norms escalated from single digits (ep1-38) to thousands (ep39-155) to billions (ep156+).
- Each subsequent SGDR restart amplified the instability — the longer cycles couldn't stabilize because they started from an already-corrupted state.
- The model was killed at ep201 (grad_norm = 3.8 billion) to free GPU resources.

### What happened

**Negative result.** SGDR T_0=15, T_mult=2 without grad clipping diverges catastrophically on DrivAerML. The progressive cosine restart schedule works well for the first ~39 epochs (3 short cycles), finding a 12.846% basin that was competitive with the champion at the same epoch count. But the restart at the cycle boundary after ep38 triggers an irreversible gradient explosion. Without grad clipping, the model enters a divergence cascade that worsens with each restart.

This confirms the PR's warning about T_max=15 divergence from #3045 — the SGDR restarts amplify the same instability mechanism. The short initial cycles actually helped (fast exploration), but the restart from deep basins back to peak LR is destructive.

### Suggested follow-ups

1. **SGDR T_0=15, T_mult=2 + `--grad-clip 1.0`**: The obvious next step. The first 38 epochs showed SGDR finds good basins fast — grad clipping could prevent the restart-induced explosion and let later long cycles push deeper.
2. **SGDR T_0=30, T_mult=2** (megumi #3086 variant): Starting with T_0 matching the champion's T_max=30 might avoid the early instability while still getting progressive lengthening.
3. **SGDR with eta_min > 0**: Instead of annealing LR all the way to 0 (which creates maximum shock on restart), keep a floor like `eta_min=1e-5` to reduce the restart magnitude.

tcapelle:
ADVISOR: Closing PR #3116 because ADVISOR: Closing — SGDR T_0=15/T_mult=2 is a dead end on DrivAerML. Best val 12.846% at ep38 (+221% above 3.997% baseline), crashed at ep~200 with grad norms in the billions. Same failure mode as megumi's SGDR (#3086): LR restart from ~0 to 5e-4 destabilizes the model at cycle boundaries. First 3 cycles (ep0-38) showed healthy training, but the restart at ep~39 immediately caused 72% val spike with no recovery. SGDR with any T_0/T_mult combination is likely dead on DM — the periodic restarts are incompatible with the no-gc stability requirement..

PR #3120, opened 2026-04-22
DrivAerML: RAdam optimizer (rectified Adam with built-in warmup)

tcapelle:
## Hypothesis

RAdam (Rectified Adam, Liu et al. 2019) adds a rectification term that corrects the adaptive learning rate variance in the early training phases, effectively providing a principled built-in warmup without needing an explicit warmup schedule. This addresses the known DrivAerML instability issue (gradient explosions from cold-start high LR) from a different angle than linear warmup.

DrivAerML is extremely sensitive to gradient stability (T_max=30 is the only stable cosine period; all other T_max values diverge). RAdam's variance rectification may provide more stable early-epoch dynamics, potentially allowing the model to reach a better basin than standard AdamW.

Reference: "On the Variance of the Adaptive Learning Rate and Beyond" (Liu et al. 2019, ICLR 2020).

## Instructions

RAdam is available in PyTorch as `torch.optim.RAdam`. Add it as an optimizer option in train.py alongside adamw/lion.

**Run 1: RAdam lr=5e-4 (same as AdamW champion)**
```bash
  --dataset drivaerml \
  --optimizer radam \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group dm-radam-optimizer
```

**Run 2: RAdam lr=7e-4 (slightly higher, RAdam's built-in warmup may stabilize it)**
```bash
  --dataset drivaerml \
  --optimizer radam \
  --lr 7e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group dm-radam-optimizer
```

To add RAdam support in train.py: `optimizer = torch.optim.RAdam(params, lr=args.lr)`. No --grad-clip, no --weight-decay (matching champion regularization strategy). T_max=30 (only stable cosine period on DrivAerML).

Report best val_primary/surface_rel_l2_pct for both runs from W&B.

## Baseline

- **W&B run:** bht6h42t

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -546,6 +546,8 @@ def build_optimizer(params, config: TrainConfig):
         optimizer = Lion(params, lr=config.lr, weight_decay=config.weight_decay)
     elif config.optimizer == "adamw":
         optimizer = torch.optim.AdamW(params, lr=config.lr, weight_decay=config.weight_decay)
+    elif config.optimizer == "radam":
+        optimizer = torch.optim.RAdam(params, lr=config.lr, weight_decay=config.weight_decay)
     else:
         raise ValueError(f"Unknown optimizer: {config.optimizer}")
     if config.use_lookahead:


tcapelle:
STUDENT senku:

## Results

### Run 1: RAdam lr=5e-4 (same as AdamW champion)
- **val_primary/surface_rel_l2_pct:** 19.95% at epoch 40 (best), **diverged** at epoch 43 (grad explosion to 398K norm)
- **Baseline:** 3.997% at epoch 467 (AdamW) — **RAdam 4.99x worse** before divergence
- **W&B run:** hh0vm4a2
- **Peak memory:** ~6 GB VRAM
- **Command:**
```bash
  --dataset drivaerml --optimizer radam --lr 5e-4 --cosine-t-max 30 --no-use-ema \
  --wandb-group dm-radam-optimizer --agent senku --wandb-name "senku/radam-lr5e4"
```

**Timeline:**
- Epochs 1-40: Healthy descent with normal cosine oscillation. Best trough values: ep13=27.66%, ep24=23.93%, ep29=22.55%, ep40=19.95%
- Epoch 41: Grad norm spike to 26.2 (was ~1.1), val jumps to 25.7%
- Epoch 42: Grad norm 31.3, loss spikes to 0.133
- Epoch 43: **Catastrophic explosion** — grad norm 398,470, val 72.7%
- Epoch 44-46: Model stabilized at ~63-65% (bad basin), no recovery

### Run 2: RAdam lr=7e-4 (higher LR, testing RAdam's warmup advantage)
- **val_primary/surface_rel_l2_pct:** 20.79% at epoch 38 (best), **diverged to NaN** at epoch 39
- **Baseline:** 3.997% at epoch 467 (AdamW) — **RAdam 5.20x worse** before divergence
- **W&B run:** tkx91w4v
- **Peak memory:** ~6 GB VRAM
- **Command:**
```bash
  --dataset drivaerml --optimizer radam --lr 7e-4 --cosine-t-max 30 --no-use-ema \
  --wandb-group dm-radam-optimizer --agent senku --wandb-name "senku/radam-lr7e4"
```

**Timeline:**
- Epochs 1-38: Healthy descent. Best trough values: ep15=26.86%, ep22=24.50%, ep33=21.51%, ep38=20.79%
- Epoch 39: **Immediate NaN** at cosine LR peak (6.33e-4). No recovery possible.

## What happened

**Both RAdam runs diverge catastrophically around epoch 38-43.** The hypothesis that RAdam's variance rectification would provide more stable early-epoch dynamics is partially confirmed — both runs showed smooth, stable convergence through the first ~37 epochs. However, RAdam's rectification only helps the **initial** warmup phase (first few epochs). Once the rectification term ρ exceeds the threshold and RAdam becomes identical to standard Adam, it provides no additional stability for the cosine LR cycling that causes DrivAerML's gradient instability.

**Key observations:**
1. **RAdam ≠ AdamW for long-term stability.** AdamW's decoupled weight decay provides implicit regularization that RAdam (with WD=0) lacks. This manifests as catastrophic divergence once the model reaches a critical sensitivity to LR peaks.
2. **RAdam's rectification is a warmup replacement, not a stability mechanism.** The first 1-3 epochs used low effective LR (RAdam's variance correction), but after that it behaves like standard Adam. The cosine LR peaks (every 30 epochs) require persistent stability, not just initial warmup.
3. **Divergence timing correlates with cosine peaks.** Run 2 (7e-4) diverged at epoch 39 (LR peak in 2nd cycle). Run 1 (5e-4) diverged at epoch 41-43 (also rising toward peak in 2nd cycle). Both survived cycle 1 but failed in cycle 2.
4. **Lower LR delays but doesn't prevent divergence.** Run 1 (5e-4) lasted 3 more epochs than Run 2 (7e-4), but still diverged.

**Comparison with AdamW baseline:** The AdamW champion (3.997% at ep467) runs with the same lr=5e-4 and T_max=30 but survives hundreds of cosine cycles. AdamW's decoupled weight decay (even at the default 1e-4) provides a regularization effect that prevents the gradient norm escalation RAdam suffers.

## Suggested follow-ups
1. **RAdam + weight decay:** Try RAdam with `--weight-decay 1e-4` to match AdamW's regularization. The WD=0 config (matching the champion which uses no explicit WD flag) may be the root cause of instability, not RAdam itself.
2. **RAdam + grad clipping:** RAdam + `--grad-clip 1.0` could prevent the gradient explosion while keeping the rectification benefit.
3. **RAdam + grad clip + WD compound:** The most promising variant would combine all three stabilization mechanisms.
4. **Shorter cosine period (T_max=15):** Reduce the LR peak amplitude to see if RAdam survives with less aggressive cycling.

tcapelle:
ADVISOR: Closing PR #3120 because ADVISOR: Closing — RAdam is a dead end on DrivAerML. lr=5e-4 best val 19.95% at ep40 then diverged; lr=7e-4 best 20.79% then NaN at ep39. Both ~5x worse than 3.997% baseline. RAdam's variance rectification only provides early-epoch warmup — it offers no protection at recurring cosine LR peaks which is where DM divergence happens. The experiment conclusively shows DM instability is LR-peak driven, not warmup-driven. Good mechanistic analysis, senku..

PR #3125, opened 2026-04-22
DrivAerML: SWA at cosine troughs (selective weight averaging)

tcapelle:
## Hypothesis

EMA failed on DrivAerML (3 experiments, all 2-3x worse) because it continuously averages weights INCLUDING those from high-LR gradient explosion epochs. SWA (Stochastic Weight Averaging, Izmailov et al. 2018) is fundamentally different: it only averages weights at LR trough epochs (cosine minima), when the model is at its most stable.

With T_max=30, the cosine LR hits its minimum every 30 epochs (ep30, ep60, ep90, ...). At these points, the model is most stable (lowest gradient norms, best instantaneous val). SWA collects ONLY these trough-epoch snapshots and averages them, ignoring all the high-LR instability in between.

This addresses the DrivAerML-specific failure mode: the champion trains for 467 epochs through ~15 cosine cycles. The gradient spikes at LR peaks (which killed EMA) are simply excluded from the SWA average.

Reference: "Averaging Weights Leads to Wider Optima and Better Generalization" (Izmailov et al. 2018, UAI).

## Instructions

Implement manual SWA that collects model weights at cosine trough epochs. PyTorch has `torch.optim.swa_utils.AveragedModel` and `SWALR`, but for this experiment, implement it manually to control exactly WHEN weights are collected:

1. After each epoch, check if the current epoch is a cosine trough (epoch % T_max == 0, i.e., ep30, ep60, ep90, ...)
2. If yes, add the current model weights to a running average (equal weight for all collected snapshots)
3. At eval time, evaluate BOTH the current model AND the SWA-averaged model
4. Log SWA metrics as `swa_val_primary/surface_rel_l2_pct`

**Run 1: SWA starting from epoch 60 (after 2 full cosine cycles)**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --swa-start-epoch 60 \
  --swa-collect-interval 30 \
  --wandb_group dm-swa-troughs
```

Implementation notes:
- Collect weights at epochs 60, 90, 120, 150, 180, ... (every T_max=30 epochs starting from epoch 60)
- Use simple equal-weight averaging: `swa_weights = (swa_weights * n + current_weights) / (n + 1)`
- BatchNorm update: after each SWA collection, run one pass through training data with `torch.optim.swa_utils.update_bn()` to update BN stats for the averaged model
- Report both standard best val AND SWA best val

No --grad-clip, no --weight-decay. Report best val_primary/surface_rel_l2_pct for both standard and SWA models.

## Baseline

- **W&B run:** bht6h42t

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -86,6 +86,64 @@ class EMAWithWarmup:
         self.backup = None
 
 
+class ManualSWA:
+    """Manual SWA: collect and average model weights at specific epochs (Izmailov et al. 2018)."""
+
+    def __init__(self, model: torch.nn.Module):
+        self.n_collected = 0
+        self.swa_params: dict[str, torch.Tensor] | None = None
+        self.backup: dict[str, torch.Tensor] | None = None
+
+    @staticmethod
+    def _clean(name: str) -> str:
+        return name.replace("_orig_mod.", "")
+
+    @torch.no_grad()
+    def collect(self, model: torch.nn.Module) -> None:
+        self.n_collected += 1
+        n = self.n_collected
+        if self.swa_params is None:
+            self.swa_params = {
+                self._clean(name): param.detach().clone()
+                for name, param in model.named_parameters()
+                if param.requires_grad
+            }
+        else:
+            for name, param in model.named_parameters():
+                key = self._clean(name)
+                if param.requires_grad and key in self.swa_params:
+                    self.swa_params[key].mul_((n - 1) / n).add_(param.detach(), alpha=1.0 / n)
+
+    @torch.no_grad()
+    def store(self, model: torch.nn.Module) -> None:
+        if self.swa_params is None:
+            return
+        self.backup = {
+            self._clean(name): param.detach().clone()
+            for name, param in model.named_parameters()
+            if param.requires_grad and self._clean(name) in self.swa_params
+        }
+
+    @torch.no_grad()
+    def copy_to(self, model: torch.nn.Module) -> None:
+        if self.swa_params is None:
+            return
+        for name, param in model.named_parameters():
+            key = self._clean(name)
+            if param.requires_grad and key in self.swa_params:
+                param.data.copy_(self.swa_params[key])
+
+    @torch.no_grad()
+    def restore(self, model: torch.nn.Module) -> None:
+        if self.backup is None:
+            return
+        for name, param in model.named_parameters():
+            key = self._clean(name)
+            if param.requires_grad and key in self.backup:
+                param.data.copy_(self.backup[key])
+        self.backup = None
+
+
 LEGACY_VAL_ALIAS = {
     "val_in_dist": "p_in",
     "val_ood_cond": "p_oodc",
@@ -124,6 +182,8 @@ class TrainConfig:
     use_lookahead: bool = True
     use_ema: bool = True
     ema_decay: float = 0.9999
+    swa_start_epoch: int = 0
+    swa_collect_interval: int = 0
     num_workers: int = -1
     pin_memory: bool = True
     persistent_workers: bool = True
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT faye:

## Results

### Key Metrics

| Metric | Baseline | SWA Experiment | Delta |
|--------|----------|---------------|-------|
| `val_primary/surface_rel_l2_pct` (best raw) | **3.997%** (ep467) | **14.298%** (ep89) | +10.30 (3.6x worse) |
| `test_primary/surface_rel_l2_pct` (best ckpt) | — | **15.277%** | — |
| SWA `val_primary/surface_rel_l2_pct` (final) | — | **88.98%** (16 collections) | catastrophic |
| SWA `test_primary/surface_rel_l2_pct` | — | **89.07%** | catastrophic |

### SWA Collection History

| Epoch | Raw val% | SWA val% | n | grad_norm |
|-------|----------|----------|---|-----------|
| 60 | 21.37 | 21.37 | 1 | 4.25 |
| 90 | 14.95 | 35.73 | 2 | 0.93 |
| 120 | 19.44 | 42.30 | 3 | 3.37 |
| 150 | 34.42 | 42.71 | 4 | 5.39 |
| 180 | 29.29 | 53.11 | 5 | 231.87 |
| 210 | 26.31 | 57.26 | 6 | 18.79 |
| 240 | 31.12 | 59.60 | 7 | 78.15 |
| 270 | 34.52 | 55.50 | 8 | 547.47 |
| 300 | 40.85 | 55.86 | 9 | 637.69 |
| 330 | 46.73 | 59.75 | 10 | 1.02M |
| 360 | 36.23 | 62.91 | 11 | 111K |
| 390 | 61.32 | 63.28 | 12 | 4.17M |
| 420 | 58.08 | 64.90 | 13 | 17.78M |
| 450 | 51.16 | 71.36 | 14 | 2.99M |
| 480 | 52.41 | 80.35 | 15 | 1.02B |
| 510 | 58.78 | 88.98 | 16 | 16.64B |

### Experiment Details

**Command:**
```bash
CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=360 \
  python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 \
  --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --swa-start-epoch 60 --swa-collect-interval 30 \
  --wandb-name "faye/dm-swa-cosine-troughs" --wandb-group "dm-swa-troughs" --agent faye
```

- **W&B run:** [hgnyelv3](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/hgnyelv3)
- **Training duration:** ~6 hours (360-min timeout), 521 epochs, 16 SWA collections
- **Peak memory:** ~2.3 GB RAM (GPU memory released on exit, ~98 GB H100 available)

### What Happened — Negative Result

**SWA failed catastrophically.** The averaged weights produced monotonically degrading performance — from 21.4% at the first collection to 89.0% after 16 collections (near random).

**Root cause analysis:**

1. **Gradient explosion without grad clipping**: DrivAerML training without EMA experiences periodic gradient explosions (first at ~ep173, grad_norm=2122). Without grad clipping, these explosions push the model into completely different loss basins at each cosine trough.

2. **Weight averaging across incompatible basins**: SWA assumes the collected checkpoints lie in the same wide basin. When the model's weights are scattered across distinct basins (due to gradient explosions destabilizing training), linear averaging produces destructive interference — the averaged model sits in a high-loss saddle region between basins.

3. **Self-reinforcing degradation**: Each new collection with exploded weights further corrupted the running average. The SWA model was irrecoverable after just 2 collections (ep90: 35.73% vs raw 14.95%).

4. **Even the raw model degraded**: Without EMA or grad clipping, the raw model best was only 14.30% at ep89 (before the first gradient explosion), 3.6x worse than the 3.997% baseline with EMA. The model never recovered from the first explosion at ep173.

**Why EMA works but SWA doesn't here**: EMA's exponential decay (α=0.999) naturally downweights old, corrupted weights as the model recovers from explosions. SWA's equal-weight averaging treats all collected snapshots identically, so a single catastrophic snapshot permanently poisons the average.

### Suggested Follow-ups

1. **SWA + grad clipping**: The fundamental issue is gradient explosions, not SWA itself. Repeating this with `--grad-clip 1.0` could stabilize training enough for SWA to work. Without explosions, trough checkpoints might genuinely lie in the same basin.

2. **SWA + EMA**: Instead of averaging raw weights, average EMA weights at cosine troughs. EMA already smooths out explosion effects, so the averaged EMA checkpoints might be more compatible.

3. **Filtered SWA**: Only include snapshots whose val metric is below a threshold (e.g., reject any snapshot with val > 20%). This would avoid poisoning the average with post-explosion weights.

4. **Late-start SWA**: Start SWA much later (e.g., last 50-100 epochs only) to avoid collecting from the unstable early-training regime. Combined with grad clipping, this could capture the actual convergence basin.

tcapelle:
ADVISOR: Closing PR #3125 because ADVISOR: SWA reaches 14.30% raw (3.7x worse than baseline), then SWA averaging degrades to 88.98% — near-random. Root cause: without gc/EMA, gradient explosions scatter weights across incompatible loss basins, then SWA's equal-weight averaging poisons the running average with post-explosion checkpoints. Your analysis is correct and thorough. SWA requires stable training to work — this config can't provide it. Closing. The EMA champion (#3072) already provides continuous weight averaging that subsumes the SWA benefit..

PR #3126, opened 2026-04-22
DrivAerML: cosine eta_min sweep (non-zero LR floor)

tcapelle:
## Hypothesis

The champion uses CosineAnnealingLR with the default eta_min=0, meaning the learning rate drops to exactly 0 at each cosine trough (every 30 epochs). During these near-zero LR phases, the model barely learns — wasting training time.

Setting a non-zero minimum LR (eta_min) keeps the optimizer active even at cosine troughs. This gives more productive training per epoch without changing the cosine cycling pattern that DM critically depends on. The LR still oscillates but now between 5e-4 and eta_min instead of 5e-4 and 0.

Since the champion trains for 467 epochs (~15 cosine cycles), even a small improvement in each cycle compounds significantly over the full run.

## Instructions

Test two eta_min values at the exact champion config:

**Run 1: eta_min=1e-5**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --cosine-eta-min 1e-5 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group dm-cosine-eta-min
```

**Run 2: eta_min=5e-5**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --cosine-eta-min 5e-5 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group dm-cosine-eta-min
```

If train.py doesn't have a --cosine-eta-min flag, add one and pass it to CosineAnnealingLR's eta_min parameter. No --grad-clip, no --weight-decay. Report best val_primary/surface_rel_l2_pct.

## Baseline

- **Config:** 4L/512d/8H, AdamW lr=5e-4, T_max=30 (eta_min=0 default), no-EMA, Fourier
- **W&B run:** bht6h42t

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -160,6 +160,7 @@ class TrainConfig:
     residual_prediction: bool = False
     re_stratified_sampling: bool = False
     cosine_t_max: int = 150
+    cosine_eta_min: float = 0.0
     geometry_points: int = 25_000
     geometry_supernodes: int = 4_096
     surface_anchor_points: int = 8_000
@@ -1625,7 +1626,7 @@ def main() -> None:
     scheduler = None
     if config.cosine_t_max > 0:
         base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
-        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max, eta_min=config.cosine_eta_min)
 
     ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
     anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None


tcapelle:
STUDENT gohan:

## Results

### Run 1: eta_min=1e-5 (W&B: kg7eolnk)

| Metric | Baseline (eta_min=0) | eta_min=1e-5 | Delta |
|--------|---------------------|--------------|-------|
| val_primary/surface_rel_l2_pct | **3.997%** (ep 467) | **7.58%** (ep 112) | +3.58% ❌ |

**Command:**
```bash
  --cosine-eta-min 1e-5 --no-use-ema --enable-fourier --model-layers 4 \
  --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 \
  --agent gohan --wandb-name "gohan/dm-cosine-eta-min-1e5" --wandb-group "dm-cosine-eta-min"
```

**Peak memory:** ~5985 MiB

**Training history:**
- Converged smoothly to ~9.6% by epoch 82, then to best of **7.58%** at epoch 112
- Mild instability at epochs 31-52 (val spiked to ~34%, recovered fully)
- **Catastrophic instability from epoch 117:** grad_norm spiked to 60.9, val to 72.8%. Slowly recovering but at epoch 144 still around 15-16%
- Run still in progress but unlikely to beat baseline given instability pattern

### Run 2: eta_min=5e-5 (W&B: 1n8peifp)

| Metric | Baseline (eta_min=0) | eta_min=5e-5 | Delta |
|--------|---------------------|--------------|-------|
| val_primary/surface_rel_l2_pct | **3.997%** (ep 467) | **6.80%** (ep 98) | +2.80% ❌ |

**Command:**
```bash
  --cosine-eta-min 5e-5 --no-use-ema --enable-fourier --model-layers 4 \
  --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 \
  --agent gohan --wandb-name "gohan/dm-cosine-eta-min-5e5" --wandb-group "dm-cosine-eta-min"
```

**Peak memory:** ~5985 MiB

**Training history:**
- Converged smoothly and faster than 1e-5 run, reaching best of **6.80%** at epoch 98
- **Catastrophic instability from epoch 109:** val spiked to 44%, then recurring massive grad_norm spikes (up to 907). Cannot sustain recovery — oscillating between 20-60%
- Run still in progress but in persistent divergence cycle

## What happened

**Neither run beats the baseline.** Both eta_min variants converge faster in early training but then hit catastrophic training instability:

1. **The non-zero LR floor appears to destabilize long training.** The baseline (eta_min=0) trained stably for 467 epochs, reaching 3.997%. Both non-zero eta_min runs diverged before epoch 120 — well before reaching the baseline's convergence point.

2. **The instability pattern is grad_norm explosions** — sudden spikes of 10x-900x normal, causing loss and val to spike dramatically. The 5e-5 run (higher floor) diverged earlier and more severely, suggesting the higher minimum LR amplifies instability.

3. **Without grad clipping, the non-zero LR floor leaves the optimizer "active" at cosine troughs**, which may amplify any noise or gradient anomalies that would otherwise be dampened when LR → 0.

4. **Early convergence was promising** — both runs tracked below where the baseline would be at the same epoch count. But the instability wipes out these gains.

## Suggested follow-ups

- **eta_min + grad_clip:** The instability suggests grad clipping would pair well with non-zero eta_min. Try eta_min=5e-5 with grad_clip=1.0 to prevent the catastrophic spikes while retaining the faster convergence.
- **Smaller eta_min (1e-6):** A much smaller floor might be enough to keep training productive at troughs without triggering instability.
- **eta_min with EMA:** EMA could smooth over the instability, preserving the best weights through gradient spikes.

tcapelle:
ADVISOR: Closing PR #3126 because Both eta_min variants (1e-5 and 5e-5) failed to beat baseline (3.997%). Best results confirmed via W&B: kg7eolnk best=7.584% at epoch 112; 1n8peifp best=6.800% at epoch 98. Both suffered catastrophic gradient explosions post-convergence — 40-42% of all training steps above 20% error, no recovery trajectory.

Root cause: without grad clipping, the non-zero LR floor removes the effective rest the optimizer gets at eta_min=0 cosine troughs, allowing gradient noise to accumulate and eventually explode. Higher floor = faster and more severe instability (5e-5 diverged earlier and harder than 1e-5).

Positive signal: early convergence was faster than baseline (best ~7-6.8% by epoch 98-112 vs baseline needing 467 epochs to reach 3.997%). A follow-up testing eta_min=5e-5 with grad_clip=1.0 is the logical next step — the combination could stabilize the fast early convergence while preventing gradient explosions..

PR #3131, opened 2026-04-22
DrivAerML: OneCycleLR schedule (single warmup-then-decay, no restarts)

tcapelle:
## Hypothesis

Every DrivAerML schedule failure shares the same root cause: **periodic LR restarts from near-zero back to peak LR**. CosineAnnealingLR restarts every T_max=30 epochs, SGDR restarts at exponentially growing intervals — both create repeated high-LR episodes that cause gradient explosions without gc.

OneCycleLR (Smith & Topin 2019) is fundamentally different: it does ONE warmup from low LR to max LR (first ~30% of training), then monotonically decays to near-zero for the remaining ~70%. After the single peak, LR only goes down — **no periodic restarts, no repeated gradient shocks**.

This directly addresses the ~120-step high-LR stability limit we've identified. With OneCycleLR, the model spends only the first ~30% of training at elevated LR, then the entire remaining 70% at progressively lower (stable) LR. The champion at T_max=30 spends roughly 50% of each cycle at elevated LR, repeated 15 times.

Reference: "Super-Convergence: Very Fast Training of Neural Networks Using Large Learning Rates" (Smith & Topin, 2019).

## Instructions

Replace CosineAnnealingLR with OneCycleLR. PyTorch provides `torch.optim.lr_scheduler.OneCycleLR`.

**Run 1: OneCycleLR max_lr=5e-4 (same peak as champion)**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --lr-schedule onecycle \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group dm-onecycle-lr
```

**Run 2: OneCycleLR max_lr=1e-3 (higher peak, relying on OneCycle's annihilation)**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 1e-3 \
  --lr-schedule onecycle \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group dm-onecycle-lr
```

Implementation: add `--lr-schedule onecycle` flag. When selected, use:
```python
scheduler = torch.optim.lr_scheduler.OneCycleLR(
    optimizer,
    max_lr=args.lr,
    total_steps=total_steps,  # total_epochs * steps_per_epoch
    pct_start=0.3,            # 30% warmup
    anneal_strategy='cos',    # cosine annealing after peak
    div_factor=25,            # initial_lr = max_lr/25
    final_div_factor=1e4      # min_lr = initial_lr/10000
)
```
Step the scheduler per BATCH (not per epoch) — OneCycleLR requires per-step scheduling. No --grad-clip, no --weight-decay. Report best val_primary/surface_rel_l2_pct.

## Baseline

- **Config:** 4L/512d/8H, AdamW lr=5e-4, CosineAnnealingLR T_max=30, no-EMA, Fourier
- **W&B run:** bht6h42t

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -159,6 +159,7 @@ class TrainConfig:
     asinh_scale: float = 0.75
     residual_prediction: bool = False
     re_stratified_sampling: bool = False
+    lr_schedule: str = "cosine"
     cosine_t_max: int = 150
     geometry_points: int = 25_000
     geometry_supernodes: int = 4_096
@@ -1623,8 +1624,24 @@ def main() -> None:
         params.extend(list(anp_head.parameters()))
     optimizer = build_optimizer(params, config)
     scheduler = None
-    if config.cosine_t_max > 0:
-        base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
+    base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
+    if config.lr_schedule == "onecycle":
+        if config.max_train_batches > 0:
+            micro_per_epoch = config.max_train_batches
+        else:
+            micro_per_epoch = len(train_loader)
+        steps_per_epoch = (micro_per_epoch + config.grad_accum_steps - 1) // config.grad_accum_steps
+        total_steps = config.epochs * steps_per_epoch
+        scheduler = torch.optim.lr_scheduler.OneCycleLR(
+            base_optimizer,
+            max_lr=config.lr,
+            total_steps=total_steps,
+            pct_start=0.3,
+            anneal_strategy='cos',
+            div_factor=25,
+            final_div_factor=1e4,
+        )
+    elif config.cosine_t_max > 0:
         scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
 
     ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
@@ -1641,6 +1658,11 @@ def main() -> None:
     if config.wandb_name:
         run_config = asdict(config)
         run_config["effective_batch_size"] = config.batch_size * config.grad_accum_steps
+        if config.lr_schedule == "onecycle" and scheduler is not None:
+            run_config["onecycle_total_steps"] = scheduler.total_steps
+            run_config["onecycle_pct_start"] = 0.3
+            run_config["onecycle_div_factor"] = 25
+            run_config["onecycle_final_div_factor"] = 1e4
         run = wandb.init(
             project=os.getenv("WANDB_PROJECT", "senpai-v1"),
             entity=os.getenv("WANDB_ENTITY"),


tcapelle:
STUDENT sanji:

## Results

### Run 1: OneCycleLR max_lr=5e-4

| Metric | Baseline | OneCycleLR 5e-4 | Delta |
|--------|----------|-----------------|-------|
| val_primary/surface_rel_l2_pct | 3.997% (ep 467) | 8.040% (ep 190) | +4.04% (worse) |

- **W&B run:** ed60ojeo
- **Peak memory:** ~6.0 GB
- **Command:** `cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --lr-schedule onecycle --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --wandb-group dm-onecycle-lr`
- **Outcome:** Model improved steadily during early warmup, reached best val=8.04% at epoch 190 (lr=3.58e-4), then destabilized as LR approached the 5e-4 peak. By epoch 250 (lr=4.68e-4), grad norms spiked to 130 and val collapsed to 48.6%. Grad norms reached 25 billion by epoch 450 — effectively dead, heading to NaN.

### Run 2: OneCycleLR max_lr=1e-3

| Metric | Baseline | OneCycleLR 1e-3 | Delta |
|--------|----------|-----------------|-------|
| val_primary/surface_rel_l2_pct | 3.997% (ep 467) | 8.041% (ep 125) | +4.04% (worse) |

- **W&B run:** 24r056y4
- **Peak memory:** ~6.0 GB
- **Command:** `cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 1e-3 --lr-schedule onecycle --no-use-ema --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --wandb-group dm-onecycle-lr`
- **Outcome:** Converged faster during early warmup, reached best val=8.04% at epoch 125 (lr=3.96e-4), then collapsed catastrophically. NaN by epoch 350. Grad norms exploded from 0.47 → 5.9 → 1450 → 4.3e14 → NaN as LR climbed from ~4e-4 to 1e-3.

### What happened

**Both runs failed.** Neither came close to the 3.997% baseline. The best metric for both was ~8.04% — achieved during warmup when LR was in the 3.5-4.0e-4 range.

The core problem: **OneCycleLR's sustained high-LR exposure is fundamentally more damaging than CosineAnnealingLR's brief periodic spikes.** The baseline's CosineAnnealingLR T_max=30 (stepped per optimizer step, not per epoch) creates very brief excursions to 5e-4 lasting only a few steps per cycle, with most time spent at lower LR. OneCycleLR instead holds the LR near its peak for hundreds of epochs around the warmup-to-decay transition.

Both runs destabilized at the same critical LR threshold (~4.5-5.0e-4), regardless of the max_lr setting — confirming the ~5e-4 stability limit identified in prior experiments. The difference is that CosineAnnealingLR's brief spikes above this threshold are tolerable, while OneCycleLR's prolonged exposure is not.

The decay phase (70% of training, epochs 300-999) could not recover the model — by the time LR started decreasing, the gradient explosion had already caused irreversible damage.

### Suggested follow-ups

1. **OneCycleLR with much lower max_lr (~2e-4):** Keep the peak well below the ~5e-4 stability limit. Both runs showed good convergence at LR ≤ 3.5e-4.
2. **OneCycleLR with shorter pct_start (e.g. 0.1):** Reduce the warmup fraction so LR reaches peak faster and spends more time in the stable decay phase.
3. **CosineAnnealingWarmRestarts with longer T_0:** Instead of abandoning periodic restarts entirely, try longer cycles (e.g. T_0=100 or T_0=200 steps) so each high-LR episode is brief but the schedule explores more.

tcapelle:
ADVISOR: Closing PR #3131 because ADVISOR: OneCycleLR is a clear dead end — both runs reach ~8.04% best (2x worse than 3.997% old baseline) then catastrophically diverge. The sustained high-LR warmup phase is even more damaging than cosine restart peaks. Importantly, these runs were without EMA+gc (old regime), so they're testing against a ceiling that's already been superseded. OneCycleLR's fundamental problem is no periodic recovery troughs — confirmed again that cosine troughs are load-bearing for DM stability. Your analysis was thorough and well-reasoned. Reassigning you to a fresh DM experiment on the champion platform..

PR #3143, opened 2026-04-23
DrivAerML: Lookahead optimizer wrapping AdamW

tcapelle:
## Hypothesis

The Lookahead optimizer (Zhang et al., 2019 — https://arxiv.org/abs/1907.08610) wraps any inner optimizer and performs k fast inner steps, then interpolates the weights back toward a set of "slow" weights with step size alpha. This implicit ensemble averaging reduces gradient noise and variance, often improving both convergence speed and final generalization. We hypothesize that wrapping AdamW (our current best optimizer) with Lookahead will reduce the DrivAerML surface_rel_l2_pct below the current 3.997% baseline.

The mechanism: every k=6 steps, slow weights ← slow weights + alpha * (fast weights - slow weights). The fast weights are then reset to the slow weights. This creates a trajectory smoothing effect that can escape sharp minima, similar in spirit to SWA/EMA but applied during optimization rather than post-hoc.

Reference implementation: `torchcontrib.optim.Lookahead` (pip install torchcontrib), or implement in ~10 lines wrapping any optimizer.

## Instructions

Add Lookahead optimizer support to `target/icml2026/train.py`:

1. Add two new CLI flags:
   - `--optimizer lookahead-adamw` (add `"lookahead-adamw"` as a valid choice alongside existing optimizer options)
   - `--lookahead-k INT` (default=6, number of fast inner steps)
   - `--lookahead-alpha FLOAT` (default=0.5, slow weight interpolation rate)

2. When `--optimizer lookahead-adamw` is specified, wrap the AdamW optimizer with Lookahead:
   ```python
   from torchcontrib.optim import Lookahead
   base_optimizer = torch.optim.AdamW(params, lr=args.lr, weight_decay=args.weight_decay)
   optimizer = Lookahead(base_optimizer, k=args.lookahead_k, alpha=args.lookahead_alpha)
   ```
   Install via: `pip install torchcontrib`

3. Call `optimizer.sync_lookahead()` before saving checkpoints to ensure slow weights are persisted (important for correct checkpoint recovery).

4. Run the following training command:

```bash
cd target/icml2026
  --dataset drivaerml \
  --optimizer lookahead-adamw \
  --lookahead-k 6 \
  --lookahead-alpha 0.5 \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group dm-optimizer-lookahead
```

All other hyperparameters (architecture, Fourier features, no-EMA, cosine T_max=30) are identical to the current champion config (PR #2898) so any improvement is attributable solely to the Lookahead wrapper.

## Baseline

Current best DrivAerML metrics (PR #2898 — piccolo, 4L/512d/8H + Fourier + no-EMA + T_max=30, 467 epochs, AdamW lr=5e-4):

| Metric | Value |
|--------|-------|
| val_primary/surface_rel_l2_pct | **3.997%** |

- Baseline W&B run: `bht6h42t`
- External target: <3.71% (AB-UPT, ~500 epochs) — 1.08x gap remaining
- Reproduce command:
```bash
cd target/icml2026
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200
```

Please report `val_primary/surface_rel_l2_pct` (best validation epoch) and the W&B run ID when complete.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -122,6 +122,8 @@ class TrainConfig:
     weight_decay: float = 1e-4
     optimizer: str = "lion"
     use_lookahead: bool = True
+    lookahead_k: int = 6
+    lookahead_alpha: float = 0.5
     use_ema: bool = True
     ema_decay: float = 0.9999
     num_workers: int = -1
@@ -546,11 +548,22 @@ def build_optimizer(params, config: TrainConfig):
         optimizer = Lion(params, lr=config.lr, weight_decay=config.weight_decay)
     elif config.optimizer == "adamw":
         optimizer = torch.optim.AdamW(params, lr=config.lr, weight_decay=config.weight_decay)
+    elif config.optimizer == "lookahead-adamw":
+        base = torch.optim.AdamW(params, lr=config.lr, weight_decay=config.weight_decay)
+        return Lookahead(base, alpha=config.lookahead_alpha, k=config.lookahead_k)
     else:
         raise ValueError(f"Unknown optimizer: {config.optimizer}")
     if config.use_lookahead:
-        optimizer = Lookahead(optimizer)
+        optimizer = Lookahead(optimizer, alpha=config.lookahead_alpha, k=config.lookahead_k)
     return optimizer
+
+
+def sync_lookahead(optimizer) -> None:
+    """Copy slow weights to fast weights so model params reflect the averaged trajectory."""
+    if isinstance(optimizer, Lookahead):
+        for slow_group, fast_group in zip(optimizer._slow_weights, optimizer.param_groups):
+            for slow, fast in zip(slow_group, fast_group["params"]):
+                fast.data.copy_(slow)
 
 
 def resolve_num_workers(config: TrainConfig, dataset_name: str) -> int:
@@ -1637,16 +1650,7 @@ def main() -> None:
     best_model_state: dict[str, torch.Tensor] | None = None
     best_anp_state: dict[str, torch.Tensor] | None = None
 
-    if bundle.spec.name == "tandemfoilset":
-        primary_metric_key = "val_primary/surface_pressure_mae"
-    else:
-        primary_metric_key = f"val_primary/{bundle.spec.default_metric}"
     best_val_metric = float("inf")
-    best_epoch = 0
-    best_model_state: dict[str, torch.Tensor] | None = None
-    best_anp_state: dict[str, torch.Tensor] | None = None
-    best_ema_shadow: dict[str, torch.Tensor] | None = None
-    best_anp_ema_shadow: dict[str, torch.Tensor] | None = None
 
     run = None
     if config.wandb_name:
@@ -1713,17 +1717,9 @@ def main() -> None:
             best_model_state = snapshot_module_state(model)
             best_anp_state = snapshot_module_state(anp_head)
 
-        current_val = eval_metrics.get(primary_metric_key)
+        current_val = eval_metrics.get(best_val_primary_metric_name)
         if current_val is not None and current_val < best_val_metric:
             best_val_metric = current_val
-            best_epoch = epoch
-            best_model_state = {k: v.cpu().clone() for k, v in model.state_dict().items()}
-            if anp_head is not None:
-                best_anp_state = {k: v.cpu().clone() for k, v in anp_head.state_dict().items()}
-            if ema is not None:
-                best_ema_shadow = {k: v.cpu().clone() for k, v in ema.shadow.items()}
-            if anp_ema is not None:
-                best_anp_ema_shadow = {k: v.cpu().clone() for k, v in anp_ema.shadow.items()}
 
         if ema is not None:
             ema.restore(model)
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT thorfinn:

## Results

### Lookahead-AdamW (k=6, alpha=0.5)

| Metric | Lookahead-AdamW | Baseline (AdamW) | Target (AB-UPT) |
|--------|----------------|-----------------|------------------|
| val_primary/surface_rel_l2_pct | **5.75%** | **3.997%** | <3.71% |
| Best epoch | 183 | 467 | ~500 |

**W&B run:** `n4sbafwr`
**Peak GPU memory:** ~6.2 GB

**Command used:**
```bash
cd target/icml2026
  --dataset drivaerml \
  --optimizer lookahead-adamw \
  --lookahead-k 6 \
  --lookahead-alpha 0.5 \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb-group dm-optimizer-lookahead \
  --agent thorfinn \
  --wandb-name "thorfinn/dm-lookahead-adamw-k6-a05"
```

### What happened

**The Lookahead optimizer did not beat the baseline and introduced catastrophic training instability.**

**Pre-collapse convergence (epochs 1–183):** Training converged steadily through the first ~6 cosine cycles, reaching a best of 5.75% at epoch 183. The convergence trajectory was similar to vanilla AdamW but slightly slower — at comparable epoch counts, the model was tracking behind the baseline's convergence curve.

**Catastrophic collapses (epoch 195+):** At epoch 195, the training suffered a catastrophic collapse — loss spiked from 0.008 to 0.578, grad_norm from 0.3 to 13.8, and val metric from ~6% to 53%. The model partially recovered but collapsed again at epoch 205 (val to 66.5%). After the second collapse, the training entered a permanent instability cycle — oscillating between 20-45% with grad norms of 9-11, never recovering to the pre-collapse levels. As of epoch 257, the training remains stuck at ~28%.

**Key observations:**
1. The Lookahead slow-weight interpolation (every k=6 steps) did not prevent the catastrophic collapses — in fact, the trajectory smoothing may have delayed early warning signs of instability while not being strong enough to prevent the eventual blowup.
2. The recurring instability spikes (smaller ones visible at epochs 94, 122, 154, 170 before the catastrophic ones) suggest the Lookahead wrapper amplifies noise in the DrivAerML surface point subsampling rather than smoothing it.
3. The baseline used vanilla AdamW with default `use_lookahead=True` (k=5, alpha=0.5). The main difference here was k=6 vs k=5, plus removing the `use_lookahead` double-wrapping. The instability may stem from k=6 allowing more divergence between fast and slow weights.

**Bug fix included:** Fixed a pre-existing `UnboundLocalError` from a merge artifact where `primary_metric_key` was used as both a function and a local variable. This was blocking all training runs on this branch.

### Suggested follow-ups

- **Gradient clipping:** Adding `--grad-clip 1.0` might prevent the catastrophic collapses observed here. The baseline didn't use grad clipping either, but it also didn't collapse — so the Lookahead wrapper may be creating conditions that require it.
- **Lower k:** Try k=3 or k=4 to keep slow weights closer to fast weights, reducing the divergence that may contribute to instability.
- **Lookahead + EMA:** The baseline uses `use_lookahead=True` by default with EMA. Combining Lookahead with EMA (`--use-ema`) might provide better trajectory smoothing.
- **This is a negative result for k=6 without grad clipping.** The hypothesis that Lookahead would improve convergence and generalization did not hold for DrivAerML at this configuration.

tcapelle:
ADVISOR: Closing PR #3143 because ADVISOR: Closing — Lookahead-AdamW (k=6, alpha=0.5) is a clear dead end. Best val 5.75% at ep183 (+1.92pp vs baseline 3.833%), then catastrophic collapse at ep195 (val → 53%, grad_norm → 13.8). Lookahead's slow-weight interpolation amplifies DM's per-car subsampling noise rather than smoothing it. The pre-collapse instability spikes at ep94/122/154/170 are warning signs the optimizer trajectory was already fragile. Dead end — DM optimizer should remain vanilla AdamW with EMA overlay. Thorfinn reassigned..

tcapelle:
STUDENT thorfinn:

## Final Results (training completed)

Training ran 529 epochs before the 360-minute timeout. Final test metrics from best model (epoch 183):

| Metric | Value |
|--------|-------|
| val_primary/surface_rel_l2_pct | **5.75%** |
| test_primary/surface_rel_l2_pct | **6.87%** |
| Best epoch | 183 / 529 |

The model never recovered from the collapses — best val stayed at epoch 183 through all 529 epochs. The test metric (6.87%) is slightly worse than the val metric (5.75%), consistent with the model not generalizing as well as the baseline.

PR #3137, opened 2026-04-23
DrivAerML: 5-epoch warmup + gc=1.0 compound (stabilized warmup)

tcapelle:
## Hypothesis

5-epoch warmup alone (#3069) achieved 5.823% best before diverging at ep523 — the warmup direction was not wrong, just unstable. The warmup hypothesis is: gradual LR ramp allows the optimizer to establish stable update directions before hitting full lr=5e-4. The gap: warmup without gc still hits catastrophic LR-peak explosions.

This compounds warmup + gc=1.0 to provide the missing stability floor. gc=1.0 (hard gradient ceiling) + 5-epoch warmup (gradual start) together should enable stable training past the ep150+ region where warmup-only runs diverged.

## Instructions

**Run: 5-epoch linear warmup + gc=1.0 + T_max=30**
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 --grad-clip 1.0 \
  --warmup-epochs 5 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
  --wandb_group dm-warmup-gc-compound
```

Note: best gc result on DM without warmup was 4.346% (gc=2.0, PR #2886) which still didn't beat baseline 3.997%. The warmup may push gc below baseline if the combination is synergistic. Also run without warmup for comparison:

**Control: gc=1.0 only (no warmup)**
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 --grad-clip 1.0 \
  --no-use-ema --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 --max-eval-batches 200 \
  --wandb_group dm-warmup-gc-compound
```

## Baseline
- **val_primary/surface_rel_l2_pct:** 3.997% at epoch 467 (no gc, no warmup)
[...]

PR #3119, opened 2026-04-22
DrivAerML: higher LR + linear warmup (widening the stable LR basin)

tcapelle:
## Hypothesis

LR=5e-4 was found to be a sharp optimum on DrivAerML — small deviations cause catastrophic divergence. However, this was tested without warmup. Linear warmup gradually builds LR from 0, which can stabilize higher peak LRs that would otherwise diverge from cold start.

A higher peak LR (7e-4 or 1e-3) with 20-epoch linear warmup to the peak, then cosine decay with T_max=30, could reach a deeper loss basin faster. The warmup allows the optimizer to find a stable trajectory before hitting the high LR region.

PRs #3069 (5-epoch warmup) and #3070 (10-epoch warmup) test warmup at the champion lr=5e-4. This PR tests warmup at HIGHER LR to see if it widens the stable operating range.

## Instructions

Implement linear LR warmup in train.py using a LinearLR scheduler chained before the cosine annealing. Run two configurations:

**Run 1: lr=7e-4 with 20-epoch warmup**
- Start LR: 1e-5, ramp linearly to 7e-4 over 20 epochs
- Then: CosineAnnealingLR with T_max=30 and eta_min=1e-6
- Use `torch.optim.lr_scheduler.SequentialLR([LinearLR(..., start_factor=1e-5/7e-4, end_factor=1.0, total_iters=20), CosineAnnealingLR(T_max=30)], milestones=[20])`

```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 7e-4 \
  --cosine-t-max 30 \
  --warmup-epochs 20 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --wandb_group dm-higher-lr-warmup
```

**Run 2: lr=1e-3 with 20-epoch warmup**
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 1e-3 \
  --cosine-t-max 30 \
  --warmup-epochs 20 \
  --no-use-ema \
  --enable-fourier \
[...]

PR #3148, opened 2026-04-23
DrivAerML: 10-epoch linear warmup + gc=1.0

tcapelle:
## Hypothesis

10-epoch linear warmup combined with gradient clipping (gc=1.0) will improve DrivAerML training stability and achieve a lower `val_primary/surface_rel_l2_pct` than the current baseline of 3.997%.

**Rationale:** Warmup alone causes divergence (PRs #3069, #3070). Adding gc=1.0 as a stability guard (chopper, PR #3137, 5-epoch warmup+gc) appears to address this. A 10-epoch warmup gives the optimizer more time to establish stable gradient statistics before the full LR kicks in, potentially reducing early-training instability compared to 5-epoch warmup. The gc=1.0 guard also prevents cosine LR restart shocks at cycle boundaries. This fills the gap between 5-epoch warmup+gc (PR #3137) and 20-epoch warmup alone (PR #3119, still in-progress).

## Instructions

Train on DrivAerML with 10-epoch linear warmup and gradient clipping gc=1.0. Use the champion 4L/512d/8H + Fourier + cosine T_max=30 config as the base.

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 1.0 \
  --warmup-epochs 10 \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --enable-fourier \
  --batch-size 1 \
  --max-train-batches 394 \
  --epochs 999 \
  --wandb-group dm-warmup-gc
```

**Key changes from baseline:**
- Added `--warmup-epochs 10` (baseline: no warmup)
- Added `--grad-clip 1.0` (baseline: no gradient clipping)
- All other settings identical to baseline

**What to report:**
- `val_primary/surface_rel_l2_pct` at best validation epoch
- Whether training was stable (no divergence/spikes)
- Epoch at best checkpoint
- W&B run link

## Baseline

Current best metrics (PR #2898, piccolo):
- W&B run: bht6h42t
- Config: 4L/512d/8H, AdamW lr=5e-4, cosine T_max=30, no-EMA, Fourier features, 360-min budget, SENPAI_MAX_EPOCHS=9999

Reproduce baseline:
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --no-use-ema \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -166,6 +166,7 @@ class TrainConfig:
     volume_anchor_points: int = 8_000
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
+    warmup_epochs: int = 0
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -1625,7 +1626,22 @@ def main() -> None:
     scheduler = None
     if config.cosine_t_max > 0:
         base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
-        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        cosine_scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        if config.warmup_epochs > 0:
+            if config.max_train_batches > 0:
+                steps_per_epoch = max(1, config.max_train_batches // max(1, config.grad_accum_steps))
+            else:
+                steps_per_epoch = len(train_loader)
+            warmup_steps = config.warmup_epochs * steps_per_epoch
+            warmup_scheduler = torch.optim.lr_scheduler.LinearLR(
+                base_optimizer, start_factor=1e-6, end_factor=1.0, total_iters=warmup_steps,
+            )
+            scheduler = torch.optim.lr_scheduler.SequentialLR(
+                base_optimizer, schedulers=[warmup_scheduler, cosine_scheduler], milestones=[warmup_steps],
+            )
+            print(f"Warmup: {warmup_steps} steps ({config.warmup_epochs} epochs × {steps_per_epoch} steps/epoch)", flush=True)
+        else:
+            scheduler = cosine_scheduler
 
     ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
     anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None
@@ -1637,10 +1653,6 @@ def main() -> None:
     best_model_state: dict[str, torch.Tensor] | None = None
     best_anp_state: dict[str, torch.Tensor] | None = None
 
-    if bundle.spec.name == "tandemfoilset":
-        primary_metric_key = "val_primary/surface_pressure_mae"
-    else:
-        primary_metric_key = f"val_primary/{bundle.spec.default_metric}"
     best_val_metric = float("inf")
     best_epoch = 0
     best_model_state: dict[str, torch.Tensor] | None = None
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT franky:

## Results

### DrivAerML: 10-epoch linear warmup + gc=1.0

| Metric | Baseline | This experiment | Delta |
| `val_primary/surface_rel_l2_pct` | **3.997%** (epoch 467) | **11.196%** (epoch 80) | +7.199% (worse) |

**W&B run:** `rp02hbxl` (wandb-applied-ai-team/senpai-v1)
**Best epoch:** 80 (of 87 completed before divergence)

### Command used
```bash
  --cosine-t-max 30 --grad-clip 1.0 --warmup-epochs 10 \
  --enable-fourier --batch-size 1 \
  --max-train-batches 394 --max-eval-batches 200 --no-use-ema \
  --epochs 999 --wandb-group dm-warmup-gc \
  --agent franky --wandb-name "franky/dm-warmup10-gc1"
```

### What happened

**The hypothesis did not work.** Training showed promising early convergence but suffered from escalating catastrophic instability events that ultimately prevented the model from reaching the baseline.

**Convergence trajectory:**
- Warmup phase (epochs 1-10): Smooth, stable LR ramp from ~0 to 5e-4. Val improved steadily from 58.7% → 22.4%. Warmup worked as intended.
- Early post-warmup (epochs 11-50): Best-metric envelope improved well: 20.5% → 18.8% → 17.4% → 16.4% → 14.9% → 12.1% → 11.5%. Oscillations from cosine restarts (T_max=30 steps) were present but recoverable.
- Epoch 50-80: Continued slow improvement to 11.20% best, but with increasingly severe instability spikes.
- Epoch 82+: **Catastrophic divergence.** Grad norms exploded exponentially: 3,712 → 71,040 → 2.1M → 41.5M. Training killed at epoch 87.

**Pattern of escalating instability:**
| Spike | Epoch range | Peak grad norm | Peak val | Recovery |
|-------|-------------|---------------|----------|----------|
| 1 | 21-32 | ~9 | 54% | ~12 epochs |
| 2 | 54-70 | ~126 | 50% | ~16 epochs |
| 3 | 82+ | 41,500,000 | 74% | **No recovery** |

Each instability event was ~10x worse than the previous one, suggesting the cosine warm restarts (T_max=30 steps) combined with gc=1.0 create a feedback loop where the model enters increasingly unstable basins after each restart cycle. The gc=1.0 clips gradients but doesn't prevent the model parameters from drifting into catastrophically unstable regions over many steps.

**Peak memory:** ~6 GB VRAM (single GPU)

### Bug fix note

Fixed a pre-existing `UnboundLocalError` in `main()` where `primary_metric_key` (a top-level function) was shadowed by a local string variable assignment, causing Python to treat it as unassigned-local on the function call at line 1650. Removed the redundant variable and unified on `best_val_primary_metric_name`.

### Suggested follow-ups

1. **Tighter gradient clipping (gc=0.3 or gc=0.5):** gc=1.0 is clearly insufficient to prevent cascade divergence. The baseline's competitor PR #3137 (chopper, 5-epoch warmup + gc) might reveal whether tighter clipping helps.
2. **Lower peak LR with warmup:** Instead of warmup to 5e-4, try warmup to 3e-4 or 2e-4 to reduce the energy of cosine restart shocks.
3. **eta_min > 0 for CosineAnnealingLR:** The current cosine goes to LR=0, which means the model stops learning at troughs but then gets a full-power restart. A non-zero eta_min (e.g., 1e-5) would smooth transitions.
4. **Longer T_max:** T_max=30 steps gives ~13 cosine cycles per epoch, which is extremely aggressive. A longer T_max (e.g., 100 or 200 steps) would reduce the frequency of restart shocks.

tcapelle:
ADVISOR: Closing PR #3148 because ADVISOR: 10-epoch warmup + gc=1.0 reaches only 11.196% (ep80) then diverges catastrophically at ep87 — gc=1.0 is insufficient to prevent cascade divergence with T_max=30 restarts. Each restart cycle ~10x worse than previous, grad norms reaching 41.5M. Warmup delays but doesn't prevent the fundamental instability. 7.4 pp worse than the 3.833% baseline. Closing..

PR #3154, opened 2026-04-23
DrivAerML: EMA=0.9995 + gc=0.5 + monotonic cosine (no restarts)

tcapelle:
## Hypothesis

The current best DrivAerML configuration (PR #3072, 3.833%) uses cosine annealing with restarts every T_max=30 epochs. Each restart creates a loss spike at the restart peak, and EMA with decay=0.9995 takes ~10 epochs to flush each spike — wasting a significant portion of our 999-epoch convergence budget on recovery rather than descent.

A single long monotonic cosine decay (T_max=999, no restarts) should:
1. Eliminate restart-induced spikes entirely, giving EMA a smooth signal throughout training
2. Allow EMA to track a steadily improving model rather than a cyclically perturbed one
3. Potentially push val_primary/surface_rel_l2_pct below 3.82% (closing gap toward the AB-UPT external target of 3.71%)

The hypothesis is that the EMA + cosine restarts combination is actively counterproductive: restarts were designed for SGD-style escape from sharp minima, but with AdamW + EMA the restart spikes are pure noise.

## Instructions

Run the following command. The only change vs the current best (PR #3072) is `--cosine-t-max 999` (was 30) — a single monotonic cosine decay over the full training run, with no restarts:

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 999 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --wandb-group dm-ema-schedule
```

**Key change:** `--cosine-t-max 999` (was `30`) — single long monotonic cosine decay, no restarts.

All other hyperparameters are identical to the current champion (PR #3072): AdamW lr=5e-4, gc=0.5, EMA=0.9995, 4L/512d/8H architecture, 50k surface points, batch-size 1.

**What to report:** Please report val_primary/surface_rel_l2_pct at best validation checkpoint, the epoch it was achieved, and the W&B run ID. Also note whether the training loss curve was visibly smoother than restart-based runs.

## Baseline

Current best metrics (PR #3072 — eren, EMA=0.9995 + gc=0.5):
- val_primary/surface_rel_l2_pct: **3.833%** (epoch 511)
- test_primary/surface_rel_l2_pct: 4.685%
- W&B run: `ncl1dh88` (eren/dm-ema-9995-gc05)
- Immediate sub-target: <3.82%

Reproduce current best:
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -1637,14 +1637,7 @@ def main() -> None:
     best_model_state: dict[str, torch.Tensor] | None = None
     best_anp_state: dict[str, torch.Tensor] | None = None
 
-    if bundle.spec.name == "tandemfoilset":
-        primary_metric_key = "val_primary/surface_pressure_mae"
-    else:
-        primary_metric_key = f"val_primary/{bundle.spec.default_metric}"
     best_val_metric = float("inf")
-    best_epoch = 0
-    best_model_state: dict[str, torch.Tensor] | None = None
-    best_anp_state: dict[str, torch.Tensor] | None = None
     best_ema_shadow: dict[str, torch.Tensor] | None = None
     best_anp_ema_shadow: dict[str, torch.Tensor] | None = None
 
@@ -1713,7 +1706,7 @@ def main() -> None:
             best_model_state = snapshot_module_state(model)
             best_anp_state = snapshot_module_state(anp_head)
 
-        current_val = eval_metrics.get(primary_metric_key)
+        current_val = eval_metrics.get(best_val_primary_metric_name)
         if current_val is not None and current_val < best_val_metric:
             best_val_metric = current_val
             best_epoch = epoch


tcapelle:
STUDENT historia:

## Results

### Metrics

| Metric | Baseline (PR #3072) | This Run | Delta |
|--------|---------------------|----------|-------|
| val_primary/surface_rel_l2_pct | **3.833%** (ep511) | **4.392%** (ep308) | +14.6% (worse) |
| External target (AB-UPT) | <3.71% | 4.392% | +18.4% (worse) |

### Key Details

- **W&B run:** `we38omf3` (historia/dm-ema-monotonic-cosine)
- **Peak memory:** ~6.0 GB
- **Command:**
  ```bash
    --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 999 \
    --ema-decay 0.9995 --wandb-group dm-ema-schedule \
    --agent historia --wandb-name "historia/dm-ema-monotonic-cosine"
  ```

### What happened — HYPOTHESIS FALSIFIED

The hypothesis that monotonic cosine decay would outperform restart-based cosine was **falsified**. The result (4.39% at ep308) is 14.6% worse than the baseline (3.83% at ep511).

#### Critical finding #1: T_max=999 is NOT monotonic

The `--cosine-t-max 999` parameter is interpreted as **999 steps** (not epochs). With 394 steps/epoch, this creates a cosine half-cycle of ~2.5 epochs (full cycle ~5 epochs). So the experiment was actually testing **rapid cycling** (T_max=999 steps ≈ T_max=2.5 epochs) vs the baseline's **moderate cycling** (T_max=30 epochs), NOT monotonic decay vs restarts. For a true monotonic decay, T_max would need to be 999×394 = 393,606.

#### Critical finding #2: EMA+gc=0.5 provides extraordinary stability — until it doesn't

The EMA=0.9995 + gc=0.5 combination delivered **remarkable training stability**:
- 308 epochs of near-monotonic descent (59.0% → 4.39%)
- Best epochs at nearly every cosine trough, with only brief 1-3 epoch upticks at peaks
- Zero gradient explosions through epoch 319
- Clear cyclic pattern: bests cluster at LR minima, brief upticks at LR maxima

However, this stability was **fragile**:
- **Epoch 320:** First instability event (138/394 clips, grad_norm=0.59). EMA degraded to 6.72% over 3 epochs, then recovered in ~12 epochs
- **Epoch 336:** Catastrophic cascade began (332/394 clips, grad_norm=27.3). Val spiked from 4.64% to 7.65%
- **Epochs 337-339:** Full divergence — ALL steps clipped, grad_norm 16-42, val ballooned to 22.9%

#### Critical finding #3: High EMA decay is a double-edged sword

With decay=0.9995, each EMA update mixes only 0.05% of the online model per step. This means:
- **Pro:** Extremely smooth convergence curve, filters out training noise
- **Con:** When the online model is corrupted by a gradient explosion, the corrupted weights persist in the EMA for ~15-20 epochs. The EMA that provides stability is the same mechanism that makes recovery slow.

After the epoch 320 event, it took 14 epochs to recover from 6.72% back to ~4.64%. The epoch 336 cascade was terminal — the online model entered a feedback loop of producing bad gradients, which poisoned the EMA, which couldn't provide good targets.

#### Convergence phases

| Phase | Epochs | Val Range | Improvement Rate | Notes |
|-------|--------|-----------|-----------------|-------|
| Rapid descent | 1-100 | 59% → 7.3% | ~0.5%/epoch | Near-monotonic, every epoch a best |
| Steady descent | 100-200 | 7.3% → 5.0% | ~0.02-0.05%/epoch | Plateaus appear (6-14 epochs), trough-driven |
| Slow descent | 200-308 | 5.0% → 4.39% | ~0.005-0.01%/epoch | Long plateaus (9-25 epochs), diminishing returns |
| First instability | 320-335 | 4.64% → 6.72% → 4.64% | Recovery | EMA absorbs corruption, slow recovery |
| Terminal collapse | 336-339+ | 4.64% → 22.9%+ | Diverging | Cascading gradient explosions |

### Bug fix included

Fixed a `primary_metric_key` local variable shadowing bug in train.py (lines 1640-1649). A local variable named `primary_metric_key` was being assigned within the same function scope as the `primary_metric_key()` function definition (line 1423), causing Python to shadow the function. This produced `UnboundLocalError` when the function was called before the local assignment. Fix: removed the duplicate local variable and used `best_val_primary_metric_name` (already correctly set by the function call at line 1634).

### Suggested follow-ups

1. **True monotonic cosine test:** Use `--cosine-t-max 393606` (999 epochs × 394 steps/epoch) to actually test the no-restarts hypothesis. The current experiment tested rapid cycling, not monotonic decay.

2. **EMA checkpoint recovery:** Implement a mechanism to save the best EMA weights and reload them after gradient explosions, preventing the cascade failure pattern.

3. **Adaptive gradient clipping:** The gc=0.5 threshold was insufficient to prevent the epoch 336 explosion (grad_norm=27.3, 55× above threshold). A more aggressive per-parameter adaptive clipping could catch outlier batches.

4. **T_max comparison at steps level:** The baseline's T_max=30 (epochs) = ~11,820 steps. This experiment's T_max=999 (steps) is 12× shorter cycles. A systematic sweep of T_max in steps (999, 5000, 11820, 50000, 393606) would map the cycle-length landscape.

5. **Lower EMA decay for DrivAerML:** decay=0.999 (vs 0.9995) would recover 10× faster from instability events — this tradeoff may be worthwhile given DrivAerML's gradient explosion tendency.

tcapelle:
ADVISOR: Closing PR #3154 because ADVISOR: 4.392% ep308, 14.6% worse than 3.833% baseline (W&B: we38omf3). HOWEVER the original hypothesis was NOT tested — T_max=999 is in STEPS not epochs, so this tested rapid 5-epoch cycling instead of true monotonic decay. True monotonic would need T_max=393606 (999ep × 394 steps/ep). The rapid cycling result confirms short cycles are worse, but monotonic cosine remains an open question. Will reassign the correct test. EMA cascade analysis (corrupted online model needs 14+ ep to flush from EMA buffer) is a valuable finding..

PR #3155, opened 2026-04-23
DrivAerML: EMA=0.9995 + gc=0.5 + longer T_max (T_max=45, T_max=60)

tcapelle:
## Hypothesis

T_max=30 was previously the only stable cosine period — restarts at shorter periods caused catastrophic divergence on DrivAerML. PR #3072 proved that gradient clipping (gc=0.5) prevents restart divergence entirely. With gc=0.5 as a stability guard, T_max=45 (50% longer) and T_max=60 (2× longer) should now be viable. Longer cosine periods give EMA more uninterrupted convergence time per cycle before the next restart perturbation, potentially allowing EMA to track a deeper valley before being perturbed. This is a direct, low-risk extension of the champion config — one parameter change, two variants.

## Instructions

Run **two variants** back-to-back. Both use `--wandb-group dm-ema-tmax-sweep` so they appear together in W&B.

**All settings are identical to the PR #3072 champion** (EMA=0.9995, gc=0.5, 4L/512d/8H, AdamW lr=5e-4) except T_max.

**Variant 1 — T_max=45:**
```bash
cd target/icml2026
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 45 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --wandb-group dm-ema-tmax-sweep \
  --run-name dm-ema-tmax45
```

**Variant 2 — T_max=60:**
```bash
cd target/icml2026
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 60 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --wandb-group dm-ema-tmax-sweep \
  --run-name dm-ema-tmax60
```

**Please report for each variant:**
- Best `val_primary/surface_rel_l2_pct` and the epoch it occurred
- Whether training was stable throughout (no divergence at restarts)
- W&B run ID
- Convergence speed impression vs T_max=30 champion (did the minimum come earlier or later in the budget?)

**Target:** sub-3.820% on `val_primary/surface_rel_l2_pct`.

## Baseline

Current best: **PR #3072** (eren — EMA=0.9995 + gc=0.5, 4L/512d/8H, AdamW lr=5e-4, T_max=30)

| Metric | Value |
|--------|-------|
| epoch | 511 |
| W&B run | `ncl1dh88` (eren/dm-ema-9995-gc05) |

External target (AB-UPT): <3.71% — gap is 0.123 pp.

Reproduce champion:
```bash
cd target/icml2026
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -1637,14 +1637,7 @@ def main() -> None:
     best_model_state: dict[str, torch.Tensor] | None = None
     best_anp_state: dict[str, torch.Tensor] | None = None
 
-    if bundle.spec.name == "tandemfoilset":
-        primary_metric_key = "val_primary/surface_pressure_mae"
-    else:
-        primary_metric_key = f"val_primary/{bundle.spec.default_metric}"
     best_val_metric = float("inf")
-    best_epoch = 0
-    best_model_state: dict[str, torch.Tensor] | None = None
-    best_anp_state: dict[str, torch.Tensor] | None = None
     best_ema_shadow: dict[str, torch.Tensor] | None = None
     best_anp_ema_shadow: dict[str, torch.Tensor] | None = None
 
@@ -1713,7 +1706,7 @@ def main() -> None:
             best_model_state = snapshot_module_state(model)
             best_anp_state = snapshot_module_state(anp_head)
 
-        current_val = eval_metrics.get(primary_metric_key)
+        current_val = eval_metrics.get(best_val_primary_metric_name)
         if current_val is not None and current_val < best_val_metric:
             best_val_metric = current_val
             best_epoch = epoch


tcapelle:
STUDENT nobara:

## Results

### Bug fix
Found and fixed a bug in `train.py` where a local variable `primary_metric_key` (string) shadowed the function `primary_metric_key()` defined at line 1423, causing `UnboundLocalError` on startup. Removed the redundant local variable assignment and replaced its usage with the already-computed `best_val_primary_metric_name`. This bug affected all datasets.

### Variant 1 — T_max=45

| Metric | Value |
|--------|-------|
| val_primary/surface_rel_l2_pct | **4.638%** |
| best epoch | 240 |
| test surface_rel_l2_pct | 5.622% |
| total epochs trained | 506 |
| W&B run | `tskmw04v` |

### Variant 2 — T_max=60

| Metric | Value |
|--------|-------|
| val_primary/surface_rel_l2_pct | **4.409%** |
| best epoch | 331 |
| test surface_rel_l2_pct | 5.414% |
| total epochs trained | 501 |
| W&B run | `fmxxu3ol` |

### Baseline comparison (T_max=30, PR #3072)

| Metric | Baseline | T_max=45 | T_max=60 |
|--------|----------|----------|----------|
| val surface_rel_l2_pct | **3.833%** | 4.638% (+0.805pp) | 4.409% (+0.576pp) |
| test surface_rel_l2_pct | **4.685%** | 5.622% (+0.937pp) | 5.414% (+0.729pp) |
| best epoch | 511 | 240 | 331 |

External target (AB-UPT): <3.71%

### Stability analysis

Training was stable in the sense that gc=0.5 prevented catastrophic divergence at restart boundaries — no NaN/Inf. However, both variants show severe degradation in later epochs: by the final epoch, val metrics had blown up to >50%, far above the best-epoch values. This suggests the model found a good minimum in the first few cosine cycles but then destabilized as training progressed, possibly because the longer T_max allows the learning rate to stay high for longer periods, overshooting the minimum.

### What happened

**Neither variant beat the baseline. Both were significantly worse.** The hypothesis that longer cosine periods would give EMA more uninterrupted convergence time did not pan out — instead, the longer high-LR phases appear to destabilize training. T_max=30 remains the sweet spot:
- Short enough that the LR drops quickly and frequently, keeping the model near good minima
- Long enough for meaningful per-cycle convergence
- The champion's 3.833% at epoch 511 was achieved with many more restart cycles (17 full cycles vs 11 for T_max=45 vs 8 for T_max=60), suggesting frequent restarts are actually beneficial for exploration

### Suggested follow-ups

- **Shorter T_max sweep (T_max=15, T_max=20)**: If T_max=30 is better than 45/60, perhaps even shorter periods with more restarts could help — though the advisor previously found shorter periods caused divergence before gc was added
- **Warm restarts with decay**: Instead of fixed T_max, progressively lengthening T_max (e.g. T_max doubles each cycle) so early cycles explore quickly and later cycles converge more slowly
- **eta_min tuning**: A non-zero eta_min could prevent the LR from going too low at cycle bottoms, potentially stabilizing later epochs

tcapelle:
ADVISOR: Closing PR #3155 because ADVISOR: T_max=45 → 4.638% (diverged after ep240), T_max=60 → 4.409% (diverged after ep331). Both worse than 3.833%. Longer cycles keep the model in high-LR territory too long, causing progressive destabilization. T_max=30 gives ~17 restart cycles in the budget, which is the right frequency. Combined with T_max=15 (4.406%, plateaued) and T_max=20 (4.943%, diverged), T_max=50 (diverges), T_max=30 is now confirmed optimal across ALL tested values. Adding T_max=45/60 to dead ends..

PR #3164, opened 2026-04-23
DrivAerML: EMA=0.9995 + gc=0.5 paper-facing full eval (two-phase)

tcapelle:
## Hypothesis

This is a paper-facing evaluation run, not a new hypothesis. We need the TRUE `test_primary/surface_rel_l2_pct` for the ICML paper from the champion config (EMA=0.9995+gc=0.5). The current test=4.685% from PR #3072 used `--max-eval-batches 200` (truncated eval). Full eval requires ~7 min/epoch, making it too expensive to run every epoch.

**Two-phase approach:**
1. Train to convergence with `--max-eval-batches 200` (same as champion)
2. Load best checkpoint and run one final full eval without `--max-eval-batches`

## Baseline

- **val_primary/surface_rel_l2_pct:** 3.833% (partial eval, PR #3072)
- **test_primary/surface_rel_l2_pct:** 4.685% (partial eval, PR #3072) — TRUNCATED, NEED TRUE NUMBER
- **Best PR:** #3072 (eren — EMA=0.9995 + gc=0.5, 4L/512d/8H, AdamW lr=5e-4, T_max=30)
- **External target:** <3.71% (AB-UPT, ~500 epochs)
- **Baseline W&B run:** ncl1dh88

### Phase 1: Train to convergence (same as champion config)

Run the champion config to convergence. This replicates PR #3072 and should achieve ~3.833% val or better:

```bash
  --ema-decay 0.9995 --wandb-group dm-paper-eval
```

The best-checkpoint saving (PR #2974) saves the best val checkpoint automatically. Note the path to the best val checkpoint from the training output.

### Phase 2: Full eval of the best checkpoint

After Phase 1 completes, load the best val checkpoint and run a full evaluation WITHOUT `--max-eval-batches`. This gives the paper-facing test metric:

```bash
  --max-train-batches 394 \
  --ema-decay 0.9995 --wandb-group dm-paper-eval \
  --resume <path-to-best-val-checkpoint> --eval-only
```

If `--eval-only` is not a supported flag, run for 0 or 1 additional epochs from the best checkpoint with no `--max-eval-batches` and report the test metric from that run.

**Important:** Do NOT include `--max-eval-batches` in Phase 2. We need the full test set evaluation.

## What to report

Please report in a PR comment:
1. **Best val_primary/surface_rel_l2_pct** from Phase 1 (should match ~3.833%)
2. **test_primary/surface_rel_l2_pct from FULL eval** (no `--max-eval-batches`) — **THIS IS THE PAPER NUMBER**
3. Whether the full-eval test is better or worse than the truncated 4.685%
4. W&B run IDs for both phases

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,8 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     save_checkpoint: bool = False
+    resume_checkpoint: str = ""
+    eval_only: bool = False
     seed: int = 0
 
 
@@ -1610,6 +1612,10 @@ def main() -> None:
 
     train_loader, val_loaders, test_loaders = build_loaders(config, bundle, num_workers=resolved_num_workers)
     model = build_model(config, bundle).to(device)
+    if config.resume_checkpoint:
+        state_dict = torch.load(config.resume_checkpoint, map_location=device, weights_only=True)
+        model.load_state_dict(state_dict)
+        print(json.dumps({"resumed_checkpoint": config.resume_checkpoint}), flush=True)
     forward_model = torch.compile(model) if config.compile_model and device.type == "cuda" else model
     anp_head = None
     if bundle.spec.name == "tandemfoilset" and config.anp_srf:
@@ -1618,17 +1624,22 @@ def main() -> None:
             output_dim=bundle.spec.surface_output_dim,
         ).to(device)
 
-    params = list(model.parameters())
-    if anp_head is not None:
-        params.extend(list(anp_head.parameters()))
-    optimizer = build_optimizer(params, config)
-    scheduler = None
-    if config.cosine_t_max > 0:
-        base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
-        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
-
-    ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
-    anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None
+    if config.eval_only:
+        ema = None
+        anp_ema = None
+        optimizer = None
+        scheduler = None
+    else:
+        params = list(model.parameters())
+        if anp_head is not None:
+            params.extend(list(anp_head.parameters()))
+        optimizer = build_optimizer(params, config)
+        scheduler = None
+        if config.cosine_t_max > 0:
+            base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
+            scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
+        anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None
     history: list[dict[str, float]] = []
     best_epoch: int | None = None
     best_val_primary_metric_name = primary_metric_key(bundle, phase="val")
@@ -1637,10 +1648,6 @@ def main() -> None:
     best_model_state: dict[str, torch.Tensor] | None = None
     best_anp_state: dict[str, torch.Tensor] | None = None
 
-    if bundle.spec.name == "tandemfoilset":
-        primary_metric_key = "val_primary/surface_pressure_mae"
-    else:
-        primary_metric_key = f"val_primary/{bundle.spec.default_metric}"
     best_val_metric = float("inf")
     best_epoch = 0
     best_model_state: dict[str, torch.Tensor] | None = None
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT faye:

## Results

### Finding: `torch.compile` causes training instability for DrivAerML

Runs 1 and 2 (with `compile_model=True`, the current default) **diverged** at epoch 60–87 with gradient norm explosion (0.6 → 65,000+). Run 3 with `--no-compile-model` trained stably to epoch 521. This is a reproducible bug — two independent compiled runs both diverged, while the non-compiled run converged normally. The champion run (ncl1dh88) was likely trained on an older codebase before `compile_model` became the default.

### Phase 1: Training (Run 3 — no-compile, seed=42)

| Metric | This Run | Champion (ncl1dh88) | Delta |
|--------|----------|---------------------|-------|
| **Best val surface_rel_l2_pct** | **4.198%** (ep432) | 3.833% (ep511) | +0.365 pp |
| Test at best val (truncated 200 batches) | 4.793% | 4.685% | +0.108 pp |
| Epochs trained | 521 (timeout) | 517 | — |

- W&B run: `qd7g8npf`
- Command:
```bash
  --ema-decay 0.9995 --wandb-group dm-paper-eval \
  --no-compile-model --seed 42 \
  --save-checkpoint \
  --agent faye --wandb-name "faye/dm-ema-paper-fulleval-phase1-run3-nocompile"
```

### Phase 2: Full Evaluation (no --max-eval-batches)

| Metric | Full Eval | Truncated (200 batches) | Champion Truncated |
|--------|-----------|------------------------|-------------------|
| **test_surface_rel_l2_pct** | **4.496%** | 4.793% | 4.685% |

- W&B run: `37gsknn6`
- The full eval test (4.496%) is **0.189 pp better** than the champion's truncated test (4.685%)
- External references: AB-UPT 3.82% | Transolver-3: 3.71%

### Diverged runs (for the record)

| Run | W&B | Compile | Best Val | Diverged At |
|-----|-----|---------|----------|-------------|
| Run 1 | `8rd0f27u` | Yes | 8.164% (ep86) | ep87 |
| Run 2 | `2icvo0n8` | Yes | 10.119% (ep62) | ep75 |
| Run 3 | `qd7g8npf` | **No** | **4.198%** (ep432) | Did not diverge |

### What happened

- The champion config (4L/512d/8H, AdamW lr=5e-4, cosine T_max=30, gc=0.5, EMA=0.9995) works well but has a `torch.compile` incompatibility on the current codebase.
- With `--no-compile-model`, run 3 converged to val=4.198% vs champion's 3.833% (+0.365 pp gap, likely due to seed/EMA warmup differences from the new `EMAWithWarmup` class vs old `EMA`).
- The full eval test metric (4.496%) is more favorable than the truncated metric (4.793%), suggesting the 200-batch subset is slightly pessimistic.

### Suggested follow-ups

1. **Bug fix: `torch.compile` + DrivAerML instability** — This should be investigated. The EMA warmup commit (e9720fe) already noted "DrivAerML negative (T_max=30 oscillations too wide for EMA)". The compile interaction makes it worse.
2. **Re-run with old EMA class** — The champion used the old fixed-decay `EMA` (no warmup). Testing with the old class might close the 0.365 pp val gap.
3. **Full eval of original champion checkpoint** — If the ncl1dh88 checkpoint is available, a direct full eval would give the true paper-facing number without needing to retrain.

tcapelle:
ADVISOR: Closing PR #3164 because ADVISOR: Thank you faye — this PR delivered two critical findings:

**Finding 1: torch.compile is broken for DrivAerML (confirmed bug)**
Two independent runs (Run 1, Run 2) both diverged at epochs 60–87 with gradient norm explosion (0.6 → 65,000+). The non-compiled run (Run 3) trained stably. This is a systematic regression, not a fluke. The suspect commit is the EMAWithWarmup merge (e9720fe). All future DrivAerML experiments must use `--no-compile-model` until this is diagnosed and fixed.

**Finding 2: The full-eval test metric is 4.496% vs truncated 4.685% — but this is NOT a new champion record**

The champion record stands at **3.833% val** (PR #3072, eren). Faye's Run 3 achieved **4.198% val** — a gap of +0.365 pp. This means faye's model did not converge as deeply as eren's, likely due to the timeout at epoch 521 vs. no timeout for the champion config.

The 4.496% full-eval number is valuable information (the truncated 4.685% overstates the champion's test error), but we cannot use faye's test metric as the paper-facing number — it is for a weaker model (4.198% val vs 3.833% val). We need a full-eval of the actual champion-level checkpoint.

**Analysis of val gap:**
The champion (ncl1dh88) was also limited by timeout (517 epochs), but still converged to 3.833%. Faye's run hit 521 epochs but only reached 4.198%. The 0.365 pp gap may be due to: (a) different random seed, (b) the compile/no-compile change affecting learning dynamics, or (c) compile was enabled in the original ncl1dh88 run on an older codebase version before the EMAWithWarmup commit. The most likely explanation is seed variance plus the fact that eren's run may have inadvertently benefited from a pre-regression codebase.

**What we actually need for the paper:**
A fresh rerun of the champion config using `--no-compile-model` for a full training run to convergence, then a full eval (no --max-eval-batches) of the best checkpoint. This will give both the true val and true test paper metrics.

**Closing this PR** — the immediate next step is a dedicated full-champion-rerun PR with `--no-compile-model` (new assignment)..

PR #3182, opened 2026-04-23
DrivAerML: torch.compile + 96 attention slices on EMA champion

tcapelle:
## Hypothesis

The current DM champion (PR #3072, 3.833% val) trains with default attention slice count and no torch.compile. We hypothesize that:

1. **96 attention slices** (vs default) increases model expressiveness and may directly improve metric quality.
2. **torch.compile** reduces per-epoch wall-clock time on DrivAerML, allowing more epochs within the fixed time budget — translating throughput gains into lower final val error.

The prior compile experiment (PR #2898) diverged to NaN at ep454 because it ran without grad-clip. The champion config already uses `--grad-clip 0.5`, which should prevent that failure mode.

This two-trial design cleanly isolates the two effects:
- **Trial 1:** compile + 96 slices (joint effect)
- **Trial 2:** no compile + 96 slices (slice effect only)

External target: <3.71% (AB-UPT, ~500 epochs). Gap to close: 0.123 pp.

## Instructions

**SAFETY NOTE:** If either run diverges (loss → NaN or val metric > 6%) before ep100, abort that run immediately and report the last stable checkpoint metrics.

**Trial 1 — torch.compile + 96 slices:**

```bash
  --dataset drivaerml \
  --enable-fourier \
  --model-attn-slice-size 96 \
  --compile-model \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --wandb_group dm-compile-slices \
  --wandb_run_name dm-compile-96slices-trial1
```

**Trial 2 — 96 slices only (no compile), isolates slice effect:**

```bash
  --dataset drivaerml \
  --enable-fourier \
  --model-attn-slice-size 96 \
  --no-compile-model \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --wandb_group dm-compile-slices \
  --wandb_run_name dm-nocompile-96slices-trial2
```

Run both trials in parallel if GPU resources allow.

**Report back:**
- `val_primary/surface_rel_l2_pct` at best val checkpoint for each trial
- Time per epoch (seconds) for each trial vs baseline
- Final epoch reached within time budget for each trial
- W&B run links for both trials

## Baseline

Current DM champion (PR #3072 — eren, EMA=0.9995 + gc=0.5):

| Metric | Value |
|--------|-------|
| Best epoch | 511 |
| W&B run | ncl1dh88 |

**Reproduce command (baseline):**
```bash
  --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 \
  --max-eval-batches 200 --ema-decay 0.9995
```

**External target:** <3.71% (AB-UPT, ~500 epochs)


tcapelle:
STUDENT norman:

## Clarification needed — default values

While reviewing the PR instructions, I found that the current `train.py` defaults already match what the hypothesis proposes to test:

1. **`model_slices: int = 96`** (line 107) — the DrivAerML champion (PR #3072) doesn't pass `--model-slices`, so it already uses 96 slices by default.
2. **`compile_model: bool = True`** (line 132) — the champion doesn't pass `--no-compile-model`, so it already compiles by default.
3. **`--model-attn-slice-size`** is not a valid CLI flag. The auto-generated flag is `--model-slices`.

This means:
- **Trial 1** (compile + 96 slices) = the existing champion config (no change)
- **Trial 2** (no-compile + 96 slices) = champion minus compile (the only meaningful comparison)

The prior experiment PR #2898 already tested compile vs no-compile (without grad-clip) and found no-compile won. The champion (PR #3072) uses the compile default (True) plus grad-clip 0.5, so it already trains with compile + grad-clip.

**Options:**
1. Run Trial 2 only (no-compile + grad-clip 0.5) to see if disabling compile improves the champion further
2. Test different slice counts (e.g., 32, 64, 128, 192) against the default 96
3. Something else you had in mind?

Please advise!

tcapelle:
ADVISOR: Closing PR #3182 because ADVISOR: Good catch — you correctly identified that model_slices=96 and compile_model=True are already the defaults. The champion already uses both. Closing this and reassigning you to a genuinely new experiment..

PR #3202, opened 2026-04-23
DrivAerML: GLU preprocess MLP — gated input feature selection

tcapelle:
## Hypothesis

The standard input projection in the surface MLP is `x = Linear(in_features, hidden_dim)`. Replacing this with a Gated Linear Unit (GLU) preprocess — `value = Linear(in_features, hidden_dim); gate = Linear(in_features, hidden_dim); x = value * sigmoid(gate)` — allows the model to selectively amplify or suppress input features before they propagate through the network.

This is the \"second innovation track\" for DrivAerML: a genuinely new physics-aware architectural idea rather than local hyperparameter retuning. The gate mechanism may be particularly valuable because DrivAerML inputs include Fourier positional encodings and geometric features (normals, curvature) with very different scales and information content. The gate learns which features are relevant at each point, acting as a soft, learned feature selection layer.

The noam branch reportedly has an existing GLU preprocess implementation — port that first rather than writing from scratch.

**Why this may beat baseline:** The current DM champion (3.833%) already uses 4L/512d with EMA + gc stability. The remaining gap to AB-UPT (<3.71%) is 0.123 pp. Architectural improvements to input representation are one of the few levers not yet pulled on this configuration.

## Instructions

**Step 1: Check noam branch for existing implementation**

```bash
git fetch origin noam
git diff radford...origin/noam -- target/icml2026/train.py | grep -A 20 -B 5 -i "glu\|gate\|preprocess"
```

If a GLU preprocess implementation exists on noam, port it. Otherwise implement from scratch as described below.

**Step 2: Implement GLU preprocess in `target/icml2026/train.py`**

Find the surface MLP input projection (likely something like `self.input_proj = nn.Linear(in_features, hidden_dim)`). Replace it with a GLU preprocess:

```python
# Replace:
#   self.input_proj = nn.Linear(in_features, hidden_dim)
# With:
self.input_proj_value = nn.Linear(in_features, hidden_dim)
self.input_proj_gate  = nn.Linear(in_features, hidden_dim)

# Replace:
#   x = self.input_proj(x)
# With:
x = self.input_proj_value(x) * torch.sigmoid(self.input_proj_gate(x))
```

**Step 3: Add CLI flag**

Add `--glu-preprocess` boolean flag (default `False`) and thread it through to the model constructor. The standard MLP path must remain unchanged when the flag is absent.

**Step 4: 3-epoch smoke test before full run**

```bash
cd target/icml2026 && python train.py --dataset drivaerml --optimizer adamw \
  --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 3 \
  --max-eval-batches 200 --ema-decay 0.9995 --glu-preprocess
```

Verify it runs without error and val metrics are finite before launching the full run.

**Step 5: Full experiment**

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --glu-preprocess \
  --wandb_group dm-innovation-track
```

**Critical config notes:**
- Do NOT add weight decay — WD hurts DrivAerML
- EMA=0.9995 requires gc=0.5 as a stability guard — EMA alone diverges at ep28 without it
- Use `--cosine-t-max 30` with per-batch stepping (T_max in batches, not epochs)
- Use `--wandb_group dm-innovation-track` to group with future innovation-track runs

**Report:** `val_primary/surface_rel_l2_pct` at the best validation checkpoint, plus W&B run ID.

## Baseline

Current DrivAerML best (radford branch):

| Metric | Value |
|--------|-------|
| `val_primary/surface_rel_l2_pct` | **3.833%** |
| Best epoch | 511 |
| Test surface_rel_l2_pct | 4.685% |

- W&B run: `ncl1dh88` (eren/dm-ema-9995-gc05, PR #3072)
- External target: AB-UPT < 3.71% (~500 epochs); gap = 0.123 pp

Reproduce command:
```bash
```

### DIFF
--- a/research/CURRENT_RESEARCH_STATE.md
+++ b/research/CURRENT_RESEARCH_STATE.md
@@ -1,11 +1,11 @@
 # SENPAI Research State
 
-- **Date:** 2026-04-23 20:30 (advisor cycle 43 — closed #3134, assigning megumi)
+- **Date:** 2026-04-23 21:00 (advisor cycle 44 — closed #3190/#3167/#3166/#3165, assigning 5 students)
 - **Branch:** radford
-- **Idle students:** 0 (megumi being assigned)
+- **Idle students:** 0 (brook/gilbert/senku/vegeta/jet being assigned)
 - **PRs ready for review:** 0
 
-## Fleet Status (56 active PRs → 59 after assignments)
+## Fleet Status (55 active PRs + 5 pending)
 
 ### DrivAerML WIP (30 PRs)
 **Noam-pivot experiments (highest priority):**
@@ -14,10 +14,14 @@
 - `#3181` himmel: asinh+residual on champion — NOAM ABLATION
 - `#3175` hinata: full noam stack on champion — NOAM ABLATION
 
+**Innovation track (code ports from noam — new directive):**
+- `#NEW` senku: DM GLU preprocess MLP — INNOVATION TRACK
+- `#NEW` vegeta: DM attention temperature annealing — INNOVATION TRACK
+- `#NEW` jet: DM DomainLayerNorm — INNOVATION TRACK
+
 **Pre-pivot champion tuning:**
 - `#3173` kakashi: shorter cosine T_max=15/20
 - `#3171` sanji: LR warmup (5-ep/10-ep) on champion
-- `#3165` senku: 4H heads (128d/head)
 - `#3164` faye: paper-facing full eval (two-phase) — PAPER-FACING
 - `#3163` shouko: attn dropout=0.05
 - `#3161` spike: eta_min=1e-5 (cosine floor)
--- a/research/EXPERIMENTS_LOG.md
+++ b/research/EXPERIMENTS_LOG.md
@@ -1,5 +1,21 @@
 # SENPAI Research Results
 
+## 2026-04-23 21:00 — PR #3190: TFP physics features isolation (brook) — CLOSED
+
+- Trial 1 (all physics): val=Infinity — cp_panel_prior_index() returns wrong index for tandemfoil_paper, applies sinh(Fourier_value) → overflow. Ux/Uy learn normally but pressure permanently broken (W&B: `0eq5jqij`). Trial 2 (wake only): val=0.7396 ep98, diverged via unbounded dx_norm/dy_norm (W&B: `kqknlctz`). CRITICAL FINDING: build_tandemfoil_paper_bundle only supports enable_fourier/wake_deficit/wake_angle — TE coord, cp_panel, vortex flags silently ignored. primary_metric_key bug fix noted.
+
+## 2026-04-23 21:00 — PR #3167: AF higher LR 8e-4/1e-3 (gilbert) — CLOSED
+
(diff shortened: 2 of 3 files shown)


tcapelle:
STUDENT senku:

## Results

### GLU Preprocess — DrivAerML

| Metric | Baseline | GLU Preprocess | Delta |
| `val_primary/surface_rel_l2_pct` | **3.833%** (ep 511) | **4.411%** (ep 313) | +0.578 pp ❌ |

**Did not beat baseline.** The GLU preprocess model reached a best val of 4.411% at epoch 313, then catastrophically diverged starting around epoch 330. By epoch 350, grad norms had exploded to 274–636 and val metric regressed to 27%.

### Details

- **W&B run:** `0wbhzbbd` (senku/dm-glu-preprocess, group: dm-innovation-track)
- **Peak GPU memory:** ~6.2 GB (H100)
- **Command:**
  ```bash
    --glu-preprocess --wandb-group dm-innovation-track \
    --agent senku --wandb-name "senku/dm-glu-preprocess"
  ```

### What happened

The GLU preprocess converged smoothly through epoch ~313 (4.411%), but the convergence curve was always behind the baseline trajectory. At epoch ~330, training became unstable — grad norms spiked from <1 to 50+, then escalated to 100–600+ by epoch 350, with val metric regressing to 27%.

The divergence likely stems from the GLU gate's multiplicative interaction with the Fourier positional encodings and geometric features. As training progresses and the gate weights sharpen, the sigmoid saturation gradients can create sudden instability, especially at high-LR phases of the cosine restarts. The grad clipping (0.5) was insufficient to contain this.

Key observation: the baseline (without GLU) ran stably to 511 epochs. The GLU preprocess introduced a latent instability that manifested ~60% through training.

### Bug fix note

Fixed a pre-existing bug: the variable `primary_metric_key` on line 1665 shadowed the function `primary_metric_key()` defined on line 1447, causing an `UnboundLocalError`. Renamed the variable to `val_metric_key`.

### Suggested follow-ups

1. **GLU with gate bias initialization:** Initialize the gate bias to a positive value (e.g., +1) so the gate starts near-open. This avoids the sharpening instability that caused divergence.
2. **SiLU/Swish instead of sigmoid gating:** Replace `sigmoid(gate)` with `SiLU(gate)` which has better gradient flow properties and is less prone to saturation.
3. **GLU on extra features only (not pos embed):** Apply the GLU only to the `project_surface_features` output, not the combined hidden. This limits the multiplicative interaction to just the geometric features.
4. **Lower LR for GLU parameters:** Use a separate parameter group with 0.1x LR for the gate weights to prevent the gate from over-sharpening.

tcapelle:
ADVISOR: Closing PR #3202 because ADVISOR: Closing — GLU preprocess does not improve DM and causes late-training divergence. Best 4.411% at ep313 (+0.578pp vs baseline 3.833%), then catastrophic gradient explosion at ep330 (grad norms → 636). The GLU gate's sigmoid saturation creates instability at cosine restart LR peaks — same failure pattern as LLRD, SWP, and error correction experiments. The convergence curve was always behind baseline trajectory even before divergence. Thanks for the bug fix (primary_metric_key shadowing) — noted for potential cherry-pick. Senku reassigned..

PR #3207, opened 2026-04-23
DrivAerML: true monotonic cosine decay (T_max=393606 steps)

tcapelle:
## Hypothesis

Previous experiment #3154 (historia) intended to test true monotonic cosine decay over the full 999-epoch training run on the DM champion config, but accidentally used `T_max=999` which — since the cosine scheduler operates in **steps**, not epochs — resulted in rapid 999-step (~2.5 epoch) mini-cycles rather than a single monotonic decay. This experiment corrects that mistake.

The correct value for a single half-cosine decay from lr=5e-4 to 0 over the full 999-epoch run is:

```
T_max = 999 epochs × 394 steps/epoch = 393,606 steps
```

**Why this may help:** The current DM champion (PR #3072, T_max=30 = ~13 steps/epoch) uses rapid cosine restarts, which introduce periodic gradient spikes at each restart boundary. With EMA=0.9995 (very slow decay — remembers ~2000 steps of history), these restarts create regular disruptions to the EMA weights. A true monotonic decay eliminates restart boundaries entirely, allowing smoother long-horizon convergence and potentially letting EMA accumulate cleaner model weight averages over the full training.

Reference: cosine annealing without restarts (Loshchilov & Hutter, 2016, SGDR) — the "WR" (warm restarts) variant dominates practice, but the monotonic form can outperform when the training budget is fixed and long.

## Instructions

**3-epoch smoke test first.** Confirm no divergence and surface_rel_l2_pct is reasonable (< 10%). Then run the full trial.

**Trial 1 — True monotonic cosine (T_max=393606 steps = full 999 epochs):**

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 393606 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --wandb_group dm-monotonic-cosine
```

**Key parameter notes:**
- `--cosine-t-max 393606` = 999 epochs × 394 steps/epoch = single half-cosine decay from lr=5e-4 to 0 over full training (no restarts)
- `--ema-decay 0.9995` = DM champion EMA (requires gc=0.5 for stability — already included)
- `SENPAI_MAX_EPOCHS=9999` = allow full 999-epoch run
- All other flags are identical to the DM champion config (#3072)
- Do NOT change `--grad-clip 0.5` — this is the stability guard for EMA on DrivAerML

**What to report:**
- Best `val_primary/surface_rel_l2_pct` and at which epoch
- Final epoch `val_primary/surface_rel_l2_pct`
- W&B run ID
- Whether the LR schedule looks correct in W&B (should be a smooth monotonic decay from 5e-4 to ~0 at epoch 999, no restarts/spikes)

## Baseline

- **Dataset:** DrivAerML
- **Current best:** 3.833% at epoch 511 (PR #3072)
- **Config:** AdamW lr=5e-4, T_max=30 (rapid cycling ~13 steps/epoch), gc=0.5, EMA=0.9995, 4L/512d/8H, Fourier, no WD
- **External target:** <3.71% (AB-UPT) — gap: 0.123 pp
- **Reproduce baseline:**
  ```bash
  cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995
  ```

**Target:** Beat 3.833%. The hypothesis is that removing cosine restart boundaries allows smoother long-horizon convergence with EMA=0.9995.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -1642,14 +1642,7 @@ def main() -> None:
     best_model_state: dict[str, torch.Tensor] | None = None
     best_anp_state: dict[str, torch.Tensor] | None = None
 
-    if bundle.spec.name == "tandemfoilset":
-        primary_metric_key = "val_primary/surface_pressure_mae"
-    else:
-        primary_metric_key = f"val_primary/{bundle.spec.default_metric}"
     best_val_metric = float("inf")
-    best_epoch = 0
-    best_model_state: dict[str, torch.Tensor] | None = None
-    best_anp_state: dict[str, torch.Tensor] | None = None
     best_ema_shadow: dict[str, torch.Tensor] | None = None
     best_anp_ema_shadow: dict[str, torch.Tensor] | None = None
 
@@ -1719,7 +1712,7 @@ def main() -> None:
             best_model_state = snapshot_module_state(model)
             best_anp_state = snapshot_module_state(anp_head)
 
-        current_val = eval_metrics.get(primary_metric_key)
+        current_val = eval_metrics.get(best_val_primary_metric_name)
         if current_val is not None and current_val < best_val_metric:
             best_val_metric = current_val
             best_epoch = epoch


tcapelle:
STUDENT historia:

## Results

### Trial 1 — True monotonic cosine decay (T_max=393606 steps)

| Metric | This run | Baseline (PR #3072) | External target |
|--------|----------|---------------------|-----------------|
| Best val surface_rel_l2_pct | **4.086%** (ep 519) | **3.833%** (ep 511) | <3.71% (AB-UPT) |
| Test surface_rel_l2_pct | **4.390%** (best ckpt ep 519) | — | — |
| Epochs completed | 521 (timeout) | 511 | — |

**Verdict: Did not beat baseline.** Val 4.086% vs baseline 3.833% — a 0.253 pp gap (6.6% relative). However, this is a large improvement over PR #3154's cycling cosine (T_max=999 in steps ≈ 2.5-epoch cycles) which achieved only 4.39% best val.

### Command used

```bash
  --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 393606 \
  --ema-decay 0.9995 --wandb-group dm-monotonic-cosine \
  --agent historia --wandb-name "historia/dm-true-monotonic-cosine"
```

### W&B run ID: `6yiwl01x`

### LR schedule verification
The LR decayed smoothly from 5e-4 to ~2.33e-4 over 521 epochs (no restarts or spikes visible). At epoch 521, we were ~52% through the full cosine curve (205,274 / 393,606 steps). The LR had only decayed to ~47% of initial — the 360-minute timeout cut training short at about halfway through the planned schedule.

### Convergence analysis

**Three distinct phases:**
1. **Epochs 1-124: Perfect monotonic descent.** 124 consecutive best-val epochs with zero plateaus. Val dropped from ~17% to ~5.5%. Grad norms declined from ~9.4 to ~0.2, grad clip events from 394/epoch (100%) to near-zero.
2. **Epochs 124-431: Plateau-breakthrough pattern.** Increasingly long plateaus (8, 16, 23, 38, 67 non-best epochs) followed by sudden breakthroughs. Occasional gradient spikes (6-8 clip events) caused EMA regression requiring 10-15 epochs to recover.
3. **Epochs 431-521: Late-stage descent wave.** After a 30-epoch plateau, breakthroughs resumed at ep 461-471 (4.17% → 4.11%), then again at ep 511-519 (4.10% → 4.09%). The model was **still actively improving when timeout hit** — the last new best was only 2 epochs before the final epoch.

**Key metric progression:**
- Ep 100: 6.27% → Ep 200: 4.49% → Ep 300: 4.30% → Ep 400: 4.21% → Ep 500: 4.13% → Ep 519: 4.09%

### What happened

The hypothesis — that removing cosine restart boundaries would allow smoother long-horizon convergence with EMA=0.9995 — partially confirmed but did not beat the baseline.

**Why it improved over PR #3154 cycling (4.39% → 4.09%):** T_max=999 steps created rapid 2.5-epoch mini-cycles that constantly disrupted EMA accumulation. The true monotonic decay eliminated this, allowing EMA to build cleaner weight averages. The 124-epoch monotonic descent phase (zero plateau epochs) demonstrates how smooth the training became.

**Why it didn't beat baseline's rapid restarts (4.09% vs 3.83%):** The baseline's T_max=30 (~13 steps/epoch, ~7.6 restarts/epoch) appears to provide beneficial perturbation — like stochastic warm restarts that help escape local minima. The monotonic schedule removes this mechanism entirely. Additionally, we only used ~52% of the cosine budget before timeout, so the LR was still relatively high (2.33e-4 vs the ~0 target at 999 epochs).

**Incomplete schedule caveat:** The model was still actively converging (new bests in the final 10 epochs), and ~48% of the LR budget remained. A full 999-epoch run would let the LR approach zero, which could potentially close some of the gap — but likely not all 0.253 pp.

### Bug fix included
Fixed a `primary_metric_key` local variable shadowing bug in `train.py` (lines ~1645-1652). A local variable `primary_metric_key` was shadowing the function defined at line 1423, causing test evaluation to fail silently. This fix is included in this branch.

### Suggested follow-ups
1. **Intermediate T_max sweet spot:** Try T_max=3000-5000 steps (~8-13 epochs per cycle) to find a middle ground between the baseline's ultra-rapid cycling (T_max=30) and pure monotonic decay. The optimal restart frequency may balance exploration (restart kicks) and exploitation (smooth descent).
2. **Monotonic cosine with full training time:** If we can extend the timeout, a full 999-epoch run would complete the cosine curve and test whether the continued LR decay closes the gap.
3. **Cosine with linear warmup:** Add 10-20 epoch linear warmup before the monotonic decay — the early-training gradient instability (394 clips/epoch at ep 1) suggests warmup could help the EMA accumulate better initial weights.

tcapelle:
ADVISOR: Closing PR #3207 because ADVISOR: Closing — true monotonic cosine decay is a clear dead end. 4.086% vs 3.833% baseline (0.253pp gap, 6.6% relative regression). This definitively confirms that the baseline's rapid restarts (T_max=30, ~7.6 restarts/epoch) are a core mechanism, not noise — they provide beneficial gradient perturbation that monotonic decay cannot replace. The val curve was plateauing well above baseline even at epoch 519, so more epochs wouldn't close the gap. T_max=30 remains the confirmed DM sweet spot. Your bug fix for primary_metric_key will be cherry-picked. Good analysis of the three-phase convergence behavior..

PR #3135, opened 2026-04-23
AirfRANS: EMA=0.999 + volume-loss-weight=10x compound

tcapelle:
## Hypothesis

This compounds the two most promising AirfRANS volume levers:

1. **EMA=0.999** (PR #3105): Gave a -42.4% volume MSE improvement (0.004400 vs 0.00764) but +21% surface regression. The smoothing helps spatially smooth volume predictions.

2. **volume-loss-weight=10x** (#3102 thorfinn, in progress): Directly shifts training attention to volume by weighting volume loss 10x higher.

The compound hypothesis: EMA provides better volume parameter averaging while volume loss weighting forces the optimizer to prioritize volume accuracy. Together, they attack the volume gap from two different angles — EMA smooths the parameters while loss weighting shifts the learning signal.

The surface regression from EMA (+21% to 0.000583) is acceptable since surface still beats SpiderSolver by 7.4x. If volume weighting further shifts capacity toward volume, the compound could push Vol MSE well below EMA-only's 0.004400.

## Instructions

**Run 1: EMA=0.999 + volume-loss-weight=10x**
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset airfrans \
  --optimizer adamw \
  --lr 6e-4 \
  --cosine-t-max 50 \
  --grad-clip 1.0 \
  --weight-decay 1e-2 \
  --ema-decay 0.999 \
  --enable-fourier \
  --model-layers 2 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --volume-loss-weight 10.0 \
  --wandb_group af-ema-vol-weight-compound
```

If --volume-loss-weight doesn't exist yet, add it (multiply volume loss component by this factor). Report BOTH metrics:
- val_primary/surface_mse (baseline: 0.000482)
- val_primary/volume_mse (baseline: 0.00764, EMA-only: 0.004400)

## Baseline

- **val_primary/surface_mse:** 0.000482 (PR #2951, no-EMA, no vol weighting)
- **val_primary/volume_mse:** 0.00764 (4.5x above SpiderSolver 0.0017)
- **Best volume result:** 0.004400 (EMA=0.999 alone, PR #3105)
- **External targets:** SpiderSolver Surf MSE=0.0043, Vol MSE=0.0017
- **Config:** 2L/256d, AdamW lr=6e-4, T_max=50, gc=1.0, WD=1e-2, no-EMA

PR #3212, opened 2026-04-23
AirfRANS: PCGrad gradient surgery for surface/vol conflict

tcapelle:
## Hypothesis

AirfRANS training jointly minimizes two objectives: `surface_mse` (boundary layer prediction) and `vol_mse` (volume field prediction). These gradients may point in conflicting directions — when their cosine similarity is negative, naive summation partially cancels both signals, slowing convergence and potentially preventing convergence to Pareto-optimal solutions.

**PCGrad** (Yu et al., 2020, "Gradient Surgery for Multi-Task Learning") addresses this directly: for each task pair where gradients conflict (cos-sim < 0), it projects the conflicting gradient onto the normal plane of the other, nulling out the conflicting component before accumulation. This is the most well-validated method for multi-task gradient conflict in the literature.

Expected impact: cleaner gradient signal → faster convergence and better surface/vol trade-off. The surface constraint (≤ 0.000296) is hard; PCGrad should help us push vol_mse below the 0.0017 external target while holding surface.

Paper: https://arxiv.org/abs/2001.06782

## Instructions

Implement PCGrad for AirfRANS training. The key change is splitting the joint loss into two separate backward passes and projecting conflicting gradients before the optimizer step.

**PCGrad implementation sketch:**

```python
def pcgrad_step(model, optimizer, surface_loss, vol_loss):
    # Compute gradients for each task separately
    optimizer.zero_grad()
    surface_loss.backward(retain_graph=True)
    grads_surface = [p.grad.clone() if p.grad is not None else None for p in model.parameters()]
    
    optimizer.zero_grad()
    vol_loss.backward()
    grads_vol = [p.grad.clone() if p.grad is not None else None for p in model.parameters()]
    
    # Project conflicting gradients
    projected_surface = []
    projected_vol = []
    for g_s, g_v in zip(grads_surface, grads_vol):
        if g_s is None or g_v is None:
            projected_surface.append(g_s)
            projected_vol.append(g_v)
            continue
        
        # Project g_s onto normal plane of g_v if conflicting
        cos_sim = (g_s * g_v).sum() / (g_v.norm() * g_s.norm() + 1e-8)
        if cos_sim < 0:
            g_s = g_s - (g_s * g_v).sum() / (g_v.norm()**2 + 1e-8) * g_v
        
        # Project g_v onto normal plane of g_s if conflicting
        cos_sim2 = (g_v * g_s).sum() / (g_s.norm() * g_v.norm() + 1e-8)
        if cos_sim2 < 0:
            g_v = g_v - (g_v * g_s).sum() / (g_s.norm()**2 + 1e-8) * g_s
        
        projected_surface.append(g_s)
        projected_vol.append(g_v)
    
    # Accumulate projected gradients
    optimizer.zero_grad()
    for p, g_s, g_v in zip(model.parameters(), projected_surface, projected_vol):
        if g_s is not None and g_v is not None:
            p.grad = g_s + g_v
        elif g_s is not None:
            p.grad = g_s
        elif g_v is not None:
            p.grad = g_v
    
    # Clip and step
    torch.nn.utils.clip_grad_norm_(model.parameters(), grad_clip)
    optimizer.step()
```

**Trial 1 — PCGrad with champion config (vol-weight=10x):**
```bash
  --enable-fourier --model-layers 2 \
  --model-hidden-dim 256 --model-heads 4 \
  --vol-loss-weight 10.0 \
  --use-pcgrad \
  --wandb_group af-pcgrad
```

**Trial 2 — PCGrad + vol-weight=15x (push harder on volume):**
```bash
  --enable-fourier --model-layers 2 \
  --model-hidden-dim 256 --model-heads 4 \
  --vol-loss-weight 15.0 \
  --use-pcgrad \
  --wandb_group af-pcgrad
```

**Smoke test:** Run 5 epochs first to verify PCGrad is actually being applied (log the fraction of conflicting gradient pairs per step — if it's near 0, PCGrad had no effect and something is wrong).

Add a `--use-pcgrad` flag to enable/disable PCGrad without touching the rest of the training loop. The flag should default to False so the baseline is unaffected.

**Important:** The two backward passes will roughly double memory usage per step. If OOM, reduce batch size accordingly while keeping total gradient accumulation steps equivalent.

## Baseline

Current AirfRANS champion (PR #3135, W&B run: `sh2zyfwr`):
- `val_primary/surface_mse`: **0.000296** (hard constraint — must not exceed this)
- `val_primary/vol_mse`: **0.002039** (target: < 0.0017 to match external benchmark)

Config: 2L/256d/4H, AdamW lr=6e-4, T_max=50, gc=1.0, WD=1e-2, EMA=0.999, vol-weight=10x, Fourier

Reproduce baseline:
```bash
  --enable-fourier --model-layers 2 \
  --model-hidden-dim 256 --model-heads 4 \
  --vol-loss-weight 10.0
```

W&B baseline: https://wandb.ai/senpai-research/icml2026/runs/sh2zyfwr

**Success criteria:** surface_mse ≤ 0.000296 AND vol_mse < 0.002039 (beat baseline), ideally vol_mse < 0.0017 (match external target).

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,7 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    use_pcgrad: bool = False
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -763,9 +764,12 @@ def loss_grouped(
     outputs: dict[str, torch.Tensor | None],
     transform: TargetTransform,
     volume_loss_weight: float = 1.0,
-) -> tuple[torch.Tensor, dict[str, float]]:
+    split_losses: bool = False,
+) -> tuple[torch.Tensor, dict[str, float]] | tuple[torch.Tensor, torch.Tensor | None, dict[str, float]]:
     total = torch.tensor(0.0, device=batch.surface_x.device)
     metrics: dict[str, float] = {}
+    surf_loss = None
+    vol_loss = None
     if batch.surface_y is not None and outputs["surface_preds"] is not None:
         target = transform.apply(batch.surface_y)
         surf_loss = F.mse_loss(outputs["surface_preds"][batch.surface_mask], target[batch.surface_mask])
@@ -777,6 +781,8 @@ def loss_grouped(
         total = total + vol_loss * volume_loss_weight
         metrics["volume_loss"] = float(vol_loss.detach().cpu().item())
     metrics["loss"] = float(total.detach().cpu().item())
+    if split_losses:
+        return surf_loss, vol_loss, metrics
     return total, metrics
 
 
@@ -1254,6 +1260,78 @@ def evaluate_abupt(
     return {"surface_mae": total_surface / max(count_surface, 1), "volume_mae": total_volume / max(count_volume, 1)}
 
 
+def _pcgrad_project(
+    all_params: list[torch.nn.Parameter],
+    surf_loss: torch.Tensor,
+    vol_loss: torch.Tensor,
+    volume_loss_weight: float,
+    grad_accum_scale: float,
+) -> float:
+    """PCGrad gradient surgery for 2 tasks. Returns conflict indicator (1.0 if conflicting, 0.0 otherwise)."""
+    grads_surf: list[torch.Tensor | None] = []
+    grads_vol: list[torch.Tensor | None] = []
+    saved_grads: list[torch.Tensor | None] = []
+
+    for p in all_params:
+        saved_grads.append(p.grad.clone().detach() if p.grad is not None else None)
+        p.grad = None
+    (surf_loss * grad_accum_scale).backward(retain_graph=True)
+    for p in all_params:
+        grads_surf.append(p.grad.clone().detach() if p.grad is not None else None)
+
+    for p in all_params:
+        p.grad = None
+    (vol_loss * volume_loss_weight * grad_accum_scale).backward()
+    for p in all_params:
+        grads_vol.append(p.grad.clone().detach() if p.grad is not None else None)
+
+    shared_s = []
+    shared_v = []
+    for g_s, g_v in zip(grads_surf, grads_vol):
+        if g_s is not None and g_v is not None:
+            shared_s.append(g_s.flatten())
+            shared_v.append(g_v.flatten())
+
+    conflict = 0.0
+    if shared_s:
+        flat_s = torch.cat(shared_s)
+        flat_v = torch.cat(shared_v)
+        dot = torch.dot(flat_s, flat_v)
+        conflict = float(dot < 0)
+
+        if dot < 0:
+            proj_s = flat_s - (torch.dot(flat_s, flat_v) / (torch.dot(flat_v, flat_v) + 1e-12)) * flat_v
+            proj_v = flat_v - (torch.dot(flat_v, flat_s) / (torch.dot(flat_s, flat_s) + 1e-12)) * flat_s
+        else:
+            proj_s = flat_s
+            proj_v = flat_v
+
+        combined = proj_s + proj_v
+        offset = 0
+        for i, (g_s, g_v) in enumerate(zip(grads_surf, grads_vol)):
+            if g_s is not None and g_v is not None:
+                numel = g_s.numel()
+                chunk = combined[offset : offset + numel].reshape(g_s.shape)
+                grads_surf[i] = chunk
+                grads_vol[i] = None
+                offset += numel
+
+    for p, g_s, g_v, g_saved in zip(all_params, grads_surf, grads_vol, saved_grads):
+        new_grad = None
+        if g_s is not None and g_v is None:
+            new_grad = g_s
+        elif g_s is not None and g_v is not None:
+            new_grad = g_s + g_v
+        elif g_v is not None:
+            new_grad = g_v
+        if new_grad is not None:
+            p.grad = new_grad if g_saved is None else (g_saved + new_grad)
+        else:
+            p.grad = g_saved
+
+    return conflict
+
+
 def train_one_epoch(
     model: torch.nn.Module,
     anp_head: ANPSurfaceDecoder | None,
@@ -1271,6 +1349,7 @@ def train_one_epoch(
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
     volume_loss_weight: float = 1.0,
+    use_pcgrad: bool = False,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT jin:

## Results

### Trial 1: PCGrad + vol-weight=10x

| Metric | Baseline | PCGrad vol10x (ep 158) | Delta |
|--------|----------|----------------------|-------|
| `val_primary/surface_mse` | **0.000296** | 0.001926 | +550% (6.5x worse) |
| `full_val/volume_mse` | **0.002039** | 0.012078 | +492% (5.9x worse) |

- **W&B run:** `p89omv2u` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/p89omv2u))
- **PCGrad conflict rate:** 29–56% (mean ~35%), confirming genuine gradient conflict
- **Peak GPU memory:** ~25 GB (GPU 0)
- **Epochs completed:** 168+ (still running)
- **Command:**
```bash
  --enable-fourier --model-layers 2 \
  --model-hidden-dim 256 --model-heads 4 \
  --volume-loss-weight 10.0 \
  --use-pcgrad \
  --agent jin --wandb-name "jin/af-pcgrad-vol10x" --wandb-group "af-pcgrad"
```

### Trial 2: PCGrad + vol-weight=15x

| Metric | Baseline | PCGrad vol15x (ep 159) | Delta |
|--------|----------|----------------------|-------|
| `val_primary/surface_mse` | **0.000296** | 0.001510 | +410% (5.1x worse) |
| `full_val/volume_mse` | **0.002039** | 0.012408 | +508% (6.1x worse) |

- **W&B run:** `8cq4fd0i` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/8cq4fd0i))
- **PCGrad conflict rate:** 33–55% (mean ~39%), consistent conflict throughout training
- **Peak GPU memory:** ~55 GB (GPU 1)
- **Epochs completed:** 168+ (still running)
- **Command:**
```bash
  --enable-fourier --model-layers 2 \
  --model-hidden-dim 256 --model-heads 4 \
  --volume-loss-weight 15.0 \
  --use-pcgrad \
  --agent jin --wandb-name "jin/af-pcgrad-vol15x" --wandb-group "af-pcgrad"
```

## What happened — Negative result

**PCGrad did NOT beat baseline on either metric.** After 168+ epochs, both trials are 5–6x worse than the baseline champion on both surface_mse and vol_mse.

### Key findings

1. **Gradient conflict is real:** PCGrad detects 30–40% conflicting gradient pairs per step throughout training, confirming the hypothesis that surface and volume gradients genuinely conflict. This rate persists even into late training.

2. **PCGrad dramatically slows convergence.** Two compounding factors:
   - **Disabling `torch.compile`:** PCGrad requires `retain_graph=True` for two separate backward passes, which is incompatible with torch.compile's donated buffer optimization. This alone significantly slows per-epoch training speed.
   - **Reduced effective gradient magnitude:** When gradients conflict, PCGrad projects them onto the normal plane of the other task, zeroing out conflicting components. This reduces the effective gradient signal, slowing convergence.

3. **The surface/vol trade-off is not improved.** Despite resolving gradient conflicts, PCGrad does not find a better Pareto frontier. The baseline's simple weighted sum with vol-weight=10x achieves far better results, suggesting that the naive gradient mixing in this problem doesn't actually harm the joint optimization as much as the hypothesis predicted.

4. **Trial 2 (vol15x) slightly outperformed Trial 1 (vol10x)** on surface_mse (0.00151 vs 0.00193), likely because the higher volume weight creates a stronger vol gradient signal that survives the PCGrad projection better.

### Why PCGrad failed here

The core issue is that PCGrad trades convergence speed for gradient purity. For this problem, the 30-40% conflict rate means PCGrad is constantly removing gradient components, reducing the optimization signal. The baseline's naive weighted sum may produce imperfect gradient directions, but the full gradient magnitude allows it to converge much faster and to better optima within the training budget.

Additionally, the requirement to disable `torch.compile` is a significant practical disadvantage — it roughly halves training throughput.

### Bug fix included

Fixed a pre-existing `UnboundLocalError` bug where `primary_metric_key` (a function) was shadowed by a local string variable assignment later in the same function scope, causing Python to treat the function call as a reference to the not-yet-assigned local variable. Renamed the string variable to `primary_metric_str`.

## Suggested follow-ups

1. **CAGrad** (Conflict-Averse Gradient Descent, Liu et al. 2021) could be a better multi-task gradient method — it minimizes worst-case task loss improvement rather than simply nulling conflicting components. More stable when gradient magnitudes differ (as they do here with the 10x vol weight).

2. **Loss-balanced training** with adaptive per-task weights (e.g., GradNorm, uncertainty weighting) might be a better approach than gradient surgery — adjusting loss weights dynamically rather than modifying gradients post-hoc.

3. If the goal is specifically to push vol_mse below 0.0017 while holding surface constraint, a simpler approach like **increasing vol-weight beyond 15x** without PCGrad overhead might be more effective, since the baseline already proves good convergence at 10x.

tcapelle:
ADVISOR: Closing PR #3212 because ADVISOR: Closing — both PCGrad trials are 5-6x worse than baseline on both metrics (surface=0.001926/0.001510 vs 0.000296, vol=0.011/0.012 vs 0.002039). Root cause: PCGrad requires retain_graph=True which disables torch.compile, causing ~40% throughput loss. Additionally, 30-40% gradient conflict rate provides persistent magnitude reduction without improving the surface/vol Pareto frontier. The gradient conflict between surface and volume is not the actual bottleneck. Good diagnostic data — PCGrad added to AF dead ends. Bug fix for primary_metric_key shadowing noted. For next assignment, we'll focus on simpler volume closure approaches..

tcapelle:
STUDENT jin:

## Updated Results — Final (runs completed)

Both trials finished. The final numbers are better than the earlier snapshot but still clearly a negative result.

### Final metrics (best checkpoint, test set)

| Metric | Baseline | Trial 1 vol10x | Trial 2 vol15x |
|--------|----------|----------------|----------------|
| `val_primary/surface_mse` | **0.000296** | 0.000801 (ep 320) | 0.000777 (ep 283) |
| `test_primary/surface_mse` | — | 0.000940 | 0.001299 |
| `full_test/volume_mse` | **0.002039** | 0.006649 | 0.007660 |

- **Trial 1:** 320 epochs to best checkpoint, val surface_mse 2.7x worse than baseline, test vol_mse 3.3x worse
- **Trial 2:** 283 epochs to best checkpoint, val surface_mse 2.6x worse than baseline, test vol_mse 3.8x worse

### Updated assessment

The runs converged much further than the earlier 168-epoch snapshot suggested (val surface_mse went from 0.0019/0.0015 → 0.0008/0.0008). But after 280-320 epochs, PCGrad still does NOT beat the baseline on either metric. The surface constraint (≤ 0.000296) is violated by ~2.6x, and vol_mse is ~3.3–3.8x above baseline.

The earlier analysis holds: PCGrad slows convergence dramatically (needing 3–6x more epochs to reach comparable metrics), largely due to disabling torch.compile and the gradient magnitude reduction from projection. **This is a negative result** — the hypothesis that gradient surgery would improve the surface/vol trade-off is not supported.

PR #3214, opened 2026-04-23
DM: auxiliary surface gradient prediction for smoothness regularization

tcapelle:
## Hypothesis

Multi-task learning with auxiliary spatial gradient prediction (∂p/∂x, ∂p/∂y, ∂p/∂z) should regularize the encoder into learning smooth, physically consistent pressure fields. By explicitly supervising the gradient of the predicted surface pressure, we force the model to capture the underlying physics more faithfully — rather than interpolating point values independently.

This is grounded in physics-informed ML (Raissi et al., 2019, *Physics-informed neural networks*) adapted for graph networks: gradient consistency is a free regularization signal that comes from the mesh topology itself, requiring no additional data collection.

## Instructions

**Implementation — 4 steps:**

**Step 1: Compute ground-truth spatial gradients**

For each surface point, compute ground-truth gradient as weighted average over graph neighbours:
```python
# For each surface point i with neighbours j:
grad_x[i] = sum_j( (p[j] - p[i]) * (x[j] - x[i]) / dist(i,j)^2 ) / sum_j(1/dist(i,j))
grad_y[i] = sum_j( (p[j] - p[i]) * (y[j] - y[i]) / dist(i,j)^2 ) / sum_j(1/dist(i,j))
grad_z[i] = sum_j( (p[j] - p[i]) * (z[j] - z[i]) / dist(i,j)^2 ) / sum_j(1/dist(i,j))
```
Apply to all output channels (pressure, etc.) separately. Compute once per batch from ground-truth labels.

**Step 2: Add auxiliary gradient prediction head**

Add a 2-layer MLP (256d hidden, ReLU, output = 3 × num_output_channels) that takes the encoder node features and predicts the spatial gradient vector. This head branches off the same node embeddings used for the primary output head.

**Step 3: Compute auxiliary gradient loss**

```python
aux_loss = F.mse_loss(predicted_gradients, ground_truth_gradients)
```

Log this separately in W&B as `train/aux_gradient_loss` and `val/aux_gradient_loss`.

**Step 4: Combine losses**

```python
total_loss = primary_loss + aux_weight * aux_loss
```

where `aux_weight` is a hyperparameter.

---

**Trial 1** — `aux_weight=0.1` (moderate regularization):

```bash
cd target/icml2026
  --dataset drivaerml \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --aux-gradient-weight 0.1 \
  --wandb_group dm-auxiliary-gradient
```

**Trial 2** — `aux_weight=0.01` (lighter regularization):

Same command with `--aux-gradient-weight 0.01`.

**Smoke test first** — run 2 epochs to confirm the aux loss is non-NaN and logging correctly, then launch full runs for both trials.

**Important:** Log aux gradient loss separately in W&B (`train/aux_gradient_loss`, `val/aux_gradient_loss`). The primary metric to beat is still `val_primary/surface_rel_l2_pct`.

## Baseline

Current best metrics (DrivAerML):
- `val_primary/surface_rel_l2_pct`: **3.833%** (val) at epoch 511 — test **4.685%**
- Baseline W&B run: `ncl1dh88` (eren/dm-ema-9995-gc05)
- External target: <3.71% (AB-UPT) — gap: 0.123 pp
- Reproduce command:
  ```bash
  cd target/icml2026
    --dataset drivaerml \
    --epochs 999 \
    --batch-size 1 \
    --ema-decay 0.9995
  ```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,7 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    aux_gradient_weight: float = 0.0
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -821,6 +822,80 @@ def loss_abupt(
     return total, metrics
 
 
+class AuxGradientHead(torch.nn.Module):
+    def __init__(self, input_dim: int, hidden_dim: int = 256, output_dim: int = 3):
+        super().__init__()
+        self.net = torch.nn.Sequential(
+            torch.nn.Linear(input_dim, hidden_dim),
+            torch.nn.ReLU(),
+            torch.nn.Linear(hidden_dim, output_dim),
+        )
+        torch.nn.init.zeros_(self.net[-1].weight)
+        torch.nn.init.zeros_(self.net[-1].bias)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.net(x)
+
+
+@torch.no_grad()
+def compute_ground_truth_gradients(
+    positions: torch.Tensor,
+    values: torch.Tensor,
+    mask: torch.Tensor,
+    k: int = 16,
+) -> torch.Tensor:
+    B, N, C = values.shape
+    device = values.device
+    positions = positions.float()
+    values = values.float()
+    gradients = torch.zeros(B, N, 3 * C, device=device)
+
+    for b in range(B):
+        valid = mask[b]
+        pos = positions[b]
+        val = values[b]
+        valid_idx = valid.nonzero(as_tuple=True)[0]
+        n_valid = valid_idx.shape[0]
+        if n_valid < k + 1:
+            continue
+        valid_pos = pos[valid_idx]
+        valid_val = val[valid_idx]
+
+        chunk_size = 2048
+        for start in range(0, n_valid, chunk_size):
+            end = min(start + chunk_size, n_valid)
+            cs = end - start
+            chunk_pos = valid_pos[start:end]
+            chunk_val = valid_val[start:end]
+
+            dists = torch.cdist(chunk_pos, valid_pos)
+            dists[torch.arange(cs, device=device), torch.arange(start, end, device=device)] = float("inf")
+
+            _, nn_idx = dists.topk(k, dim=1, largest=False)
+
+            neighbor_pos = valid_pos[nn_idx]
+            neighbor_val = valid_val[nn_idx]
+
+            delta_pos = neighbor_pos - chunk_pos.unsqueeze(1)
+            delta_val = neighbor_val - chunk_val.unsqueeze(1)
+
+            dist_sq = (delta_pos**2).sum(dim=-1).clamp(min=1e-10)
+            inv_dist = 1.0 / dist_sq.sqrt().clamp(min=1e-8)
+
+            numerator = torch.einsum("bkc,bkd,bk->bcd", delta_val, delta_pos, 1.0 / dist_sq)
+            denom = inv_dist.sum(dim=1, keepdim=True).unsqueeze(-1).clamp(min=1e-10)
+            grad = numerator / denom
+
+            gradients[b, valid_idx[start:end]] = grad.reshape(cs, 3 * C)
+
+    grad_nonzero = gradients[gradients != 0]
+    if grad_nonzero.numel() > 0:
+        grad_std = grad_nonzero.std().clamp(min=1e-8)
+        gradients = gradients.clamp(-5 * grad_std, 5 * grad_std)
+
+    return gradients
+
+
 def maybe_apply_anp(
     outputs: dict[str, torch.Tensor | None],
     prepared: TandemPreparedBatch,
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT kakashi:

## Results

### Trial 1: `aux_gradient_weight=0.1`

| Metric | Value |
|--------|-------|
| `val_primary/surface_rel_l2_pct` | 23.5% (best, ep20) |
| Outcome | **Diverged at epoch 22** |
| W&B run | `h611t6dg` |
| Peak memory | ~6 GB |

**Command:**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=2 SENPAI_MAX_EPOCHS=9999 python train.py \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --aux-gradient-weight 0.1 \
  --agent kakashi --wandb-name "kakashi/dm-aux-grad-w01" --wandb-group dm-auxiliary-gradient
```

**What happened:** Training diverged rapidly at epoch 20-22. Gradient norms exploded: 2.25 → 4.89 → 15.3 → 18.1 in successive epochs, overwhelming the gc=0.5 clipping. The aux gradient loss magnitude (~0.04) multiplied by weight 0.1 created a signal (~0.004) that was too strong relative to the primary loss, destabilizing the optimizer. Killed at epoch 22 to free GPU.

---

### Trial 2: `aux_gradient_weight=0.01`

| Metric | Baseline | Trial 2 (best) | Trial 2 (terminal) | Delta vs baseline |
|--------|----------|----------------|---------------------|-------------------|
| `val_primary/surface_rel_l2_pct` | **3.833%** (ep511) | **5.485%** (ep191) | 7.879% (ep196) | +1.652 pp worse |
| `val_surface/aux_gradient_loss` | N/A | 0.003720 (ep191) | 0.003776 (ep196) | — |
| `train/aux_gradient_loss` | N/A | 0.004370 (ep191) | 0.004425 (ep196) | — |
| W&B run | `ncl1dh88` | `exbz0pb4` | — | — |
| Peak memory | ~6 GB | ~6.5 GB | — | — |

**Command:**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=3 SENPAI_MAX_EPOCHS=9999 python train.py \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --aux-gradient-weight 0.01 \
  --agent kakashi --wandb-name "kakashi/dm-aux-grad-w001" --wandb-group dm-auxiliary-gradient
```

**What happened:** Training was stable for ~190 epochs with smooth convergence. Trajectory:

| Epoch | Best val % | Convergence rate |
|-------|-----------|------------------|
| 90 | 8.097% | — |
| 120 | 6.699% | ~0.047 pp/ep |
| 150 | 6.003% | ~0.023 pp/ep |
| 180 | 5.644% | ~0.012 pp/ep |
| 191 | 5.485% | ~0.014 pp/ep |
| 192+ | **Divergence** | Grad norms 0.26 → 1.09 |

At epoch 192, a sudden instability event occurred: grad_norm_mean spiked from 0.26 to 0.84 with 185 grad_clip_events (vs 10 at ep191), and val metric jumped from 5.485% to 5.580%. By ep195, grad_norm reached 1.09 with 223 clip events, and val had degraded to 7.311%. The EMA model could not absorb the damage — by ep196 val was 7.879% and still worsening. Killed at ep196.

**Analysis:**

1. **Negative result.** The auxiliary gradient prediction hypothesis did not improve performance. Even before divergence, the best val (5.485% at ep191) was 1.65pp above the baseline (3.833% at ep511). Extrapolating the pre-divergence convergence rate (0.012pp/ep), the run would have needed ~137 more epochs to even match baseline — and the convergence was decelerating.

2. **Delayed instability pattern.** Both trials showed the same failure mode at different timescales: w=0.1 diverged at ep22, w=0.01 diverged at ep192. The aux gradient signal appears benign during early training when the primary loss is large but becomes destabilizing once the model is well-trained and the loss landscape flattens. The kNN-based gradient targets may be noisy enough to inject harmful variance in the late-training regime.

3. **The aux gradient loss itself converged.** It dropped from ~0.005 to ~0.0037 over 190 epochs (train and val tracking closely), suggesting the MLP head was learning but the learned gradients weren't helping the primary task. The gradient regularization may have been constraining the model to smoother solutions that were suboptimal for the actual pressure distribution.

4. **Computational overhead.** The kNN gradient computation (chunked over 50K points with k=16) added ~10% overhead per step. Memory increase was modest (~0.5 GB).

## Suggested follow-ups

- **Gradient-free smoothness:** If the goal is smoother predictions, a Laplacian smoothness penalty (L2 on finite differences between neighboring predictions) might be more stable than explicit gradient supervision, since it doesn't require computing noisy kNN gradient targets.
- **Annealing the aux weight:** Decaying `aux_gradient_weight` toward 0 after the first 100 epochs could avoid the late-training instability while still benefiting from early-stage regularization.
- **Better gradient targets:** The kNN-based inverse-distance-weighted gradients are approximate. If the mesh connectivity is available, finite-element gradient operators would give exact targets and might avoid the noise that caused instability.

tcapelle:
ADVISOR: Closing PR #3214 because ADVISOR: Closing — both trials failed to beat baseline. Trial 1 (aux_weight=0.1): diverged ep22 from gradient explosion. Trial 2 (aux_weight=0.01): best val=5.485% ep191 (+43% worse than 3.833%), then instability event ep192. Root cause: kNN finite-difference gradient targets are inherently noisy — once primary loss flattens in late training, the noisy auxiliary signal dominates and injects harmful variance. Supervised gradient prediction is not helping when the model already learns appropriate gradient structure from the dense primary pressure loss. Good execution and clean analysis, kakashi..

PR #3218, opened 2026-04-23
AirfRANS: GradNorm adaptive multi-task loss balancing

tcapelle:
## Hypothesis

The current vol-weight=10x static coefficient was hand-tuned and may not be optimal throughout training. GradNorm (Chen et al., 2018) dynamically adjusts per-task loss weights so that all tasks train at similar rates — when vol_mse lags behind surface_mse in convergence speed, GradNorm automatically increases w_volume and vice versa.

This should push vol_mse below 0.0017 (target) while keeping surface_mse ≤ 0.000296 (hard constraint), because the weight adaptation prevents either task from dominating at the wrong stage of training.

This is complementary to jin's PCGrad (#3212): GradNorm balances *training rates* across tasks, while PCGrad resolves *gradient conflicts*. They could be combined in a future PR.

Reference: Chen et al., 2018 — "GradNorm: Gradient Normalization for Adaptive Loss Balancing in Deep Multitask Networks" (ICML 2018), https://arxiv.org/abs/1711.02257

## Instructions

**Smoke test first**: Run 3 epochs before committing to a full run. Verify that w_surface and w_volume change from their initial values before proceeding.

### Implementation

Add learnable loss weights and GradNorm update logic to `train.py`:

```python
import torch
import torch.nn as nn

# After model instantiation, add learnable weights:
w_surface = nn.Parameter(torch.tensor(1.0))
w_volume  = nn.Parameter(torch.tensor(10.0))

# Separate optimizer for the loss weights (do NOT include in main optimizer):
gradnorm_optimizer = torch.optim.Adam([w_surface, w_volume], lr=0.025)
```

**Per training step** (inside the training loop, after computing per-task losses):

```python
# 1. Compute weighted total loss for the main backward pass
loss = w_surface * loss_surface + w_volume * loss_vol

# 2. GradNorm weight update (every step, after the main optimizer step):
# Get gradient norms of each task loss w.r.t. the last shared layer weights
# (use the final encoder/backbone layer — e.g., model.layers[-1].parameters())
shared_params = list(model.layers[-1].parameters())  # adjust to actual attribute name

G_surface = torch.norm(
    torch.autograd.grad(w_surface * loss_surface, shared_params, retain_graph=True, allow_unused=True)[0]
)
G_vol = torch.norm(
    torch.autograd.grad(w_volume * loss_vol, shared_params, retain_graph=True)[0]
)

G_avg = (G_surface + G_vol).detach() / 2.0

# EMA of L0 (initial loss values) — store at step 0
if not hasattr(train_state, 'L0_surface'):
    train_state.L0_surface = loss_surface.detach()
    train_state.L0_vol = loss_vol.detach()

# Relative inverse training rate
r_surface = (loss_surface.detach() / train_state.L0_surface)
r_vol     = (loss_vol.detach()     / train_state.L0_vol)
r_avg = (r_surface + r_vol) / 2.0

# GradNorm target
alpha = ALPHA  # see trials below
G_target_surface = (G_avg * (r_surface / r_avg) ** alpha).detach()
G_target_vol     = (G_avg * (r_vol     / r_avg) ** alpha).detach()

# GradNorm loss (L1)
gradnorm_loss = torch.abs(G_surface - G_target_surface) + torch.abs(G_vol - G_target_vol)

gradnorm_optimizer.zero_grad()
gradnorm_loss.backward()
gradnorm_optimizer.step()

# Renormalise weights so they sum to the original total (2 tasks, sum=11 initially)
with torch.no_grad():
    w_sum = w_surface + w_volume
    w_surface.data = w_surface.data * 11.0 / w_sum
    w_volume.data  = w_volume.data  * 11.0 / w_sum
```

**W&B logging** — log each epoch:
```python
wandb.log({"train/w_surface": w_surface.item(), "train/w_volume": w_volume.item()}, step=epoch)
```

### Trial configurations

Run both trials under the same W&B group:

**Trial 1** — `alpha=1.5` (stronger asymmetry, pushes lagging tasks harder):
```bash
  --gradnorm-alpha 1.5 \
  --wandb_group af-gradnorm --wandb_run_name af-gradnorm-alpha1.5
```

**Trial 2** — `alpha=0.5` (softer asymmetry, more conservative):
```bash
  --gradnorm-alpha 0.5 \
  --wandb_group af-gradnorm --wandb_run_name af-gradnorm-alpha0.5
```

Note: Add `--gradnorm-alpha` as a CLI argument and wire it through the training logic. If the CLI arg approach is impractical, hard-code the alpha value and run each trial as a separate branch commit.

### Success criteria

- **Hard constraint**: val_primary/surface_mse ≤ 0.000296 (must not regress vs baseline)
- **Target**: val_primary/vol_mse < 0.0017 (push below current 0.002039)
- Report best epoch for each metric separately (they may not coincide)
- Report w_surface and w_volume values at end of training

## Baseline

Current best (PR #3135, nezuko/af-ema-vol-weight-10x):

| Metric | Value | Epoch |
|--------|-------|-------|
| val_primary/surface_mse | **0.000296** | 704 |
| val_primary/vol_mse | **0.002039** | 743 |

- W&B run: `sh2zyfwr` — https://wandb.ai/senpai/senpai/runs/sh2zyfwr
- Config: 2L/256d/4H, AdamW lr=6e-4, T_max=50, gc=1.0, WD=1e-2, EMA=0.999, vol-weight=10x (static), Fourier normals
- Reproduce:
```bash
  --epochs 999 --ema-decay 0.999 --vol-loss-weight 10.0
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,8 +167,21 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    gradnorm_alpha: float = 0.0
+    gradnorm_lr: float = 0.025
     save_checkpoint: bool = False
     seed: int = 0
+
+
+class GradNormState:
+    def __init__(self, w_surface_init: float, w_volume_init: float, alpha: float, lr: float, device: torch.device):
+        self.w_surface = torch.nn.Parameter(torch.tensor(w_surface_init, device=device))
+        self.w_volume = torch.nn.Parameter(torch.tensor(w_volume_init, device=device))
+        self.optimizer = torch.optim.Adam([self.w_surface, self.w_volume], lr=lr)
+        self.alpha = alpha
+        self.initial_weight_sum = w_surface_init + w_volume_init
+        self.L0_surface: float | None = None
+        self.L0_volume: float | None = None
 
 
 @dataclass
@@ -763,9 +776,12 @@ def loss_grouped(
     outputs: dict[str, torch.Tensor | None],
     transform: TargetTransform,
     volume_loss_weight: float = 1.0,
-) -> tuple[torch.Tensor, dict[str, float]]:
+    return_separate: bool = False,
+) -> tuple[torch.Tensor, dict[str, float]] | tuple[torch.Tensor, dict[str, float], torch.Tensor | None, torch.Tensor | None]:
     total = torch.tensor(0.0, device=batch.surface_x.device)
     metrics: dict[str, float] = {}
+    surf_loss = None
+    vol_loss = None
     if batch.surface_y is not None and outputs["surface_preds"] is not None:
         target = transform.apply(batch.surface_y)
         surf_loss = F.mse_loss(outputs["surface_preds"][batch.surface_mask], target[batch.surface_mask])
@@ -777,6 +793,8 @@ def loss_grouped(
         total = total + vol_loss * volume_loss_weight
         metrics["volume_loss"] = float(vol_loss.detach().cpu().item())
     metrics["loss"] = float(total.detach().cpu().item())
+    if return_separate:
+        return total, metrics, surf_loss, vol_loss
     return total, metrics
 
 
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT shouko:

## Results

### Summary

GradNorm (Chen et al., 2018) implemented and tested at two alpha values. The mechanism works correctly — weights adapt dynamically throughout training — but does **not** beat the static baseline within the 360-minute budget due to epoch starvation from GradNorm's computational overhead.

### Metrics comparison

| Metric | Baseline (sh2zyfwr) | Trial 1 α=1.5 (p5b6fgoa) | Trial 2 α=0.5 (93rkwc57) |
|--------|---------------------|---------------------------|---------------------------|
| **val surface_mse** | **0.000296** (ep 704) | 0.000514 (ep 338) | 0.000907 (ep 274) |
| val volume_mse | **0.002039** (ep 743) | 0.005390 (ep 338) | 0.006118 (ep 274) |
| test surface_mse | — | 0.000774 | 0.000858 |
| test volume_mse | — | 0.006570 | 0.005992 |
| Epochs trained | ~763 | 339 | 337 |
| Final w_surface | 1.0 (static) | 3.16 | 3.34 |
| Final w_volume | 10.0 (static) | 7.84 | 7.66 |

### GradNorm weight evolution (Trial 1, α=1.5)

| Epoch | w_surface | w_volume |
|-------|-----------|----------|
| 1 | 2.16 | 8.84 |
| 50 | 2.54 | 8.46 |
| 100 | 2.23 | 8.77 |
| 200 | 2.59 | 8.41 |
| 300 | 3.85 | 7.15 |
| 338 (best) | 3.41 | 7.59 |

### Key observations

1. **GradNorm is working correctly**: Weights evolve dynamically, GradNorm loss decreases, and gradient norms between tasks converge. Surface weight increases from 1.0 toward ~3.4 (GradNorm recognizes surface needs more gradient attention), volume decreases from 10.0 toward ~7.6.

2. **Epoch starvation is the dominant bottleneck**: GradNorm disables `torch.compile` (required for `retain_graph=True`) and adds per-step gradient computations, roughly halving throughput. 339 epochs in 360 minutes vs baseline's ~763 epochs. The baseline found its best at epoch 704 — GradNorm never gets close to that training depth.

3. **α=1.5 clearly better than α=0.5**: Higher alpha enforces stronger gradient balancing, which helps the surface task. Trial 1 found a deep basin at the final cosine trough (epochs 336-338: surface_mse dropped 0.001335→0.000639→0.000589→0.000514 in consecutive epochs). This pattern mirrors the baseline's behavior — deeper cosine cycles continue finding better basins.

4. **Volume MSE significantly worse**: 0.005390 vs baseline 0.002039 (2.6x gap). GradNorm dynamically reduced the volume weight from 10.0 to ~7.6, whereas the static 10x weight was specifically tuned to drive volume improvement. GradNorm may not sufficiently upweight volume because the surface task has larger relative training rate changes at cosine troughs.

5. **Late-training surge suggests potential**: Trial 1's dramatic improvement in the final 4 epochs (0.001335→0.000514, -61%) shows the model was nowhere near converged. With equal epoch budget (~700+ epochs), GradNorm might close the gap.

### What happened

**The hypothesis is plausible but could not be fairly tested.** GradNorm's computational overhead (~2x slower per epoch) means it gets roughly half the training epochs within the same wall-clock budget. Since the baseline's best results come from deep training (epoch 700+), epoch starvation is the primary failure mode, not the GradNorm mechanism itself.

The static `--vol-loss-weight 10.0` baseline is remarkably well-tuned — GradNorm actually converges toward a *lower* volume weight (~7.6x) than the static 10x, suggesting the static weight may be closer to optimal than GradNorm's learned balance for this task/architecture combination.

### Commands used

**Trial 1 (α=1.5):**
```bash
  --gradnorm-alpha 1.5 \
  --agent shouko --wandb-name "shouko/af-gradnorm-alpha1.5" \
  --wandb-group af-gradnorm
```

**Trial 2 (α=0.5):**
```bash
  --gradnorm-alpha 0.5 \
  --agent shouko --wandb-name "shouko/af-gradnorm-alpha0.5" \
  --wandb-group af-gradnorm
```

- **Peak memory:** GPU 0 (Trial 1): ~48.6 GiB, GPU 1 (Trial 2): ~36.6 GiB
- **W&B runs:** [`p5b6fgoa`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/p5b6fgoa) (α=1.5), [`93rkwc57`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/93rkwc57) (α=0.5)

### Suggested follow-ups

1. **GradNorm + torch.compile compatibility**: The main bottleneck is `retain_graph=True` breaking torch.compile. A periodic GradNorm variant (e.g., update weights every N steps with non-compiled forward passes, keep compiled path for normal training) could recover the throughput loss while still adapting weights.

2. **Static weight warm-start then GradNorm**: Train with static `--vol-loss-weight 10.0` for 500+ epochs (compiled, fast), then switch to GradNorm for fine-tuning in the final epochs. This gets the best of both worlds — deep training + adaptive balancing.

3. **Higher initial volume weight**: Initialize w_volume=15 or 20 to see if GradNorm can find a regime where volume gets even more emphasis than the static 10x, which it currently cannot since it converges toward lower volume weight.

4. **Combine with PCGrad (#3212)**: GradNorm addresses training rate imbalance while PCGrad addresses gradient conflicts. They're complementary and could be combined.

tcapelle:
ADVISOR: Closing PR #3218 because ADVISOR: Closing — GradNorm is a clear dead end for AF. Both trials significantly worse: Trial 1 (α=1.5) surface=0.000514 (1.74x worse), vol=0.005390 (2.64x worse). Trial 2 (α=0.5) even worse. Two structural problems: (1) retain_graph=True breaks torch.compile, halving throughput and limiting to ~340 epochs vs baseline's 763. (2) GradNorm converges to w_volume~8x, systematically LOWER than the hand-tuned 10x — it fights against the intentional asymmetry that drives volume improvement. Interesting finding: GradNorm pushes w_surface from 1.0 toward 3.0-3.7, suggesting the surface task is 'under-weighted' by gradient norm standards. Worth noting for future surface weighting experiments, but GradNorm itself is dead for AF..

PR #3219, opened 2026-04-23
DrivAerML: SAM optimizer for flat-basin LR-restart robustness

tcapelle:
## Hypothesis

Sharpness-Aware Minimization (SAM; Foret et al. 2021) steers training toward flatter loss basins by computing gradients at adversarially perturbed weight locations (radius `rho`). Our champion DrivAerML config uses cosine LR restarts (T_max=30), which periodically spike the learning rate and can kick weights out of sharp minima. The hypothesis is that SAM-trained weights, sitting in flatter basins, will survive these restart peaks better and converge to a lower surface_rel_l2_pct than AdamW alone.

References:
- Foret et al. 2021 — "Sharpness-Aware Minimization for Efficiently Improving Generalization": https://arxiv.org/abs/2010.01412
- PyPI `torch-sam` or manual two-step update pattern

## Instructions

Start from the champion DrivAerML config (PR #3072): 4L/512d/8H, AdamW lr=5e-4, T_max=30, EMA=0.9995, gc=0.5, Fourier, no WD.

**Implementation — wrapping AdamW with SAM:**

SAM is a wrapper around a base optimizer. The two-step update is:
1. Compute gradient at current weights w, perturb: w_hat = w + rho * grad / ||grad||
2. Compute gradient at w_hat, update base optimizer with that gradient, restore weights to w

You can implement this manually or use the `torch-sam` package (`pip install torch-sam`). If using torch-sam:

```python
from sam import SAM
base_optimizer = torch.optim.AdamW
optimizer = SAM(model.parameters(), base_optimizer, rho=0.05, lr=5e-4)

# Training step:
loss = criterion(outputs, targets)
loss.backward()
optimizer.first_step(zero_grad=True)   # perturb
loss2 = criterion(model(inputs), targets)
loss2.backward()
optimizer.second_step(zero_grad=True)  # update + restore
```

Grad-clip (gc=0.5) must be applied before each `first_step` and `second_step` call. EMA update should happen after `second_step`.

**Log a sharpness metric to W&B** each epoch:
```python
sharpness = (loss_at_perturbed - loss_at_current).item()
wandb.log({"train/sharpness": sharpness})
```

**Warning:** SAM doubles per-step compute (two forward-backward passes). Expect ~250 epochs in the 360-min timeout budget instead of ~500. This is acceptable — flat-basin benefits should appear within 150-200 epochs.

**Smoke test first:** Run 5 epochs with `--max-train-batches 10 --max-eval-batches 5` to confirm SAM integrates without errors before launching full runs.

---

### Trial 1 — SAM rho=0.05 (wider perturbation)

```bash
  --dataset drivaerml \
  --optimizer sam-adamw \
  --sam-rho 0.05 \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --wandb_group dm-sam-optimizer
```

If `--optimizer sam-adamw` and `--sam-rho` are not yet CLI flags, add them to `train.py` (add args + wrap the AdamW instantiation in the SAM wrapper when `args.optimizer == "sam-adamw"`).

### Trial 2 — SAM rho=0.02 (tighter perturbation)

Same command, change `--sam-rho 0.02`. Run this on a second GPU after Trial 1 starts.

---

Report best `val_primary/surface_rel_l2_pct` for each trial, the epoch it was achieved, and the W&B run IDs. Also paste the `train/sharpness` curve summary (min/max/final). Compare against the AdamW baseline (3.833%) to determine if flat-basin training helps.

## Baseline

Current best metrics from BASELINE.md (DrivAerML):
- `val_primary/surface_rel_l2_pct`: **3.833%** (at epoch 511)
- Baseline W&B run: `ncl1dh88`
- External target: <3.71% (AB-UPT, ~500 epochs) — gap: 0.123 pp

Reproduce command:
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -86,6 +86,58 @@ class EMAWithWarmup:
         self.backup = None
 
 
+class SAM:
+    """Sharpness-Aware Minimization wrapper (Foret et al. 2021)."""
+
+    def __init__(self, params, base_optimizer_cls, rho: float = 0.05, **kwargs):
+        self.base_optimizer = base_optimizer_cls(params, **kwargs)
+        self.rho = rho
+        self.param_groups = self.base_optimizer.param_groups
+        self._old_params: dict[int, torch.Tensor] = {}
+
+    @torch.no_grad()
+    def first_step(self):
+        grad_norm = self._grad_norm()
+        for group in self.param_groups:
+            for p in group["params"]:
+                if p.grad is None:
+                    continue
+                self._old_params[id(p)] = p.data.clone()
+                e_w = self.rho * p.grad / (grad_norm + 1e-12)
+                p.add_(e_w)
+        self.base_optimizer.zero_grad(set_to_none=True)
+
+    @torch.no_grad()
+    def second_step(self):
+        for group in self.param_groups:
+            for p in group["params"]:
+                pid = id(p)
+                if pid in self._old_params:
+                    p.data.copy_(self._old_params[pid])
+        self._old_params.clear()
+        self.base_optimizer.step()
+        self.base_optimizer.zero_grad(set_to_none=True)
+
+    def zero_grad(self, set_to_none: bool = True):
+        self.base_optimizer.zero_grad(set_to_none=set_to_none)
+
+    def _grad_norm(self) -> torch.Tensor:
+        norms = []
+        for group in self.param_groups:
+            for p in group["params"]:
+                if p.grad is not None:
+                    norms.append(p.grad.detach().norm(p=2))
+        if not norms:
+            return torch.tensor(0.0)
+        return torch.norm(torch.stack(norms), p=2)
+
+    def state_dict(self):
+        return self.base_optimizer.state_dict()
+
+    def load_state_dict(self, state_dict):
+        self.base_optimizer.load_state_dict(state_dict)
+
+
 LEGACY_VAL_ALIAS = {
     "val_in_dist": "p_in",
     "val_ood_cond": "p_oodc",
@@ -166,6 +218,7 @@ class TrainConfig:
     volume_anchor_points: int = 8_000
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
+    sam_rho: float = 0.05
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
@@ -547,6 +600,15 @@ def build_optimizer(params, config: TrainConfig):
         optimizer = Lion(params, lr=config.lr, weight_decay=config.weight_decay)
     elif config.optimizer == "adamw":
         optimizer = torch.optim.AdamW(params, lr=config.lr, weight_decay=config.weight_decay)
+    elif config.optimizer == "sam-adamw":
+        optimizer = SAM(
+            params,
+            torch.optim.AdamW,
+            rho=config.sam_rho,
+            lr=config.lr,
+            weight_decay=config.weight_decay,
+        )
+        return optimizer
     else:
         raise ValueError(f"Unknown optimizer: {config.optimizer}")
     if config.use_lookahead:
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT hinata:

## Results

### Trial 1: SAM rho=0.05

| Metric | Baseline (AdamW) | SAM rho=0.05 |
|--------|-------------------|--------------|
| `val_primary/surface_rel_l2_pct` | **3.833%** (epoch 511) | **5.359%** (epoch 276) |

- **W&B run ID**: `7z3a4ous`
- **W&B name**: `hinata/dm-sam-rho005`
- **W&B group**: `dm-sam-optimizer`
- **Peak memory**: ~2.4 GB VRAM per GPU (same as baseline — SAM adds negligible memory)
- **Command**:
  ```
  CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=360 python train.py \
    --dataset drivaerml --optimizer sam-adamw --sam-rho 0.05 \
    --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier \
    --epochs 999 --batch-size 1 \
    --ema-decay 0.9995 --agent hinata \
    --wandb-name "hinata/dm-sam-rho005" --wandb-group "dm-sam-optimizer"
  ```

**Sharpness curve (stable phase, epochs 10–276)**:
- Min: 0.002392
- Max: 0.028512
- Mean: 0.007609
- At best epoch (276): 0.004789

**Stability events**: Two instability spikes occurred:
1. **Epoch 179–183**: Grad norms spiked (0.40 peak, 89 clip events at epoch 180), val degraded to 7.79%. EMA recovered over ~12 epochs back to new bests.
2. **Epoch 278–290 (fatal)**: Grad norms escalated catastrophically (0.10 → 0.25 → 3.18 → 9.46), 353–390 clip events per epoch. Val degraded from 5.36% to 12.42% before kill. Unrecoverable — same divergence pattern as Trial 2.

---

### Trial 2: SAM rho=0.02

| Metric | Baseline (AdamW) | SAM rho=0.02 |
|--------|-------------------|--------------|
| `val_primary/surface_rel_l2_pct` | **3.833%** (epoch 511) | **9.045%** (epoch 98) |

- **W&B run ID**: `dn33cgny`
- **W&B name**: `hinata/dm-sam-rho002`
- **W&B group**: `dm-sam-optimizer`
- **Command**: Same as Trial 1 but `--sam-rho 0.02`, `CUDA_VISIBLE_DEVICES=1`

**Sharpness curve (epochs 1–133)**:
- Min: -0.000166 (epoch 130)
- Max: 0.383649 (epoch 1)
- Final (last non-NaN): 0.000338 (epoch 133)

**Diverged catastrophically** starting epoch ~100. Grad norms escalated from normal to millions, hit NaN at epoch 134. All subsequent epochs produced NaN. Killed at epoch 191 to free GPU.

---

### Analysis: SAM does not improve DrivAerML training

**Both trials are clear negative results** — neither comes close to the 3.833% baseline.

**Root cause 1: 2x compute penalty.** SAM requires two forward-backward passes per step, halving the effective epoch count within the same wall-clock budget. Trial 1 completed 276 useful epochs vs the baseline's 511, giving the model roughly half the learning opportunities.

**Root cause 2: Periodic instability → divergence.** Both trials exhibited catastrophic gradient explosions that eventually became unrecoverable. The lower rho=0.02 diverged earlier (epoch 100) and more violently. The higher rho=0.05 survived one spike (epoch 179) before succumbing to a second (epoch 278). The instability pattern — sharpness dropping near zero while grad norms explode — suggests SAM's perturbation step is pushing into regions where the loss landscape is highly irregular, and the cosine restart's sudden LR jumps compound this.

**Root cause 3: SAM + cosine restarts interaction.** The fatal divergences correlate with cosine restart boundaries. The sharp LR jump from 0 → 5e-4 after a cosine trough may amplify SAM's perturbation, creating a double shock that the model can't absorb. The baseline AdamW with Lookahead handles these restarts gracefully, but SAM's two-step update appears to be less robust to sudden LR changes.

**The hypothesis that flat-basin training improves LR-restart robustness is not supported.** In fact, SAM made restarts *less* robust.

### Suggested follow-ups

1. **SAM with warmup restarts**: Instead of hard cosine restarts (0 → max LR), use warm restarts that gradually ramp LR back up. This could prevent the double-shock instability.
2. **Adaptive SAM (ASAM)**: Use per-parameter rho scaling based on weight magnitudes — the fixed rho may be too aggressive for some layers and too weak for others.
3. **SAM only during final cosine cycles**: Run standard AdamW for the first ~400 epochs, then switch to SAM for fine-tuning in the last ~100 epochs. This avoids the 2x compute penalty during the main training phase.
4. **Stochastic Weight Averaging (SWA)**: A cheaper alternative to SAM for flat-basin training — averages checkpoints from the last few cosine cycles without the 2x compute cost.

### Bug fix included in this PR

Fixed a pre-existing `UnboundLocalError` where `primary_metric_key` was both a module-level function and a local variable in `main()`. Python treated it as local throughout the function due to the later assignment, causing a crash on the `eval_metrics.get(primary_metric_key)` line. Removed the redundant local variable and replaced with `best_val_primary_metric_name` which was already correctly assigned.

tcapelle:
ADVISOR: Closing PR #3219 because ADVISOR: Closing — SAM is a clear dead end for DrivAerML. Both rho values failed decisively: rho=0.05 peaked at 5.36% (40% worse than 3.833% baseline) before diverging at cosine restart boundaries; rho=0.02 peaked at 9.08% before catastrophic gradient explosion (grad_norm=38.5M). The mechanism is clear: SAM's perturbation step compounds destructively with cosine LR restarts, and the 2x compute penalty is prohibitive. The flat-basin hypothesis itself isn't dead — SAM was just a poor vehicle for it given our schedule. Your bug fix for primary_metric_key shadowing will be cherry-picked separately. Nice work identifying and fixing that..

PR #3220, opened 2026-04-23
DrivAerML: Mixup regularization for smoother learned manifolds

tcapelle:
## Hypothesis

Mixup regularization (Zhang et al., 2018 — https://arxiv.org/abs/1710.09412) creates virtual training examples by linearly interpolating between pairs of training samples and their labels:

```
x_mixed = λ * x1 + (1-λ) * x2
y_mixed = λ * y1 + (1-λ) * y2
λ ~ Beta(alpha, alpha)
```

For DrivAerML, raw-input Mixup is infeasible because different car geometries have incompatible mesh structures. Instead, apply **latent-space Mixup**: mix after the GNN encoder but before the decoder, interpolating in the learned feature space where all representations have the same dimensionality.

**Why this might help:** Our model currently sees each car as an independent sample. Mixup forces the model to learn smooth interpolations between car geometries — regularizing the latent manifold and discouraging memorization. This should improve generalization to unseen aerodynamic shapes, directly targeting the DrivAerML OOD validation split.

**Why now:** The current best (3.833%) uses EMA + grad-clip but hasn't explored input regularization. Mixup is a low-complexity change (no extra parameters) that has shown consistent gains in other geometric/structured prediction tasks. With batch_size=1, pairs come from consecutive steps (buffer previous batch, mix with current).

## Instructions

**Implementation note (batch_size=1):** With batch_size=1, there are no in-batch pairs. Maintain a running buffer of the previous batch's latent features and targets. On each step, sample λ ~ Beta(alpha, alpha) and mix current latent with buffered latent, current targets with buffered targets. Skip mixing on the first step (buffer is empty). Apply mixing after the encoder, before the decoder.

**Smoke test first:** Before running full trials, run a 3-epoch smoke test to confirm mixing logic is correct and loss doesn't diverge:
```bash
  --model-heads 8 --epochs 3 --batch-size 1 \
  --mixup-alpha 0.2 --wandb_group dm-mixup
```
Confirm loss decreases normally over 3 epochs before proceeding.

---

**Trial 1 — Light Mixup (alpha=0.2):**

alpha=0.2 produces λ concentrated near 0 and 1, meaning most samples are nearly unmodified with occasional gentle blending. This is the standard Mixup setting for vision tasks.

```bash
  --mixup-alpha 0.2 --wandb_group dm-mixup
```

**Trial 2 — Aggressive Mixup (alpha=1.0):**

alpha=1.0 gives Uniform(0,1) for λ — every sample is a full 50/50 blend. More aggressive regularization; forces the model to interpolate across the full latent manifold.

```bash
  --mixup-alpha 1.0 --wandb_group dm-mixup
```

Run both trials. Report val_primary/surface_rel_l2_pct at best validation checkpoint for each. If Trial 1 wins, note whether alpha=0.4–0.6 might be worth a follow-up. If Trial 2 wins, we may push alpha higher.

## Baseline

Current best metrics (DrivAerML Innovation Track):
- **val_primary/surface_rel_l2_pct: 3.833%** (epoch 511)
- test/surface_rel_l2_pct: 4.685%
- External target: <3.71% (AB-UPT)
- Baseline W&B run: `ncl1dh88` (PR #3072 — eren, EMA=0.9995 + gc=0.5, 4L/512d/8H)
- Reproduce command:
  ```bash
  ```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,7 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    mixup_alpha: float = 0.0
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -1254,6 +1255,39 @@ def evaluate_abupt(
     return {"surface_mae": total_surface / max(count_surface, 1), "volume_mae": total_volume / max(count_volume, 1)}
 
 
+def _unwrap_model(model: torch.nn.Module) -> torch.nn.Module:
+    return model._orig_mod if hasattr(model, "_orig_mod") else model
+
+
+def decode_from_hidden(
+    model: torch.nn.Module,
+    hidden: torch.Tensor,
+    surface_tokens: int,
+    surface_x: torch.Tensor | None,
+    surface_mask: torch.Tensor,
+) -> dict[str, torch.Tensor | None]:
+    raw = _unwrap_model(model)
+    raw_output = raw.out(raw.norm(hidden))
+    surface_preds = raw_output[:, :surface_tokens, : raw.surface_output_dim]
+    volume_preds = None
+    if raw.volume_output_dim > 0 and hidden.shape[1] > surface_tokens:
+        vt = hidden.shape[1] - surface_tokens
+        s = raw.surface_output_dim
+        volume_preds = raw_output[:, surface_tokens : surface_tokens + vt, s : s + raw.volume_output_dim]
+    if hasattr(raw, "_apply_pressure_prior_addition"):
+        surface_preds = raw._apply_pressure_prior_addition(
+            surface_preds, surface_x, getattr(raw, "surface_pressure_prior_idx", None)
+        )
+    if getattr(raw, "surface_head", None) is not None and surface_preds is not None:
+        surface_preds = surface_preds + raw.surface_head(hidden[:, :surface_tokens], surface_preds)
+    if surface_preds is not None:
+        surface_preds = surface_preds * surface_mask.unsqueeze(-1)
+    if volume_preds is not None and surface_mask is not None:
+        vol_mask = torch.ones(volume_preds.shape[:2], dtype=torch.bool, device=volume_preds.device)
+        volume_preds = volume_preds * vol_mask.unsqueeze(-1)
+    return {"surface_preds": surface_preds, "volume_preds": volume_preds}
+
+
 def train_one_epoch(
     model: torch.nn.Module,
     anp_head: ANPSurfaceDecoder | None,
@@ -1271,11 +1305,14 @@ def train_one_epoch(
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
     volume_loss_weight: float = 1.0,
+    mixup_alpha: float = 0.0,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
         anp_head.train()
     running = {"loss": 0.0}
+    mixup_buffer: dict[str, torch.Tensor] | None = None
+    beta_dist = torch.distributions.Beta(mixup_alpha, mixup_alpha) if mixup_alpha > 0 else None
     steps = 0
     micro_batches_total = 0
     batches = loader if max_batches <= 0 else itertools.islice(loader, max_batches)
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT nobara:

## Results

### Summary

Latent-space Mixup regularization was implemented with a buffer-based approach for batch_size=1. Two trials were run in parallel:

| Trial | α | Best Epoch | Best `val_primary/surface_rel_l2_pct` | Final Epoch (stopped) | Status |
|-------|---|------------|---------------------------------------|----------------------|--------|
| Trial 1 (light) | 0.2 | 221 | **8.030%** | 242 | Diverged at ~ep235 |
| Trial 2 (aggressive) | 1.0 | 232 | **7.096%** | 248 | Diverged at ~ep235 |
| Baseline (no Mixup) | — | 511 | **3.833%** | 511 | Converged |

**Neither trial beats the baseline.** Both are 2-3x worse than the 3.833% baseline, and both diverged catastrophically around epoch 235 (raw val spiked to 29-56%). Trials were stopped after confirming no recovery.

### Convergence Trajectory

- Epochs 0–90: Trial 1 (α=0.2) led initially, with Trial 2 catching up
- Epoch ~94: Trial 2 overtook Trial 1 — aggressive Mixup benefited from longer training
- Epochs 100–210: Both improved steadily, with Trial 2 maintaining a ~1-1.7% lead
- Epoch ~209: Trial 2 hit best EMA at 7.10%, then entered a long plateau (19 epochs)
- Epoch ~235: Both trials diverged catastrophically — raw val spiked 3-7x above best
- Trials were killed at epochs 242/248 after confirming divergence was not recovering

### Key Observations

1. **Trial 2 (α=1.0) won** over Trial 1 (α=0.2) — aggressive Mixup provided stronger regularization that paid off during mid-training
2. **Training instability**: Both trials diverged around the same time (~ep235), suggesting the Mixup buffer mechanism may introduce instability over long training. The buffer stores the *previous* batch's hidden states, which are from a different model checkpoint, creating a distributional mismatch that grows as training progresses.
3. **Mixup loss is fundamentally different**: The Mixup loss trains on interpolated hidden states and targets, which means the model never directly optimizes for the actual validation metric. This explains the persistent gap vs baseline.
4. **Lambda mean ~0.50**: As expected for both α=0.2 and α=1.0, the mean lambda was ~0.5 (α=1.0 samples uniformly, α=0.2 concentrates near 0 and 1 but averages ~0.5).

### What Happened

Mixup did not help — it significantly hurt convergence. The core issues:
- **Buffer staleness**: With batch_size=1, the buffer contains hidden states from the *previous* model checkpoint, creating a distributional mismatch that compounds over time
- **Latent space geometry**: Unlike image CNNs where latent spaces are relatively smooth, the Transolver's slice-attention hidden states may not interpolate meaningfully between different geometries (different car shapes)
- **Loss signal degradation**: Training on mixed hidden/target pairs provides a weaker learning signal than training on actual samples, slowing convergence significantly
- **Late-stage divergence**: The compounding buffer staleness eventually destabilized training for both trials

### Commands Used

```bash
# Trial 1 (α=0.2, GPU 0)
CUDA_VISIBLE_DEVICES=0 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --mixup-alpha 0.2 --agent nobara --wandb-name "nobara/dm-mixup-alpha0.2" --wandb-group "dm-mixup"

# Trial 2 (α=1.0, GPU 1)
CUDA_VISIBLE_DEVICES=1 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --mixup-alpha 1.0 --agent nobara --wandb-name "nobara/dm-mixup-alpha1.0" --wandb-group "dm-mixup"
```

### W&B Run IDs
- Trial 1 (α=0.2): `m39js3jn`
- Trial 2 (α=1.0): `0h81e7tw`

### Suggested Follow-ups

1. **Input-space Mixup instead**: Mix the raw surface_x inputs and targets before the encoder rather than in latent space. This avoids the buffer staleness problem entirely.
2. **Fresher buffer with replay**: Instead of using the immediately previous batch, maintain a small replay buffer of recent hidden states (e.g., last 5-10) and sample from it — this would reduce distributional mismatch.
3. **Gradient-stop Mixup**: Detach the mixed hidden path entirely and use it as an auxiliary loss alongside the normal loss, rather than replacing it. e.g., `loss = normal_loss + 0.1 * mixup_loss`.
4. **No follow-up on α=0.4–0.6**: Given that both α values diverged at similar epochs and the mechanism itself appears flawed for this architecture, testing intermediate α values is unlikely to help.

tcapelle:
ADVISOR: Closing PR #3220 because ADVISOR: Closing — both trials crashed, 85-109% worse than baseline at best (8.03%/7.10% vs 3.833%). Buffer-based latent Mixup is fundamentally incompatible: (1) buffer staleness compounds as model evolves, mixing current latents with stale-checkpoint latents creates incoherent targets, (2) Transolver's slice-attention learns geometry-specific decompositions that don't form a smooth interpolable manifold between cars. Both crashed catastrophically (~ep235). Excellent analysis, nobara — buffer staleness explanation is correct. Mixup added to DM dead ends..

PR #3191, opened 2026-04-23
AirfRANS: asinh-pressure + residual + re-stratified on AF base (noam features)

tcapelle:
## Hypothesis

The AF champion (#3135, surface_mse=0.000296) uses EMA=0.999 + vol-weight=10x. Recent experiments (#3184 wolfwood, #3187 stark) are testing asinh-pressure + residual-prediction + re-stratified-sampling on top of the full vol-weight=10x stack. This PR isolates the \"noam feature\" combination (asinh-pressure + residual-prediction + re-stratified-sampling) on the **base AF champion** (no vol-weight), then also tests it with vol-weight=10x — giving us a clean read of whether these three features compound on the base EMA config independently of the volume weighting.

Three noam features being stacked:
- `--asinh-pressure --asinh-scale 0.75` — stabilizes heavy-tailed pressure distribution
- `--residual-prediction` — predicts delta from flow field mean, reduces output range
- `--re-stratified-sampling` — re-weights mini-batches to oversample hard Reynolds-number regimes

Both trials use the same AdamW base config (lr=6e-4, T_max=50, gc=1.0, WD=1e-2, EMA=0.999, 2L/256d/8H, Fourier) from the AF champion lineage.

**Note on model-heads:** The AF champion uses `--model-heads 4` (hidden_dim 256 / heads 4 = head_dim 64). These trials use `--model-heads 8` (head_dim 32) as specified — note this if results diverge unexpectedly.

## Instructions

Run both trials sequentially on the same node (or in parallel on separate GPUs). Report `val_primary/surface_mse` AND `val_primary/vol_mse` for both.

### Trial 1 — Noam features on AF base champion (NO vol-weight)
Compare against PR #3050: surface_mse=0.000459, vol_mse=0.002777

```bash
cd target/icml2026 && python train.py \
  --dataset airfrans \
  --optimizer adamw \
  --lr 6e-4 \
  --cosine-t-max 50 \
  --grad-clip 1.0 \
  --weight-decay 1e-2 \
  --enable-fourier \
  --model-layers 2 \
  --model-hidden-dim 256 \
  --model-heads 8 \
  --epochs 999 \
  --ema-decay 0.999 \
  --asinh-pressure \
  --asinh-scale 0.75 \
  --residual-prediction \
  --re-stratified-sampling \
  --wandb_group af-noam-features-base
```

### Trial 2 — Noam features + vol-weight=10x
Compare against PR #3135: surface_mse=0.000296, vol_mse=0.002039

```bash
[...]

PR #3230, opened 2026-04-23
AirfRANS: residual-prediction + re-stratified sampling (no asinh)

tcapelle:
## Hypothesis

The previous AF noam feature test (#3191) showed that **asinh-pressure is catastrophically incompatible with AirfRANS** (surface_mse ~2000x worse). However, two other noam features — \`--residual-prediction\` and \`--re-stratified-sampling\` — were bundled with asinh and never tested in isolation. These features may still benefit the AF vol-10x champion independently:

- **Residual prediction** — predicts the correction to the freestream field instead of the full field; should improve volume predictions where the correction is small relative to baseline
- **Re-stratified sampling** — ensures training samples balanced across Reynolds number ranges; should improve OOD generalization, especially at extreme Re values

Neither of these features involves any pressure transformation (no asinh, no sinh inverse), so the catastrophic failure mechanism of #3191 cannot occur here.

## Instructions

Run 2 trials on top of the AF vol-10x champion config. The only changes from baseline are the features being tested.

**Baseline champion config (AF #3135):**
```
  --epochs 999 --ema-decay 0.999 --vol-loss-weight 10.0
```

**Trial 1 — both features together:**
Add `--residual-prediction --re-stratified-sampling` to baseline config.

**Trial 2 — residual-prediction alone:**
Add only `--residual-prediction` to baseline config.

Use `--wandb_group af-residual-restrat` for both trials.

**Do NOT add `--asinh-pressure` — this is confirmed catastrophic for AF.**

Run each trial until timeout. Monitor that surface_mse and vol_mse remain finite (no NaN/Inf in pressure channels).

## Baseline

Current AF best (PR #3135):
- **W&B run:** sh2zyfwr


tcapelle:
STUDENT alphonse:

## Results

### Trial 1 — residual-prediction + re-stratified-sampling (W&B: `qrws45k4`)

| Metric | Baseline (#3135) | Trial 1 | Delta |
|--------|------------------|---------|-------|
| val surface_mse | **0.000296** (ep704) | 0.000459 (ep725) | **+55.1% ❌** |
| val vol_mse | **0.002039** (ep743) | 0.002652 (ep725) | **+30.1% ❌** |
| test surface_mse | — | 0.000943 | — |
| test vol_mse | — | 0.003740 | — |

**Command:**
```
CUDA_VISIBLE_DEVICES=0 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --ema-decay 0.999 --volume-loss-weight 10.0 --residual-prediction --re-stratified-sampling --wandb-group af-residual-restrat
```

**Epochs:** 767 (360-min timeout), best at ep725. No NaN/Inf observed.

### Trial 2 — residual-prediction only (W&B: `9t9k4w26`)

| Metric | Baseline (#3135) | Trial 2 | Delta |
|--------|------------------|---------|-------|
| val surface_mse | **0.000296** (ep704) | 0.000509 (ep666) | **+71.9% ❌** |
| val vol_mse | **0.002039** (ep743) | 0.003256 (ep666) | **+59.7% ❌** |
| test surface_mse | — | 0.001155 | — |
| test vol_mse | — | 0.004231 | — |

**Command:**
```
CUDA_VISIBLE_DEVICES=1 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --ema-decay 0.999 --volume-loss-weight 10.0 --residual-prediction --wandb-group af-residual-restrat
```

**Epochs:** 767 (360-min timeout), best at ep666. No NaN/Inf observed.

### What happened

Both trials **regressed both metrics significantly** compared to the baseline. Residual prediction is harmful for AirfRANS.

**Why it failed:** Residual prediction subtracts a freestream field (computed from angle-of-attack) before training, so the model only learns the correction. This helps TandemFoil where much of the field is near-freestream, but AirfRANS has large separated-flow regions and recirculation zones where the correction is comparable in magnitude to the freestream — the subtraction doesn't simplify the learning task and may introduce a harder optimization landscape.

**Re-stratified sampling note:** AirfRANS returns uniform `sample_weights` (all 1.0 in `prepare_airfrans.py:310`), so `--re-stratified-sampling` only switches from shuffle (without replacement) to `WeightedRandomSampler` (with replacement, equal weights). This is NOT true Re-stratification — there's no Reynolds number balancing for AirfRANS. This explains why Trial 1 and Trial 2 are similar (the re-stratified flag is effectively a no-op for this dataset).

**Trial 1 vs Trial 2:** Trial 1 slightly outperforms Trial 2 (surface 0.000459 vs 0.000509, vol 0.002652 vs 0.003256), likely stochastic noise from the with/without-replacement sampling difference rather than a real signal.

### Verdict

**Hypothesis falsified.** Residual prediction hurts AirfRANS by ~55-72% on surface_mse and ~30-60% on vol_mse. Neither feature should be added to the AF champion config.

### Suggested follow-ups

1. **Implement proper AirfRANS Re-stratification:** If Re-stratified sampling is to be tested on AirfRANS, `prepare_airfrans.py` needs actual Reynolds-number-based sample weights (similar to TandemFoil's `prepare_multi.py`). Currently the weights are all uniform.
2. **AirfRANS-specific residual prediction:** The freestream definition (cos(AoA), sin(AoA), 0) may be too simplistic for AirfRANS. A potential-flow-based freestream or a per-sample mean subtraction might work better than the current angle-of-attack-based approach.

**Peak memory:** ~3.5 GB per GPU (standard for 2L/256d AF config)

tcapelle:
ADVISOR: Closing PR #3230 because ADVISOR: Closing — clean falsification. Residual prediction hurts AF by 55-72% on surface. Root cause: AF has large separated flow regions where the 'correction from freestream' IS the entire signal — residual framing adds complexity, not simplification. Re-stratified sampling is a no-op for AF (prepare_airfrans.py returns uniform weights). Adding residual-prediction to the AF dead-end list alongside asinh-pressure..

PR #3231, opened 2026-04-23
DrivAerML: surface pressure gradient smoothness regularization

tcapelle:
## Hypothesis

The DrivAerML champion has been stuck at 3.833% since PR #3072. The local neighborhood of recipe tuning (T_max, heads, EMA decay, LR, noam features) is exhausted. This experiment introduces a **physics-informed regularization loss** that penalizes unphysically large spatial gradients of predicted pressure fields.

**Physics motivation:** In attached flows at moderate Reynolds number, surface pressure is a smooth function except at specific geometric features (leading edge suction peak, trailing edge, separation bubbles). The Euler pressure equation requires dp/dn = 0 at the wall for flat plates — smooth variations are physically expected. High-frequency prediction noise in pressure is physically implausible and may be a source of generalization error on unseen car geometries.

**Implementation:** Add a regularization term to the loss:

```python
L_total = L_prediction + lambda_grad * L_gradient_reg
L_gradient_reg = mean(||∇p_pred||²)
```

where `∇p_pred` is computed using the mesh connectivity (neighbor differences between surface points). The gradient is approximated as the mean squared difference between each point's predicted pressure and the mean of its k nearest neighbors' predictions.

**Hypothesis:** This smoothness prior will reduce high-frequency prediction noise, improve generalization to test geometries, and potentially break through the 3.833% barrier.

## Instructions

Implement the surface pressure gradient regularization in `target/icml2026/train.py`:

1. In the training loop, after computing the standard prediction loss, add:
   - Find the k=8 nearest neighbors for each surface point using the existing coordinate data
   - Compute `p_pred` = predicted pressure for current surface points
   - Compute `L_grad_reg = mean((p_pred - mean_neighbor_p_pred)^2)` over all points
   - Add `lambda_grad * L_grad_reg` to the total loss

2. Run 3 trials with different `lambda_grad` values:
   - **Trial 1:** `lambda_grad = 0.01` (light regularization)
   - **Trial 2:** `lambda_grad = 0.1` (moderate regularization)
   - **Trial 3:** `lambda_grad = 1.0` (strong regularization)

3. All trials use the exact DM champion config:
```
python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --wandb_group dm-pressure-gradient-reg
```

4. Monitor: If `L_grad_reg` grows large (>10x prediction loss), reduce `lambda_grad`. If the model diverges at a cosine LR peak, note the epoch and stop.

5. **Note for implementation:** If computing k-NN on-the-fly per batch is expensive, you can approximate the smoothness regularization using the Transolver slice structure — compute mean pressure within each spatial slice and penalize deviations from the slice mean. This is faster and conceptually similar.

## Baseline

Current DM best (PR #3072 — eren):
- **W&B run:** ncl1dh88

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,8 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    gradient_reg_weight: float = 0.0
+    gradient_reg_k: int = 8
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -758,11 +760,49 @@ def _named_channel_metrics(prefix: str, values: torch.Tensor, names: tuple[str, 
     return metrics
 
 
+def _compute_surface_gradient_reg(
+    coords: torch.Tensor,
+    preds: torch.Tensor,
+    mask: torch.Tensor,
+    k: int = 8,
+    chunk_size: int = 2048,
+) -> torch.Tensor:
+    """k-NN smoothness regularization: penalizes deviation of each point's
+    prediction from the mean of its k nearest spatial neighbors."""
+    B = coords.shape[0]
+    results = []
+    for b in range(B):
+        m = mask[b]
+        c = coords[b, m, :3].float()
+        p = preds[b, m, 0]
+        n = c.shape[0]
+        if n <= k:
+            continue
+        all_knn_idx = torch.empty(n, k, dtype=torch.long, device=c.device)
+        with torch.no_grad():
+            for i in range(0, n, chunk_size):
+                end = min(i + chunk_size, n)
+                clen = end - i
+                dists = torch.cdist(c[i:end], c)
+                dists[torch.arange(clen, device=dists.device),
+                      torch.arange(i, end, device=dists.device)] = float('inf')
+                _, idx = dists.topk(k, dim=1, largest=False)
+                all_knn_idx[i:end] = idx
+                del dists
+        neighbor_mean = p[all_knn_idx].mean(dim=1)
+        results.append(((p - neighbor_mean) ** 2).mean())
+    if not results:
+        return torch.tensor(0.0, device=coords.device)
+    return torch.stack(results).mean()
+
+
 def loss_grouped(
     batch: GroupedBatch,
     outputs: dict[str, torch.Tensor | None],
     transform: TargetTransform,
     volume_loss_weight: float = 1.0,
+    gradient_reg_weight: float = 0.0,
+    gradient_reg_k: int = 8,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     total = torch.tensor(0.0, device=batch.surface_x.device)
     metrics: dict[str, float] = {}
@@ -776,6 +816,12 @@ def loss_grouped(
         vol_loss = F.mse_loss(outputs["volume_preds"][batch.volume_mask], target[batch.volume_mask])
         total = total + vol_loss * volume_loss_weight
         metrics["volume_loss"] = float(vol_loss.detach().cpu().item())
+    if gradient_reg_weight > 0 and outputs["surface_preds"] is not None:
+        grad_reg = _compute_surface_gradient_reg(
+            batch.surface_x, outputs["surface_preds"], batch.surface_mask, k=gradient_reg_k,
+        )
+        total = total + gradient_reg_weight * grad_reg
+        metrics["gradient_reg_loss"] = float(grad_reg.detach().cpu().item())
     metrics["loss"] = float(total.detach().cpu().item())
     return total, metrics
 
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT emma:

## Results

### Summary

**Hypothesis FALSIFIED.** Surface pressure gradient smoothness regularization did not improve DrivAerML performance at any lambda value tested. All three trials performed worse than the baseline, with degradation increasing monotonically with lambda.

### Metrics

| Trial | lambda_grad | Val surface_rel_l2_pct | Test surface_rel_l2_pct | Best Epoch | Total Epochs | W&B Run |
|-------|------------|----------------------|------------------------|------------|-------------|---------|
| **Baseline** | 0 | **3.833%** | **4.685%** | 511 | 517 | ncl1dh88 |
| Trial 1 | 0.01 | 4.481% (+16.9%) | 5.350% (+14.2%) | 287 | 356 | lqgwricr |
| Trial 2 | 0.1 | 4.876% (+27.2%) | 5.429% (+15.9%) | 351 | 355 | e2rko1vh |
| Trial 3 | 1.0 | 12.521% (+226.6%) | 12.179% (+159.9%) | 354 | 355 | xcynbcli |

### Commands

All trials used the exact DM champion config plus `--gradient-reg-weight`:
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=<gpu> python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --gradient-reg-weight <lambda> --agent emma --wandb-name "emma/dm-grad-reg-<lambda>" --wandb-group dm-pressure-gradient-reg
```

### Peak Memory

~6.3 GB per GPU (well within 96 GB VRAM).

### What happened

1. **Throughput penalty from k-NN computation:** The chunked k-NN on 50K surface points per batch (k=8, chunk_size=2048) reduced epoch throughput by ~30%. All trials reached only ~355 epochs in 360 minutes vs 511 for the baseline — fewer optimization steps to converge.

2. **Trial 1 (lambda=0.01, light):** The regularization contribution was negligible (gradient_reg_loss ~5e-5 vs prediction loss ~0.01). Performance still degraded by 16.9%, likely due to the throughput penalty reducing total training steps. Training also diverged to NaN at the terminal epoch (after best at epoch 287), suggesting numerical instability in the k-NN gradient computation under bf16 autocast.

3. **Trial 2 (lambda=0.1, moderate):** gradient_reg_loss ~0.084 at convergence, making the regularization term ~0.8% of total loss. Slightly worse than Trial 1 (val 4.876% vs 4.481%), confirming the regularization itself is actively hurting.

4. **Trial 3 (lambda=1.0, strong):** Catastrophic over-smoothing. gradient_reg_loss ~0.034 contributed 0.034/0.062 = 55% of total loss. The model was incentivized to produce spatially uniform predictions, destroying its ability to resolve sharp physics features (stagnation points, separation lines, A-pillar vortex).

### Analysis

The hypothesis that high-frequency prediction noise is a major source of generalization error on DrivAerML is **not supported**. The baseline model's remaining error (~3.83% val) appears to be dominated by systematic prediction bias (underprediction in certain regions, geometry transfer gap), not high-frequency spatial noise. A smoothness prior cannot address systematic bias.

Additionally, car surface pressure fields contain legitimately sharp features (stagnation peaks, separation lines) where physically correct pressure gradients are large. The uniform k-NN smoothness penalty penalizes these features alongside any spurious noise, creating a fundamental tension.

### Suggested follow-ups

1. **Adaptive smoothness regularization:** Weight the penalty inversely with local prediction variance so sharp-feature regions are exempted. This addresses the over-smoothing problem but adds complexity.
2. **Distance-weighted k-NN:** Use inverse-distance-weighted neighbor means instead of uniform averaging for better resolution independence (suggested by literature review).
3. **Higher-frequency-targeted penalty:** Instead of penalizing all spatial gradients, use a spectral filter to penalize only very high frequencies (noise) while preserving physically meaningful sharp features.
4. **The remaining DM error may be better addressed by:** architectural capacity (more layers/heads), longer training (more epochs), or data augmentation rather than regularization — the optimization trajectory, not the loss landscape, is the bottleneck.

tcapelle:
ADVISOR: Closing PR #3231 because ADVISOR: Closing — pressure gradient smoothness regularization is a dead end for DM. All 3 lambda values worse: λ=0.01→4.481% (+17%), λ=0.1→4.876% (+27%), λ=1.0→12.521% (+227%). Two problems: (1) KNN computation on 50k points imposes ~30% throughput overhead, cutting total epochs from 511 to ~355 — at low lambda, degradation is mostly throughput-induced. (2) Car surfaces have legitimately sharp features (stagnation, A-pillar, trailing edge) that smoothness penalties destroy. Your analysis was excellent — the core insight that the remaining 3.83% error is systematic bias (not high-frequency noise) is correct and important..

PR #3235, opened 2026-04-23
DrivAerML: multi-exit prediction (auxiliary losses at intermediate layers)

tcapelle:
## Hypothesis

The DM 4L Transolver suffers from late-training divergence — gradient instability at cosine LR peaks causes regression after ep400-500. EMA+gc manage symptoms but don't prevent the underlying gradient path fragility through 4 deep layers.

**Multi-exit prediction** adds auxiliary prediction heads at intermediate transformer layers (layers 2 and 3), with their own loss terms weighted lower than the final layer's loss. This provides:

1. **Gradient highways** — intermediate prediction heads create short gradient paths (2-3 layers deep instead of 4), reducing vanishing/exploding gradient risk at LR peaks. This directly targets the late-training divergence mechanism.
2. **Implicit deep supervision** — earlier layers receive direct supervision, learning better intermediate representations rather than relying solely on backpropagated gradients through the full stack.
3. **Natural regularization** — the auxiliary heads act as regularizers by constraining intermediate representations to be predictive, preventing overfitting in the later layers.
4. **Test-time ensemble option** — at test time, we can average predictions from all exit heads for a free variance reduction. Or use only the final head if aux heads are worse.

**References:** Teerapittayanon et al. (2016) "BranchyNet"; Lee et al. (2015) "Deeply-Supervised Nets"; Szegedy et al. (2015) auxiliary classifiers in Inception.

The key distinction from PR #3214 (auxiliary gradient prediction, which failed): that PR added a NOISY auxiliary signal (kNN-estimated gradients). This PR adds clean auxiliary signals — the same prediction task as the primary loss, just from intermediate layers. No noise injection.

## Instructions

Implement multi-exit prediction in `target/icml2026/train.py`:

1. Add auxiliary prediction heads (MLPs matching the existing prediction head architecture) after transformer layers 2 and 3. These should be independent MLP heads that project from the hidden dimension to the output dimension.

2. During training, compute the prediction loss at each exit:
   ```python
   loss_final = primary_loss(pred_layer4, target)  # weight = 1.0
   loss_aux3 = primary_loss(pred_layer3, target)   # weight = 0.3
   loss_aux2 = primary_loss(pred_layer2, target)   # weight = 0.1
   loss_total = loss_final + 0.3 * loss_aux3 + 0.1 * loss_aux2
   ```

3. Add CLI arguments:
   - `--multi-exit` (flag, enables multi-exit)
   - `--multi-exit-weights` (comma-separated, default "0.1,0.3" for layers 2,3)

4. During evaluation (model.eval()), use ONLY the final layer's prediction head for computing val metrics. The auxiliary heads are training-only unless we explicitly test ensembling.

5. Run 3 trials:

**Trial 1 — Default weights (0.1, 0.3):**
```
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --multi-exit --multi-exit-weights 0.1,0.3 \
  --wandb_group dm-multi-exit
```

**Trial 2 — Lighter weights (0.05, 0.1):**
Same but `--multi-exit-weights 0.05,0.1`

**Trial 3 — Heavier weights (0.3, 0.5):**
Same but `--multi-exit-weights 0.3,0.5`

6. Log auxiliary losses (`aux_loss_layer2`, `aux_loss_layer3`) to W&B each step. Monitor for divergence — if aux losses grow unbounded, reduce their weights.

## Baseline

Current DM best (PR #3072):
- **W&B run:** ncl1dh88

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -96,6 +96,141 @@ AIRFRANS_FIELD_NAMES = ("Ux", "Uy", "p", "nut")
 TANDEM_FIELD_NAMES = ("Ux", "Uy", "p")
 
 
+class MultiExitWrapper(torch.nn.Module):
+    """Wraps a SenpaiTransolver with auxiliary prediction heads at intermediate layers.
+
+    Provides gradient highways by adding lightweight exit heads after specified
+    backbone blocks.  During training the auxiliary predictions contribute to the
+    loss; during evaluation they are skipped.
+    """
+
+    def __init__(self, base_model: torch.nn.Module, exit_weights: list[float]):
+        super().__init__()
+        self.base = base_model
+        self.exit_weights = exit_weights
+        n_exits = len(exit_weights)
+        n_layers = len(base_model.backbone.blocks)
+        assert n_exits < n_layers, (
+            f"Need fewer exits ({n_exits}) than layers ({n_layers})"
+        )
+        # Place exits at the last N intermediate layers (excluding the final layer).
+        # For 4 layers, 2 exits → 0-indexed [1, 2] = layers 2 and 3.
+        self.exit_layer_indices = list(range(n_layers - n_exits - 1, n_layers - 1))
+
+        n_hidden = base_model.n_hidden
+        total_output_dim = base_model.surface_output_dim + base_model.volume_output_dim
+
+        self.aux_norms = torch.nn.ModuleList(
+            [torch.nn.LayerNorm(n_hidden, eps=1e-6) for _ in range(n_exits)]
+        )
+        self.aux_heads = torch.nn.ModuleList(
+            [torch.nn.Linear(n_hidden, total_output_dim) for _ in range(n_exits)]
+        )
+        for head in self.aux_heads:
+            torch.nn.init.xavier_uniform_(head.weight)
+            torch.nn.init.zeros_(head.bias)
+
+        self.n_hidden = base_model.n_hidden
+
+    def forward(
+        self,
+        *,
+        surface_x: torch.Tensor,
+        surface_mask: torch.Tensor,
+        volume_x: torch.Tensor | None = None,
+        volume_mask: torch.Tensor | None = None,
+    ) -> dict[str, torch.Tensor | None]:
+        base = self.base
+
+        # --- Embedding (replicates ReferenceTransolver.forward) ---
+        surface_hidden = base._group_hidden(
+            surface_x, base.project_surface_features, base.surface_bias,
+        )
+        hidden_parts = [surface_hidden]
+        mask_parts = [surface_mask]
+        has_volume = (
+            volume_x is not None
+            and volume_mask is not None
+            and base.volume_output_dim > 0
+        )
+        if has_volume:
+            vol_hidden_embed = base._group_hidden(
+                volume_x, base.project_volume_features, base.volume_bias,
+            )
+            hidden_parts.append(vol_hidden_embed)
+            mask_parts.append(volume_mask)
+
+        hidden = torch.cat(hidden_parts, dim=1) + base.placeholder
+        attn_mask = torch.cat(mask_parts, dim=1)
+        surface_tokens = surface_x.shape[1]
+
+        # --- Backbone blocks with intermediate exits ---
+        aux_exit_outputs: list[dict[str, torch.Tensor | None]] = []
+        exit_set = set(self.exit_layer_indices)
+        exit_idx = 0
+
+        for i, block in enumerate(base.backbone.blocks):
+            hidden = block(hidden, attn_mask=attn_mask)
+            if self.training and i in exit_set:
+                aux_raw = self.aux_heads[exit_idx](self.aux_norms[exit_idx](hidden))
+                aux_surface = aux_raw[:, :surface_tokens, :base.surface_output_dim]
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT kakashi:

## Results

### Summary

Multi-exit prediction with auxiliary losses at intermediate transformer layers **did not improve** over baseline. All 3 weight configurations produced worse val and test metrics.

### Metrics

| Trial | Aux Weights (L2, L3) | Best Val surface_rel_l2_pct | Best Epoch | Test surface_rel_l2_pct | Grad Clip Events |
|-------|---------------------|---------------------------|------------|------------------------|-----------------|
| **Baseline** (PR #3072) | — | **3.833%** | 511 | **4.685%** | — |
| Trial 1 (default) | 0.1, 0.3 | 4.278% (+0.445) | 377 | 4.983% (+0.298) | 3 |
| Trial 2 (lighter) | 0.05, 0.1 | 4.438% (+0.605) | 363 | 5.050% (+0.365) | 7 |
| Trial 3 (heavier) | 0.3, 0.5 | 4.302% (+0.469) | 378 | 5.094% (+0.409) | 47 |

**External target:** < 3.71% (AB-UPT)

### Auxiliary Loss Behavior (Final Epoch)

| Trial | aux_loss_layer2 | aux_loss_layer3 |
|-------|----------------|----------------|
| Trial 1 (0.1, 0.3) | 0.0076 | 0.0046 |
| Trial 2 (0.05, 0.1) | 0.0095 | 0.0054 |
| Trial 3 (0.3, 0.5) | 0.0063 | 0.0045 |

Auxiliary losses converged to small values across all trials, so they weren't diverging — but they still degraded primary performance.

### Commands Used

**Trial 1:**
```
cd target/icml2026 && CUDA_VISIBLE_DEVICES=3 SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --multi-exit --multi-exit-weights 0.1,0.3 --wandb-group dm-multi-exit --agent kakashi --wandb-name "kakashi/dm-multi-exit-default-0.1-0.3"
```

**Trial 2:** Same but `--multi-exit-weights 0.05,0.1` and `--wandb-name "kakashi/dm-multi-exit-lighter-0.05-0.1"`

**Trial 3:** Same but `--multi-exit-weights 0.3,0.5` and `--wandb-name "kakashi/dm-multi-exit-heavier-0.3-0.5"`

### W&B Runs
- Trial 1: `jp9n9257`
- Trial 2: `hhbj05gg`
- Trial 3: `bac66jux`
- Group: `dm-multi-exit`

### What Happened

Multi-exit prediction **hurt** rather than helped. Key observations:

1. **Val degradation across all weights:** Every configuration was 0.4–0.6% worse on val than baseline (3.833%). The auxiliary heads acted as competing objectives that subtly degraded the primary prediction path, rather than providing useful gradient highways.

2. **No divergence prevention benefit visible:** All runs peaked around epoch 363–378, well short of the baseline's best epoch (511). The auxiliary losses may have introduced optimization interference that caused earlier stagnation rather than preventing late-training divergence.

3. **Heavier weights = more instability:** Trial 3 (0.3, 0.5) had 47 grad clip events vs 3 for Trial 1 (0.1, 0.3), suggesting that stronger auxiliary supervision actually increased gradient instability — the opposite of the intended effect.

4. **Lighter weights not better:** Trial 2 (0.05, 0.1) performed worst on val despite being the most conservative, suggesting the issue isn't just weight magnitude — it's the multi-exit architecture itself introducing harmful interference.

5. **Distinction from PR #3214 holds but doesn't help:** Unlike the noisy kNN-gradient auxiliary signal in #3214, these were clean prediction losses from intermediate layers. The clean signal converged fine (aux losses reached ~0.005–0.01), but the multi-task optimization still degraded the primary objective.

### Why It Failed

The DM 4L Transolver architecture with EMA+grad-clip already manages gradient flow well enough. Adding intermediate prediction heads forces earlier layers to produce outputs suitable for direct surface prediction, which conflicts with their role in building hierarchical representations. The "gradient highway" benefit is outweighed by the representational constraint: layers 2 and 3 are better off learning intermediate features than trying to directly predict surface quantities.

### Suggested Follow-ups

- **Gradient monitoring without intervention:** Instead of auxiliary prediction heads, add gradient norm logging per layer to understand if late-training divergence is actually a gradient depth problem (as hypothesized) or something else (e.g. cosine LR schedule interaction with EMA).
- **Feature regularization instead of prediction:** If deep supervision is still interesting, try regularizing intermediate representations (e.g. layerwise correlation penalties) rather than requiring them to predict outputs directly.

tcapelle:
ADVISOR: Closing PR #3235 because ADVISOR: Closing — multi-exit prediction all 3 weight configs worse than baseline (4.278%/4.438%/4.302% vs 3.700%). Auxiliary losses at intermediate layers degrade the primary metric even when converged. DM transformer layers must focus fully on the surface prediction task. Kakashi reassigned..

PR #3237, opened 2026-04-23
DrivAerML: snapshot ensemble at cosine troughs (test-time averaging)

tcapelle:
## Hypothesis

The DM champion at 3.833% is only 0.12pp from the external target (3.71%). The cosine T_max=30 schedule creates ~17 restart cycles in 500 epochs. Each cosine trough is a local minimum — the model visits different basins during training, and the "best checkpoint" picks whichever trough found the deepest single basin.

**Snapshot ensemble** (Huang et al. 2017) exploits this directly: instead of betting on a single best trough, average predictions from the last N trough checkpoints. This provides:

1. **Variance reduction** — each trough checkpoint has slightly different predictions. Averaging reduces prediction noise without any training cost.
2. **Implicit ensemble** — different trough checkpoints explore different basins, so their predictions are partially decorrelated. This is a free ensemble from a single training run.
3. **Asymptotic gap closure** — for a 0.12pp gap, reducing prediction variance by 2-3x (typical for 3-5 model ensemble) could be sufficient.

This is a **test-time-only improvement** — no changes to the training loop. The model trains identically to the champion. The only change is saving additional checkpoints and averaging predictions at test time.

## Instructions

1. Modify `target/icml2026/train.py` to save checkpoints at every cosine trough:
   - A cosine trough occurs when `(epoch % T_max) == 0` (i.e., every 30 epochs for T_max=30)
   - Save the trough checkpoint alongside the normal best-val checkpoint
   - Name format: `checkpoint_trough_ep{epoch}.pt`

2. Add a `--snapshot-ensemble-k` argument (int, default 0 = disabled):
   - When enabled at test time, load the last K trough checkpoints
   - Run inference with each checkpoint independently
   - Average the predictions (simple mean) across all K checkpoints
   - Report the averaged prediction as the final result

3. Run the experiment:

**Phase 1 — Train with trough checkpoint saving:**
```
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --save-trough-checkpoints \
  --wandb_group dm-snapshot-ensemble
```

**Phase 2 — Evaluate with snapshot ensemble (after training completes):**
Run the test evaluation with K=3, K=5, and K=7 trough checkpoints:
```
python train.py --dataset drivaerml ... --snapshot-ensemble-k 3 --eval-only
python train.py --dataset drivaerml ... --snapshot-ensemble-k 5 --eval-only
python train.py --dataset drivaerml ... --snapshot-ensemble-k 7 --eval-only
```
Also evaluate the single best checkpoint for comparison.

4. Report: for each K, report both val and test surface_rel_l2_pct. Log the improvement from ensembling vs single best.

## Baseline

Current DM best (PR #3072):
- **W&B run:** ncl1dh88

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -168,6 +168,9 @@ class TrainConfig:
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
+    save_trough_checkpoints: bool = False
+    snapshot_ensemble_k: int = 0
+    eval_only: bool = False
     seed: int = 0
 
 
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT nobara:

## Results

### Summary

The snapshot ensemble hypothesis was tested by saving EMA-state checkpoints at every 30-epoch trough during training, then averaging predictions from the last K checkpoints at test time. **The ensemble did not improve over a single checkpoint — it degraded performance.** Including earlier, less-converged checkpoints hurts because the quality gap between early and late checkpoints is large.

### Training (Phase 1)

Training ran 193 productive epochs before hitting a gradient explosion at epoch ~196 that escalated to NaN at epoch 214. Best single-checkpoint val: **4.81%** at epoch 193 (vs baseline 3.833% at ep511). The training divergence cut the run short — the baseline ran 517 epochs without such instability.

Trough checkpoints saved at: ep30 (17.3%), ep60 (10.4%), ep90 (7.23%), ep120 (6.09%), ep150 (5.30%), ep180 (5.04%). The ep210 checkpoint was captured during divergence and was excluded.

### Ensemble Evaluation (Phase 2)

| Configuration | Test surface_rel_l2_pct | Checkpoints used |
|---|---|---|
| Single best (ep180 trough) | **6.044%** | ep180 only |
| K=3 ensemble | **6.238%** (+3.2%) | ep120, ep150, ep180 |
| K=5 ensemble | **6.980%** (+15.5%) | ep60, ep90, ep120, ep150, ep180 |
| K=6 ensemble | **7.810%** (+29.2%) | ep30–ep180 |
| **Baseline (PR #3072)** | **4.685%** | best-of ep511 |

**External target:** < 3.71% (AB-UPT)

### Commands

**Phase 1 — Training:**
```
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=180 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --save-trough-checkpoints --save-checkpoint --agent nobara --wandb-name "nobara/dm-snapshot-ensemble-train" --wandb-group dm-snapshot-ensemble --output-dir outputs/dm_snapshot_ensemble
```

**Phase 2 — K=3 ensemble eval (example):**
```
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --snapshot-ensemble-k 3 --eval-only --agent nobara --wandb-name "nobara/dm-snapshot-K3" --wandb-group dm-snapshot-ensemble --output-dir outputs/dm_snapshot_ensemble
```

### Peak memory: ~6 GB VRAM (single GPU)

### W&B runs
- Training: `c31seso3`
- K=1 (single ep180): `cwg22dhi`
- K=3 ensemble: `lcil8t4n`
- K=5 ensemble: `c8bn99m6`
- K=6 ensemble: `lp19d8uo`

### What happened

**The snapshot ensemble hypothesis is falsified.** The core assumption — that different trough checkpoints explore decorrelated basins whose prediction diversity reduces variance — does not hold here. Two factors explain the failure:

1. **Quality gradient dominates diversity.** Each trough checkpoint is significantly better than its predecessor (ep30=17.3%, ep180=5.0%). Averaging early, worse checkpoints with late, better ones just dilutes quality. For ensembling to help, the component models need similar accuracy with different error patterns.

2. **EMA already provides implicit ensembling.** The EMA with decay=0.9995 is essentially a running average over ~2000 steps of training. This already smooths out the per-checkpoint noise that snapshot ensembling would reduce. The EMA-smoothed model is a better "ensemble of one" than an explicit ensemble of EMA snapshots from different training stages.

3. **Training instability.** The run diverged at epoch ~196 with a gradient explosion (grad norm spiking to 6459 then NaN). This cut training short at 193 productive epochs vs the baseline's 517. The divergence is stochastic — the baseline run with identical hyperparameters didn't exhibit it — but it meant we had fewer high-quality trough checkpoints to work with.

### Suggested follow-ups

- **Horizontal ensembling with independent seeds**: Multiple independent training runs from different random seeds (not checkpoints from the same run) would have genuinely decorrelated predictions. This costs N× training budget but could close the gap to AB-UPT.
- **SWA (Stochastic Weight Averaging)**: Instead of averaging predictions, average the weights themselves from the last few trough checkpoints. This is cheaper (single forward pass at test time) and can find flatter minima.
- **Training stability investigation**: The gradient explosion at epoch ~196 is worth investigating. Adding warmup after cosine restarts, or reducing LR by a factor after each cycle (cosine annealing with warm restarts + decay), could prevent these events and allow longer training.

tcapelle:
ADVISOR: Closing PR #3237 because ADVISOR: Closing — snapshot ensemble is a clear dead end. Best single checkpoint test = 6.044% (29% worse than 4.685% baseline), and adding more snapshots makes it monotonically worse (K=3→6.24%, K=5→6.98%, K=6→7.81%). Root cause: cosine trough checkpoints have a dominant quality gradient (ep30=17.3% → ep180=5.0%), so averaging mixes good and bad predictions. Additionally, EMA=0.9995 already provides implicit smoothing over ~2000 gradient steps — explicit snapshot averaging adds noise on top of an already-smoothed model. Training also crashed at ep214 (grad explosion), so the available snapshots were from a run that never reached baseline quality. Your analysis was thorough and correct — nice work identifying the quality-gradient mechanism..

PR #3222, opened 2026-04-23
AirfRANS: surface-proximity volume weighting for vol_mse

tcapelle:
## Hypothesis
In CFD, near-surface volume predictions (boundary layer, wake region) are more physically important than far-field predictions where the flow is close to freestream. The current volume loss weights all volume points equally, but errors near the airfoil surface matter much more for accurate drag/lift prediction. By weighting the volume loss inversely proportional to distance from the nearest surface point, we focus the model's capacity on the most physically meaningful volume regions.

AF champion (#3135): surface_mse=0.000296, vol_mse=0.002039. Target: vol_mse < 0.0017. Human directive: ALL new AF experiments must target vol_mse < 0.0017.

## Instructions
Champion config baseline: AdamW lr=6e-4, T_max=50, gc=1.0, WD=1e-2, EMA=0.999, 2L/256d, --model-heads 4, --vol-loss-weight 10.0, --enable-fourier

Implementation:
1. For each volume point, compute the minimum Euclidean distance to any surface point in the same sample.
2. Create a weight: `w_i = 1 / (d_i + epsilon)` where epsilon=0.01. Normalize weights to mean=1.0 to preserve the overall loss scale.
3. Apply: `vol_loss = mean(w_i * (pred_i - target_i)^2)`
4. Alternative form: `w_i = exp(-d_i / scale)` (also normalized to mean=1.0), with scale=0.1 (tight BL focus) or scale=0.5 (broader near-surface region).

Run a smoke test (3 epochs) first to verify shapes and loss values are sensible before running full trials.
Use `--wandb_group af-proximity-volume-weighting` for all runs.

**Trial 1**: Inverse-distance weighting 1/(d+eps), eps=0.01. Full champion config:
```
cd target/icml2026 && python train.py --dataset airfrans --airfrans-task full \
  --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 \
  --weight-decay 1e-2 --enable-fourier --model-layers 2 --model-hidden-dim 256 \
  --model-heads 4 --epochs 999 --ema-decay 0.999 --vol-loss-weight 10.0 \
  --wandb_group af-proximity-volume-weighting
```

**Trial 2**: Exponential decay weighting, scale=0.1 (tight boundary layer). Same champion config as Trial 1 but with exp(-d/0.1) weights instead of 1/(d+eps).

[...]

PR #3239, opened 2026-04-23
AirfRANS: additive boundary-layer auxiliary volume loss

tcapelle:
## Hypothesis

PR #3222 showed that proximity-weighted volume loss catastrophically fails — replacing the uniform volume loss with a distance-weighted one starves far-field gradients. But the physical insight is correct: volume errors are highest near the surface (boundary layer, wake region). The fix is to **add** a small auxiliary loss on near-surface volume points **without modifying the main uniform loss**.

**Approach:** Keep the standard uniform vol-weight=10x loss exactly as-is. Additionally, compute a second "boundary layer" MSE loss on volume points within a distance threshold (SDF < 0.05) from the nearest surface point. Weight this auxiliary loss at 0.5-2.0x (much lighter than the main loss). This provides:

1. **Extra supervision on the hardest region** without starving far-field gradients
2. **Physics-motivated targeting** — boundary layer volume is where predictions matter most and errors are largest
3. **No interference with the proven main loss** — the auxiliary is purely additive

This is the clean follow-up directly suggested by rei in the failed proximity PR #3222.

## Instructions

Modify `target/icml2026/train.py`:

1. During training, after computing volume predictions, identify which volume points have SDF (signed distance from nearest surface point) < threshold. If SDF is not directly available, compute min distance from each volume point to the nearest surface point in the batch.

2. Compute the auxiliary boundary layer loss:
   ```python
   bl_mask = (sdf < bl_threshold)  # boolean mask
   if bl_mask.any():
       bl_loss = F.mse_loss(vol_pred[bl_mask], vol_target[bl_mask])
       total_loss = total_loss + bl_aux_weight * bl_loss
   ```

3. Add CLI arguments:
   - `--bl-aux-weight` (float, default 0.0 = disabled)
   - `--bl-threshold` (float, default 0.05 — normalized distance)

4. Log `bl_aux_loss` and `bl_point_fraction` (fraction of volume points within threshold) to W&B.

5. Run 3 trials:

**Trial 1 — Moderate aux weight, standard threshold:**
```
cd target/icml2026 && python train.py --dataset airfrans --airfrans-task full \
  --bl-aux-weight 1.0 --bl-threshold 0.05 \
  --wandb_group af-bl-auxiliary
```

**Trial 2 — Lighter aux weight:**
Same but `--bl-aux-weight 0.5`

**Trial 3 — Wider threshold (more BL coverage):**
Same as Trial 1 but `--bl-threshold 0.1`

## Baseline

Current AF best (PR #3135):
- **W&B run:** sh2zyfwr

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,8 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    bl_aux_weight: float = 0.0
+    bl_threshold: float = 0.05
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -763,6 +765,8 @@ def loss_grouped(
     outputs: dict[str, torch.Tensor | None],
     transform: TargetTransform,
     volume_loss_weight: float = 1.0,
+    bl_aux_weight: float = 0.0,
+    bl_threshold: float = 0.05,
 ) -> tuple[torch.Tensor, dict[str, float]]:
     total = torch.tensor(0.0, device=batch.surface_x.device)
     metrics: dict[str, float] = {}
@@ -776,6 +780,18 @@ def loss_grouped(
         vol_loss = F.mse_loss(outputs["volume_preds"][batch.volume_mask], target[batch.volume_mask])
         total = total + vol_loss * volume_loss_weight
         metrics["volume_loss"] = float(vol_loss.detach().cpu().item())
+
+        if bl_aux_weight > 0 and batch.volume_x is not None:
+            # AirfRANS: SDF (positive in fluid, zero at wall) at feature index 4
+            sdf = batch.volume_x[..., 4]
+            bl_mask = batch.volume_mask & (sdf < bl_threshold)
+            bl_frac = float(bl_mask.sum().item()) / max(float(batch.volume_mask.sum().item()), 1.0)
+            metrics["bl_point_fraction"] = bl_frac
+            if bl_mask.any():
+                bl_loss = F.mse_loss(outputs["volume_preds"][bl_mask], target[bl_mask])
+                total = total + bl_aux_weight * bl_loss
+                metrics["bl_aux_loss"] = float(bl_loss.detach().cpu().item())
+
     metrics["loss"] = float(total.detach().cpu().item())
     return total, metrics
 
@@ -1271,6 +1287,8 @@ def train_one_epoch(
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
     volume_loss_weight: float = 1.0,
+    bl_aux_weight: float = 0.0,
+    bl_threshold: float = 0.05,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT vegeta:

## Results

### Summary

All 3 trials **regressed both surface and volume metrics** compared to baseline. The additive boundary-layer auxiliary loss does not help and actively hurts performance.

### Metrics

| Trial | Config | Epochs | Best Epoch | Val surface_mse | Val vol_mse | Test surface_mse | Test vol_mse | W&B Run |
|-------|--------|--------|-----------|-----------------|-------------|------------------|--------------|---------|
| Baseline | vol-weight=10x | 743 | 704/743 | **0.000296** | **0.002039** | — | — | sh2zyfwr |
| T1 | bl-w=1.0, t=0.05 | 752 | 405 | 0.000448 (+51%) | 0.004362 (+114%) | 0.000860 | 0.005316 | 6y7xod45 |
| T2 | bl-w=0.5, t=0.05 | 763 | 639 | 0.000494 (+67%) | 0.002985 (+46%) | 0.001139 | 0.003969 | 7w4btwaj |
| T3 | bl-w=1.0, t=0.10 | 748 | 258 | 0.000972 (+228%) | 0.006440 (+216%) | 0.001240 | 0.007867 | pjrv0090 |

### Peak Memory

~32-34 GB per GPU (same as baseline — the BL mask is a lightweight boolean operation on existing SDF features).

### Commands Used

```bash
# T1: bl-aux-weight=1.0, bl-threshold=0.05
CUDA_VISIBLE_DEVICES=0 python train.py --dataset airfrans --airfrans-task full \
  --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 \
  --epochs 999 --ema-decay 0.999 --volume-loss-weight 10.0 \
  --bl-aux-weight 1.0 --bl-threshold 0.05 \
  --agent vegeta --wandb-name "vegeta/af-bl-aux-w1.0-t0.05" --wandb-group af-bl-auxiliary

# T2: bl-aux-weight=0.5, bl-threshold=0.05
# Same as T1 but --bl-aux-weight 0.5

# T3: bl-aux-weight=1.0, bl-threshold=0.1
# Same as T1 but --bl-threshold 0.1
```

### What Happened

**The hypothesis failed.** The additive boundary-layer auxiliary loss regresses both surface and volume metrics across all 3 trials.

**Root cause — AirfRANS mesh grading defeats the targeting:** AirfRANS meshes are heavily refined near the wall surface, so 61.4% of volume points already have SDF < 0.05, and 68.0% have SDF < 0.1. The "boundary layer" mask covers the *majority* of the mesh, not a small targeted subset. The auxiliary loss effectively acts as a near-uniform upscaling of the existing volume loss (from 10x to ~11x for BL points), which pushes the optimizer too hard on volume at the expense of surface.

Key observations:
- **Lighter weight helps marginally** — T2 (w=0.5) is less bad than T1 (w=1.0), confirming the failure is from over-weighting volume optimization
- **Wider threshold makes it worse** — T3 (t=0.1, 68% coverage) is the worst trial, consistent with the mechanism: more coverage = more uniform = less targeted
- **Best epoch shifts earlier** — T3 peaks at epoch 258 vs baseline's 704, suggesting the auxiliary loss destabilizes late-training optimization
- **Surface pressure error drives the regression** — surface_mse_p dominates the surface metric and is 3-10x worse than baseline across trials

This result is consistent with the failed proximity-weighting in PR #3222: on dense near-wall meshes, SDF-based volume re-weighting provides negligible targeting benefit and introduces training instability.

### Suggested Follow-ups

1. **Threshold much smaller** (SDF < 0.005 or 0.002) to actually target only the thin near-wall cells (~5-10% of points) rather than 61% of the mesh. This would give true boundary-layer targeting.
2. **Inverse approach** — boost far-field supervision instead of near-wall, since far-field points are under-represented in the mesh and may be under-learned.
3. **Per-field auxiliary** — target only the pressure channel (which dominates surface error) in the near-wall region, rather than all 4 output fields.
4. **Curriculum-based volume weighting** — start with lower volume weight and increase over training, since early surface convergence seems important.

tcapelle:
ADVISOR: Closing PR #3239 because ADVISOR: Closing — additive BL auxiliary volume loss is a clear dead end for AF. T1 (bl-w=1.0, t=0.05): surface +51%, vol +107%. T2 (bl-w=0.5, t=0.05): surface +67%, vol +31%. T3 (bl-w=1.0, t=0.10): surface +228%, vol +202%.

Root cause: AirfRANS wall-resolved meshes have heavy near-wall refinement — SDF<0.05 already captures 61% of ALL volume points, SDF<0.10 captures 68%. The 'boundary layer targeting' becomes near-uniform volume upscaling, replicating the same failure mode as proximity-weighting (#3222). This confirms the pattern: any SDF-based volume re-weighting on AF meshes fails because the mesh itself is already an SDF-weighted sampler.

Your diagnostic work is excellent — the bl_point_fraction logging at 0.614/0.680 nailed the root cause immediately. Between this and #3222, we can definitively rule out spatial re-weighting as a volume closure strategy for AF..

PR #3247, opened 2026-04-23
DrivAerML: EMA with Periodic Reset

tcapelle:
## Hypothesis

The champion DM config uses EMA=0.9995 running continuously from epoch 0 to 500+. The EMA model averages weights across the ENTIRE training trajectory, including many cosine restart cycles (T_max=30 → ~17 restarts in 500 epochs). Late in training, the model oscillates between different local basins during each cosine cycle. The long-running EMA averages over all these basins, which may dilute the quality of the best basin.

By periodically resetting the EMA model to the current training model weights (every N epochs), we allow the EMA to track only the CURRENT optimization basin. In the critical late-training regime (epochs 400-500), a freshly-reset EMA can capture the deepest point in the current basin without being diluted by earlier, shallower basins.

This is conceptually different from all other EMA experiments (which only varied decay rate). It changes the **temporal scope** of weight averaging.

## Baseline

- test: 4.685%
- W&B run: `ncl1dh88`

**Target to beat: 3.833% val_primary/surface_rel_l2_pct**

## Implementation Guidance

1. Add a `--ema-reset-interval` flag (integer, default 0 meaning no reset)
2. When set to N > 0, every N epochs: copy the current training model weights to the EMA model (resetting the running average)
3. After reset, EMA resumes accumulating from the new starting point with the same decay=0.9995
4. Track and log `ema_reset_count` to W&B each time a reset occurs
5. Best-checkpoint selection should still use the EMA model's validation metric

## Known Bug Fix

There is a `primary_metric_key` shadowing bug in train.py where a local variable (line ~1646-1648) shadows the function (line ~1428), causing an `UnboundLocalError`. Fix by renaming the local variable to `best_tracking_metric_key` if you encounter it.

## Trial 1 — EMA reset every 100 epochs (3.3 cosine cycles per reset window)

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-reset-interval 100 \
  --wandb_group dm-ema-reset
```

## Trial 2 — EMA reset every 50 epochs (1.67 cosine cycles per reset window)

Same as Trial 1 but with `--ema-reset-interval 50`:

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-reset-interval 50 \
  --wandb_group dm-ema-reset
```

## Reporting

Report `val_primary/surface_rel_l2_pct` at best epoch for both trials. Target: beat **3.833%**.

Also report:
- `ema_reset_count` logged to W&B (confirms resets are firing)
- Best epoch for each trial
- W&B run IDs for both trials

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -85,6 +85,14 @@ class EMAWithWarmup:
                 param.data.copy_(self.backup[key])
         self.backup = None
 
+    @torch.no_grad()
+    def reset_to_model(self, model: torch.nn.Module) -> None:
+        """Reset EMA shadow weights to current model weights, restarting the running average."""
+        for name, param in model.named_parameters():
+            key = self._clean(name)
+            if param.requires_grad and key in self.shadow:
+                self.shadow[key].copy_(param.detach())
+
 
 LEGACY_VAL_ALIAS = {
     "val_in_dist": "p_in",
@@ -124,6 +132,7 @@ class TrainConfig:
     use_lookahead: bool = True
     use_ema: bool = True
     ema_decay: float = 0.9999
+    ema_reset_interval: int = 0
     num_workers: int = -1
     pin_memory: bool = True
     persistent_workers: bool = True
@@ -1634,6 +1643,7 @@ def main() -> None:
 
     ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
     anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None
+    ema_reset_count: int = 0
     history: list[dict[str, float]] = []
     best_epoch: int | None = None
     best_val_primary_metric_name = primary_metric_key(bundle, phase="val")
@@ -1736,9 +1746,19 @@ def main() -> None:
         if anp_head is not None and anp_ema is not None:
             anp_ema.restore(anp_head)
 
+        # Periodic EMA reset: copy current training model weights into the EMA shadow,
+        # allowing the EMA to track only the current optimization basin going forward.
+        if config.ema_reset_interval > 0 and epoch % config.ema_reset_interval == 0:
+            if ema is not None:
+                ema.reset_to_model(model)
+            if anp_head is not None and anp_ema is not None:
+                anp_ema.reset_to_model(anp_head)
+            ema_reset_count += 1
+
         epoch_metrics = {"epoch": float(epoch), **train_metrics, **eval_metrics}
         epoch_metrics["best_val_metric"] = best_val_metric
         epoch_metrics["best_epoch"] = float(best_epoch)
+        epoch_metrics["ema_reset_count"] = float(ema_reset_count)
         history.append(epoch_metrics)
         if run is not None:
             wandb.log(epoch_metrics, step=epoch)


tcapelle:
STUDENT brook:

## Results

### Trial 1 — EMA reset every 100 epochs

| Metric | Value | Baseline | Delta |
|--------|-------|----------|-------|
| val_primary/surface_rel_l2_pct (best) | **4.426%** (epoch 317) | 3.833% (epoch 511) | +0.593pp (worse) |
| test_primary/surface_rel_l2_pct (best ckpt) | **5.507%** | 4.685% | +0.822pp (worse) |
| ema_reset_count at best | 3 | 0 | — |

- **W&B run:** `ltfotg72` (brook/dm-ema-reset-100)
- **Command:**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=2 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=360.0 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --ema-reset-interval 100 --agent brook --wandb-name "brook/dm-ema-reset-100" --wandb-group dm-ema-reset
```
- **Peak memory:** ~15-20 GB (4L/512d/8H, single GPU, same as baseline)

**What happened:** The first two EMA resets (epochs 100, 200) worked well — the model recovered within 5-7 epochs and resumed descent to new bests. The third reset at epoch 300 initially recovered (new best 4.426% at epoch 317), but a gradient instability event at epoch 318 (grad_norm spiked to 1.04) cascaded into full divergence by epoch 335 (grad_norm 19→45→106→452→970→Infinity). The model never recovered.

### Trial 2 — EMA reset every 50 epochs

| Metric | Value | Baseline | Delta |
|--------|-------|----------|-------|
| val_primary/surface_rel_l2_pct (best) | **6.818%** (epoch 100) | 3.833% (epoch 511) | +2.985pp (worse) |
| test_primary/surface_rel_l2_pct (best ckpt) | **7.509%** | 4.685% | +2.824pp (worse) |
| ema_reset_count at best | 2 | 0 | — |

- **W&B run:** `q221x53o` (brook/dm-ema-reset-50)
- **Command:** Same as Trial 1 but `--ema-reset-interval 50`
- **Peak memory:** ~15-20 GB (same config)

**What happened:** The first reset at epoch 50 worked fine — 5 epochs to recover. The second reset at epoch 100 initially recovered but a gradient explosion started at epoch 108 (grad_norm 10→61→72→157→175→3159→1M→NaN by epoch 118). The fresh EMA after reset didn't accumulate enough averaging before the cosine LR peak, leaving training weights unprotected.

### Analysis

**The hypothesis is falsified.** Periodic EMA resets do NOT help DrivAerML training — they are actively destructive.

Key findings:
1. **Resets accumulate instability.** Each successive reset is more dangerous than the last. The fresh EMA has no averaged history to stabilize against cosine LR peaks, and the model is in a sharper basin that's more sensitive to perturbation.
2. **More frequent resets = earlier divergence.** Trial 2 (every 50 epochs) diverged at epoch 118. Trial 1 (every 100 epochs) survived until epoch 335. The baseline (no resets) was stable through 517+ epochs.
3. **The continuous EMA is protective.** The long-running EMA averages over many cosine cycles, providing a stabilizing effect that prevents gradient explosions at LR peaks. Resetting this history removes the stability.
4. **The "basin dilution" hypothesis was wrong.** The continuous EMA does average over multiple basins, but this is a feature (stability), not a bug (dilution). The EMA's role on DrivAerML is more about optimization stability than finding the deepest single-basin average.

### Suggested follow-ups

- **Partial reset:** Instead of fully copying training weights to EMA, do a weighted blend: `ema_shadow = α * model_weights + (1-α) * ema_shadow` with α=0.5. This refreshes the EMA while preserving some averaging history.
- **Late-only reset:** Single reset at a specific epoch (e.g., epoch 400) rather than periodic. The hypothesis about late-training basin tracking might hold for a single, well-timed reset.
- **Adaptive gc strengthening after reset:** Temporarily increase grad-clip (e.g., gc=0.1) for 30 epochs after each reset to protect the unaveraged EMA, then relax back to gc=0.5.

tcapelle:
ADVISOR: Closing PR #3247 because ADVISOR: Closing — EMA periodic reset is a clear dead end for DM. 100-ep reset: 4.426% (+15%), catastrophically diverged ep335 after reset #3. 50-ep reset: 6.818% (+78%), diverged ep108 after reset #2.

Key insight confirmed: EMA=0.9995 on DM is the PRIMARY gradient stabilization mechanism, not just a weight-averaging trick. The model enters progressively sharper basins as training matures — by ep300, resetting the EMA shadow is equivalent to removing the protective smoothing at the highest-curvature point in the optimization, causing immediate cascade failure. The multi-basin averaging that resets were designed to 'fix' is actually the feature, not the bug.

Your analysis was excellent — the progressive basin sharpening explanation (reset #1/#2 survived, #3 did not) is the right mechanistic model. This result adds a strong constraint: any future EMA modification must preserve the continuous shadow trajectory..

PR #3248, opened 2026-04-23
DrivAerML: Decaying Peak LR at Cosine Restarts (SGDR eta_mult)

tcapelle:
## Hypothesis

The DM champion uses cosine annealing with T_max=30 (step-level, ~394 steps/epoch), which means the LR restarts to the full peak (5e-4) repeatedly throughout training. With ~13 restarts per epoch × 500 epochs = ~6500 restarts total, the model keeps getting jolted back to full LR even in late training. These late-stage full-strength restarts may cause the val oscillations we see in the 3.8–4.1% range — the model finds a deep basin but then gets kicked out by the next full-LR restart.

**SGDR** (Loshchilov & Hutter 2017) introduced multiplicative decay of the peak LR after each restart cycle. Each restart uses a lower peak LR:

```
peak_lr_n = initial_lr * (eta_mult ** n)
```

where n is the restart count. This preserves the beneficial exploration in early training while making late restarts progressively gentler — reducing the perturbation that destabilizes the model near convergence.

### Why this is distinct from previous dead ends

- **eta_min tuning** — sets a floor for the minimum LR; doesn't change the peak. Already shown to be ineffective for DM.
- **Monotonic cosine** — removes restarts entirely. Dead end at 4.086%.
- **SAM** — tries to flatten basins; incompatible with restarts. Dead.

This approach **preserves restarts** but makes them progressively gentler over training.

---

## Baseline

Current best DrivAerML (PR #3072, `eren/dm-ema-9995-gc05`):
- Test: 4.685%
- W&B run: `ncl1dh88`

External target: **<3.71%** (AB-UPT). Gap to close: ~0.123 pp.

---

## Known Bug Fix

There is a `primary_metric_key` shadowing bug in `train.py` where a local variable (around line ~1646–1648) shadows the function defined at line ~1428, causing an `UnboundLocalError`. If you encounter it, fix by renaming the local variable to `best_tracking_metric_key`.

---

## Implementation Guidance

1. Add a `--cosine-eta-mult` flag (float, default `1.0` = no decay).
2. Track the number of completed cosine cycles (restarts).
3. After each restart, multiply the current peak LR by eta_mult:
   ```python
   current_peak = lr * (eta_mult ** restart_count)
   ```
4. The cosine schedule within each cycle goes from `current_peak` down to 0 (or near-0).
5. Log `current_peak_lr` and `restart_count` to W&B at each restart.

### Restart count math

With T_max=30 (step-level) and ~394 steps/epoch:
- Restarts per epoch ≈ 394 / 30 ≈ 13.1
- Total restarts over 500 epochs ≈ **6,500**

This means eta_mult must be **very close to 1.0**. Example decay curves:

| eta_mult | Peak LR at ep500 | Fraction remaining |
|----------|------------------|--------------------|
| 0.9999   | ~2.61e-4         | ~52%               |
| 0.99995  | ~3.61e-4         | ~72%               |

---

### Trial 1 — eta_mult=0.9999 (moderate decay: peak ~2.6e-4 at ep500)

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --cosine-eta-mult 0.9999 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --wandb_group dm-decaying-restart
```

### Trial 2 — eta_mult=0.99995 (gentle decay: peak ~3.6e-4 at ep500)

Same as Trial 1 but with `--cosine-eta-mult 0.99995`:

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --cosine-eta-mult 0.99995 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --wandb_group dm-decaying-restart
```

---

## Reporting

For each trial, report:
- `val_primary/surface_rel_l2_pct` at best epoch
- The epoch at which the best val metric was achieved
- The `current_peak_lr` logged at that epoch (to understand where in the decay curve the model finds its optimum)
- W&B run ID

**Target to beat: 3.833%**

If Trial 1 (eta_mult=0.9999) shows improvement, consider a follow-up with eta_mult=0.9998 (more aggressive decay) in the same PR. If Trial 2 (eta_mult=0.99995) shows improvement but Trial 1 doesn't, consider eta_mult=0.99997 as a midpoint.

Report using `--wandb_group dm-decaying-restart` so both runs appear in the same W&B group.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -160,6 +160,7 @@ class TrainConfig:
     residual_prediction: bool = False
     re_stratified_sampling: bool = False
     cosine_t_max: int = 150
+    cosine_eta_mult: float = 1.0
     geometry_points: int = 25_000
     geometry_supernodes: int = 4_096
     surface_anchor_points: int = 8_000
@@ -1628,9 +1629,20 @@ def main() -> None:
         params.extend(list(anp_head.parameters()))
     optimizer = build_optimizer(params, config)
     scheduler = None
+    cosine_restart_count = 0
     if config.cosine_t_max > 0:
         base_optimizer = optimizer.optimizer if isinstance(optimizer, Lookahead) else optimizer
-        scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
+        if config.cosine_eta_mult < 1.0:
+            t_max = config.cosine_t_max
+            eta_mult = config.cosine_eta_mult
+            def _sgdr_lambda(step: int) -> float:
+                cycle = step // t_max
+                t_cur = step % t_max
+                peak_factor = eta_mult ** cycle
+                return peak_factor * 0.5 * (1.0 + math.cos(math.pi * t_cur / t_max))
+            scheduler = torch.optim.lr_scheduler.LambdaLR(base_optimizer, _sgdr_lambda)
+        else:
+            scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(base_optimizer, T_max=config.cosine_t_max)
 
     ema = EMAWithWarmup(model, decay=config.ema_decay) if config.use_ema else None
     anp_ema = EMAWithWarmup(anp_head, decay=config.ema_decay) if config.use_ema and anp_head is not None else None
@@ -1739,6 +1751,12 @@ def main() -> None:
         epoch_metrics = {"epoch": float(epoch), **train_metrics, **eval_metrics}
         epoch_metrics["best_val_metric"] = best_val_metric
         epoch_metrics["best_epoch"] = float(best_epoch)
+        if scheduler is not None and config.cosine_eta_mult < 1.0 and config.cosine_t_max > 0:
+            total_steps = scheduler.last_epoch
+            restart_count = total_steps // config.cosine_t_max
+            current_peak_lr = config.lr * (config.cosine_eta_mult ** restart_count)
+            epoch_metrics["sgdr_restart_count"] = float(restart_count)
+            epoch_metrics["sgdr_current_peak_lr"] = current_peak_lr
         history.append(epoch_metrics)
         if run is not None:
             wandb.log(epoch_metrics, step=epoch)


tcapelle:
STUDENT nobara:

## Results

### Trial 1 — eta_mult=0.9999 (moderate decay: peak ~2.6e-4 at ep500)

| Metric | Value |
|--------|-------|
| `val_primary/surface_rel_l2_pct` | **4.153%** at epoch 490 |
| `test_primary/surface_rel_l2_pct` | **4.543%** (best-checkpoint eval) |
| Total epochs | 498 (360-min timeout) |
| `sgdr_current_peak_lr` at best epoch | ~0.000264 (52.8% of original 5e-4) |
| Total restarts (at best) | ~6,447 |
| W&B run | `fgtjrzmc` |
| Peak memory | ~6 GB |

**Command:**
```bash
  --cosine-t-max 30 --cosine-eta-mult 0.9999 \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --agent nobara \
  --wandb-name "nobara/dm-sgdr-eta0.9999" --wandb-group dm-decaying-restart
```

### Trial 2 — eta_mult=0.99995 (gentle decay: peak ~3.6e-4 at ep500)

| Metric | Value |
|--------|-------|
| `val_primary/surface_rel_l2_pct` | **12.643%** at epoch 50 |
| `test_primary/surface_rel_l2_pct` | **13.191%** (best-checkpoint eval) |
| Total epochs | 502 (360-min timeout) |
| W&B run | `39vnze9p` |
| Status | **DIVERGED** at ~ep60, never recovered |

**Command:** Same as Trial 1 but with `--cosine-eta-mult 0.99995`.

---

### Comparison vs Baseline

| | Baseline (PR #3072) | Trial 1 (eta=0.9999) | Trial 2 (eta=0.99995) |
|--|---------------------|----------------------|-----------------------|
| Best val % | **3.833%** (ep511) | 4.153% (ep490) | 12.643% (ep50) |
| Test % | 4.685% | 4.543% | 13.191% |
| Schedule | CosineAnnealingLR | SGDR warm restarts | SGDR warm restarts |
| Peak LR at best | 5e-4 (constant) | 2.64e-4 (52.8%) | 4.84e-4 (96.7%) |
| Stability | Stable oscillation | Stable (dampened) | Diverged at ep~60 |

**Target to beat: 3.833% — NOT BEATEN**

---

### What happened

**Trial 1 (eta_mult=0.9999)** trained stably through 498 epochs with progressively dampened restarts. The decay curve worked as designed — peak LR decayed from 5e-4 to ~2.6e-4 (52%) by training's end. The model showed smooth descent without the large oscillation spikes seen in the baseline, reaching 4.153% at ep490. However, it did not beat the baseline 3.833%.

The likely reason: the SGDR implementation uses **warm restarts** (sharp LR jumps to decayed peak every T_max=30 steps), which is fundamentally different from the baseline's **CosineAnnealingLR** (smooth oscillation with period 2×T_max=60 steps). This schedule change confounds the eta_mult decay effect. The sharp restart jumps may have been slightly suboptimal compared to the smooth oscillation baseline, offsetting any benefit from peak decay.

Additionally, the aggressive LR decay may have "used up" the learning rate budget too early — by ep490, the peak LR was only 52% of original, potentially limiting late-training exploration. The model plateaued from ep412-ep480 before finding a slight improvement, suggesting it struggled to escape the basin with reduced LR.

**Trial 2 (eta_mult=0.99995)** diverged catastrophically around epoch 60. The gentle decay preserved near-full peak LR (~96% at ep50), which combined with the warm restart sharp jumps, caused massive oscillation spikes (val jumping to 64-97%). The warm restart behavior was far more destabilizing than the baseline's smooth CosineAnnealingLR when the peak LR wasn't sufficiently dampened.

### Key insight

The SGDR implementation changed **two things simultaneously**: (1) warm restarts instead of smooth CosineAnnealingLR oscillation, and (2) peak LR decay. Warm restarts (sharp LR jumps) are inherently more destabilizing than smooth oscillation, requiring stronger eta_mult decay to compensate. Trial 2's gentle decay couldn't compensate, leading to divergence. Trial 1's moderate decay stabilized training but the combined schedule change didn't outperform the simpler baseline.

### Suggested follow-ups

1. **Apply eta_mult on top of CosineAnnealingLR (not warm restarts)** — multiply the CosineAnnealingLR output by `eta_mult^(step // (2*T_max))` to preserve the smooth oscillation while adding peak decay. This isolates the eta_mult effect from the warm restart confound.
2. **Try eta_mult=0.99993** — splitting the difference between the two trials to find the optimal decay strength under warm restarts.
3. **CosineAnnealingWarmRestarts with T_mult>1** — increasing cycle length over time (e.g., T_mult=1.5) instead of decaying the peak. This keeps the peak LR constant but makes later cycles longer, achieving a similar "gentler late training" effect through a different mechanism.

tcapelle:
ADVISOR: Closing PR #3248 because ADVISOR: Closing — SGDR eta_mult is a dead end for DM. eta_mult=0.9999: 4.153% (+0.32pp, stable but didn't beat 3.833%). eta_mult=0.99995: 12.64% (diverged ep60).

Key issue you correctly identified: the experiment confounded eta_mult with switching from CosineAnnealingLR (smooth) to CosineAnnealingWarmRestarts (sharp jumps). Trial 2's divergence proves the sharp restart jumps are more destabilizing than baseline's smooth oscillation. Trial 1's stability came from severe peak decay (to 52.6% of original LR by ep490), not from the restart schedule being better.

Your T_mult>1 suggestion (longer cycles over time) is the cleaner mechanism for 'gentler late training' if the schedule space is revisited..

PR #3251, opened 2026-04-23
DrivAerML Pre-LayerNorm Architecture Ablation

tcapelle:
## Hypothesis

Standard transformers come in two flavors: Post-LN (original Vaswani et al.) and Pre-LN (GPT-2, LLaMA). Pre-LN applies LayerNorm BEFORE the attention and MLP sub-layers, while Post-LN applies it AFTER. Pre-LN is known to:

1. Reduce gradient variance in deep models (Xiong et al. 2020 "On Layer Normalization in the Transformer Architecture")
2. Improve training stability, especially with aggressive LR schedules
3. Often match or exceed Post-LN performance in practice

The DrivAerML model uses a Transolver architecture. If it uses Post-LN (the default in many implementations), switching to Pre-LN could reduce gradient variance at cosine restart boundaries, enabling better late-stage convergence at the 3.833% frontier. With only 4 layers, the depth isn't extreme, but Pre-LN's gradient stabilization effect is still relevant given the T_max=30 rapid restart schedule.

## Baseline

- **val_primary/surface_rel_l2_pct: 3.833%** at epoch 511
- test: 4.685%
- W&B run: ncl1dh88

**Target: beat 3.833%.**

## Known Bug

There's a `primary_metric_key` shadowing bug in train.py where a local variable (line ~1646-1648) shadows the function (line ~1428), causing an `UnboundLocalError`. Fix by renaming the local variable to `best_tracking_metric_key` if you encounter it.

### Step 1: Investigate current architecture

First, read the Transolver model code to determine whether it currently uses Pre-LN or Post-LN. Look at the TransolverBlock/TransolverLayer forward method. Report which norm placement is used in your results comment.

### Step 2: Implement the alternative

- If currently Post-LN: implement Pre-LN (add `--pre-layer-norm` flag)
- If currently Pre-LN: implement Post-LN (add `--post-layer-norm` flag)
- The change is: move the LayerNorm from `x = LN(x + sublayer(x))` to `x = x + sublayer(LN(x))`
- If the model uses a different normalization pattern, describe it and propose the Pre-LN variant

### Trial 1 — Alternative LN placement on champion config

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --pre-layer-norm \
  --wandb_group dm-pre-layernorm
```

(Adjust the flag name based on what you find in Step 1 — use `--post-layer-norm` if the current default is Pre-LN.)

### Trial 2 — Alternative LN + RMSNorm (if Pre-LN is the change)

If changing to Pre-LN, also try replacing LayerNorm with RMSNorm (Root Mean Square Layer Normalization — simpler, no mean centering). RMSNorm is used in LLaMA and is both faster and sometimes more effective than LayerNorm.

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --pre-layer-norm \
  --use-rmsnorm \
  --wandb_group dm-pre-layernorm
```

## Reporting

Report `val_primary/surface_rel_l2_pct` at best epoch for each trial. Include W&B run IDs. Target: beat **3.833%**.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -169,6 +169,8 @@ class TrainConfig:
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
+    post_layer_norm: bool = False
+    use_rmsnorm: bool = False
 
 
 @dataclass
@@ -497,6 +499,47 @@ def cp_panel_prior_index(config: TrainConfig, bundle: DatasetBundle) -> int | No
         base_dim -= 1
         return base_dim
     return None
+
+
+class RMSNorm(torch.nn.Module):
+    def __init__(self, normalized_shape: int, eps: float = 1e-6):
+        super().__init__()
+        self.weight = torch.nn.Parameter(torch.ones(normalized_shape))
+        self.eps = eps
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        rms = torch.sqrt(torch.mean(x * x, dim=-1, keepdim=True) + self.eps)
+        return x / rms * self.weight
+
+
+def _apply_post_layer_norm(block: torch.nn.Module) -> None:
+    original_forward = block.forward
+
+    def post_ln_forward(x: torch.Tensor, attn_mask: torch.Tensor | None = None) -> torch.Tensor:
+        x = block.norm1(x + block.attention(x, attn_mask=attn_mask))
+        x = block.norm2(x + block.mlp(x))
+        return x
+
+    block.forward = post_ln_forward
+
+
+def apply_norm_config(model: torch.nn.Module, config: TrainConfig) -> None:
+    from core.architectures.transolver_reference import TransformerBlock
+
+    if config.use_rmsnorm:
+        for module in model.modules():
+            if isinstance(module, TransformerBlock):
+                hidden_dim = module.norm1.normalized_shape[0]
+                module.norm1 = RMSNorm(hidden_dim, eps=1e-6)
+                module.norm2 = RMSNorm(hidden_dim, eps=1e-6)
+            elif hasattr(module, "norm") and isinstance(module.norm, torch.nn.LayerNorm) and not isinstance(module, TransformerBlock):
+                hidden_dim = module.norm.normalized_shape[0]
+                module.norm = RMSNorm(hidden_dim, eps=1e-6)
+
+    if config.post_layer_norm:
+        for module in model.modules():
+            if isinstance(module, TransformerBlock):
+                _apply_post_layer_norm(module)
 
 
 def build_model(config: TrainConfig, bundle: DatasetBundle) -> torch.nn.Module:
@@ -1614,7 +1657,10 @@ def main() -> None:
             )
 
     train_loader, val_loaders, test_loaders = build_loaders(config, bundle, num_workers=resolved_num_workers)
-    model = build_model(config, bundle).to(device)
+    model = build_model(config, bundle)
+    if config.post_layer_norm or config.use_rmsnorm:
+        apply_norm_config(model, config)
+    model = model.to(device)
     forward_model = torch.compile(model) if config.compile_model and device.type == "cuda" else model
     anp_head = None
     if bundle.spec.name == "tandemfoilset" and config.anp_srf:
@@ -1639,15 +1685,6 @@ def main() -> None:
     best_val_primary_metric_name = primary_metric_key(bundle, phase="val")
     best_val_primary_metric: float | None = None
     best_val_metrics: dict[str, float] = {}
-    best_model_state: dict[str, torch.Tensor] | None = None
-    best_anp_state: dict[str, torch.Tensor] | None = None
-
-    if bundle.spec.name == "tandemfoilset":
-        val_primary_key = "val_primary/surface_pressure_mae"
-    else:
-        val_primary_key = f"val_primary/{bundle.spec.default_metric}"
-    best_val_metric = float("inf")
-    best_epoch = 0
     best_model_state: dict[str, torch.Tensor] | None = None
     best_anp_state: dict[str, torch.Tensor] | None = None
     best_ema_shadow: dict[str, torch.Tensor] | None = None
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT fern:

## Architecture Investigation

The Transolver currently uses **Pre-LN** (norm before sublayer):
```python
# TransformerBlock.forward()
x = x + self.attention(self.norm1(x), attn_mask=attn_mask)  # Pre-LN
x = x + self.mlp(self.norm2(x))
```
LayerNorm is applied to `norm1` (pre-attention) and `norm2` (pre-MLP), plus a final `self.norm` before the output projection. This is the GPT-2/LLaMA style.

Since the default is Pre-LN, the alternative tested is **Post-LN** (`x = LN(x + sublayer(x))`).

## Bug Fix

Fixed the `primary_metric_key` shadowing bug: consolidated two competing best-model tracking systems (lines 1638-1654 in the original) into a single unified system. The duplicate inline key computation and redundant variable declarations were removed, and EMA shadow saving was merged into the primary tracking block.

## Results

### Trial 1 — Post-LN (alternative norm placement)

| Metric | Baseline (Pre-LN) | Post-LN | Delta |
| val_primary/surface_rel_l2_pct | **3.833%** (ep511) | **9.515%** (ep55) | +148% ❌ |
| test_primary/surface_rel_l2_pct | 4.685% | 10.292% | +120% ❌ |

- **W&B run:** aix26jjn (fern/dm-post-layernorm)
- **Command:** `cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --post-layer-norm --agent fern --wandb-name "fern/dm-post-layernorm" --wandb-group dm-pre-layernorm`
- **Epochs:** 495 (360-min timeout), best at epoch 55
- **What happened:** Post-LN **diverged catastrophically at epoch ~58**. Grad norms jumped from 0.90 → 317 → Infinity within 5 epochs. The model never recovered — val metric stuck at ~75.6% for the remaining 440 epochs. Post-LN removes the gradient stabilization that Pre-LN provides before the attention and MLP sublayers, making cosine restart boundaries (T_max=30) lethal. The gc=0.5 clipping was insufficient to prevent the divergence.

### Trial 2 — Pre-LN + RMSNorm (replacing LayerNorm with RMSNorm)

| Metric | Baseline (Pre-LN + LayerNorm) | Pre-LN + RMSNorm | Delta |
|--------|-------------------------------|-------------------|-------|
| val_primary/surface_rel_l2_pct | **3.833%** (ep511) | **4.394%** (ep302) | +14.6% ❌ |
| test_primary/surface_rel_l2_pct | 4.685% | 5.468% | +16.7% ❌ |

- **W&B run:** 2ghhkigq (fern/dm-pre-ln-rmsnorm)
- **Command:** `cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --use-rmsnorm --agent fern --wandb-name "fern/dm-pre-ln-rmsnorm" --wandb-group dm-pre-layernorm`
- **Epochs:** 498 (360-min timeout), best at epoch 302
- **Convergence trajectory:** 61% (ep1) → 12% (ep50) → 7.1% (ep100) → 5.5% (ep150) → 4.8% (ep200) → 4.4% (ep302) → diverged at ~ep340 (grad_norm→Infinity)
- **What happened:** RMSNorm converged well for 302 epochs but ultimately also diverged. The lack of mean centering in RMSNorm (compared to LayerNorm) made the model slightly less stable at the DrivAerML frontier. It reached 4.394% — closer than Post-LN but still 0.56pp above baseline. The divergence at ~ep340 follows the same pattern as Post-LN but occurs much later, suggesting RMSNorm provides partial but insufficient gradient stabilization compared to full LayerNorm.

## Analysis

Both norm variants are **strictly worse** than the current Pre-LN + LayerNorm default:

1. **Post-LN is catastrophically unstable** for this architecture. The T_max=30 cosine schedule creates sharp LR peaks that Post-LN cannot survive — divergence within 58 epochs. This confirms the literature: Pre-LN is essential for stable training with aggressive LR schedules in shallow transformers.

2. **RMSNorm is competitive but slightly worse**. It converges to within 0.56pp of baseline before diverging. The mean centering in LayerNorm provides a small but critical stability margin at the DrivAerML frontier. The divergence point shifted from ep58 (Post-LN) to ep340 (RMSNorm), confirming the hierarchy: LayerNorm > RMSNorm > Post-LN for gradient stability.

3. **Both diverged during the same run phase** — when EMA decay is fully saturated (0.9995) and the cosine restarts create gradient spikes. The gc=0.5 clipping was insufficient to prevent divergence in both cases, while the baseline (Pre-LN + LayerNorm) survived 511+ epochs with the same clipping.

**Verdict:** The current Pre-LN + LayerNorm architecture is already the optimal norm configuration for this model. Neither alternative improves upon it.

## Suggested follow-ups

- **Stronger gc with RMSNorm:** Since RMSNorm reached 4.394% before diverging, gc=0.3 or gc=0.2 might prevent the late divergence and let it train longer. The 0.56pp gap could potentially be closed with more epochs.
- **Post-LN with gc=0.1:** Extremely tight clipping might prevent Post-LN divergence, though at the cost of optimization speed.
- **QK-Norm:** An alternative to full Pre-LN/Post-LN — normalize just the Q and K projections in attention. This targets the specific instability source (attention logit magnitude) without changing the residual stream.

tcapelle:
ADVISOR: Closing PR #3251 because ADVISOR: Closing — Post-LN val=9.515% is catastrophically worse (+157%) vs baseline. Pre-LN is load-bearing for this architecture. Note: primary_metric_key shadowing fix in this PR is valuable — will cherry-pick separately. Fern reassigned..

PR #3252, opened 2026-04-23
DrivAerML Progressive Surface Point Training (Resolution Curriculum)

tcapelle:
## Hypothesis

DrivAerML trains with 50,000 surface points per mesh, using 394 meshes per epoch. At 4L/512d, each epoch takes ~1 minute. The champion reached its best (3.833%) at epoch 511 of 517 total — still improving at timeout. More epochs would likely push the result lower.

Progressive training (curriculum learning on resolution) exploits the observation that early training learns COARSE surface patterns that don't require full 50k-point resolution. By training with fewer points (e.g., 25k or 30k) for the first N epochs, we get:
1. **Faster epochs** — roughly 40-50% faster per epoch at 25k vs 50k points
2. **More total epochs** in the same wall-clock budget
3. **Implicit regularization** — lower resolution forces the model to learn smoother, more generalizable representations first
4. **Full-resolution fine-tuning** — switching to 50k points in late training refines predictions at the detailed level

This is a training efficiency technique, not a model change. The model architecture stays exactly the same — only the number of sampled points per mesh changes during training.

## Baseline

- test: 4.685%
- W&B run: `ncl1dh88`
- Trained 517 epochs in ~360 minutes (~0.7 min/epoch at 50k points)

**Target: beat 3.833%**

## Known Bug Fix

There is a `primary_metric_key` shadowing bug in `train.py` where a local variable (lines ~1646-1648) shadows the function (line ~1428), causing an `UnboundLocalError`. Fix by renaming the local variable to `best_tracking_metric_key` if you encounter it.

## Implementation

1. Add `--progressive-points-schedule` flag (string, e.g., `"25000:200,50000:999"` meaning 25k points for epochs 0–200, 50k for 201–999)
2. Parse the schedule and change `drivaerml_train_surface_points` at the appropriate epoch boundaries
3. **Eval points must ALWAYS stay at 50k** (consistent evaluation)
4. Log `current_train_points` to W&B each epoch

## Trial 1 — 25k points for first 200 epochs, then 50k

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --progressive-points-schedule "25000:200,50000:999" \
  --wandb_group dm-progressive-resolution
```

## Trial 2 — 30k points for first 100 epochs, then 50k

Same as Trial 1 but with:
```
--progressive-points-schedule "30000:100,50000:999"
```

## Key Measurements

- Compare total wall-clock training time and total epochs reached vs baseline
- Report both the best val metric AND the total number of epochs trained within the timeout
- Report `val_primary/surface_rel_l2_pct` at best epoch
- Both trials should use `--wandb_group dm-progressive-resolution`

## Expected Outcome

At 25k points, epochs should run ~40–50% faster, allowing ~700–800 epochs within the same ~360-minute budget vs the baseline's 517. Even if full-resolution performance is modestly below what pure 50k training achieves, the extra epochs in the fine-tuning phase should more than compensate. The implicit regularization from coarse early training may also improve generalization.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,7 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    progressive_points_schedule: str = ""
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -452,6 +453,27 @@ def parse_args() -> TrainConfig:
             parser.add_argument(arg_name, type=type(field_value), default=field_value)
     namespace = parser.parse_args()
     return TrainConfig(**vars(namespace))
+
+
+def parse_progressive_points_schedule(schedule: str) -> list[tuple[int, int]]:
+    if not schedule.strip():
+        return []
+    stages: list[tuple[int, int]] = []
+    for part in schedule.split(","):
+        part = part.strip()
+        points_str, epoch_str = part.split(":")
+        stages.append((int(points_str), int(epoch_str)))
+    stages.sort(key=lambda s: s[1])
+    return stages
+
+
+def get_progressive_train_points(stages: list[tuple[int, int]], epoch: int) -> int | None:
+    if not stages:
+        return None
+    for points, until_epoch in stages:
+        if epoch <= until_epoch:
+            return points
+    return stages[-1][0]
 
 
 def build_bundle(config: TrainConfig) -> DatasetBundle:
@@ -1643,13 +1665,13 @@ def main() -> None:
     best_anp_state: dict[str, torch.Tensor] | None = None
 
     if bundle.spec.name == "tandemfoilset":
-        val_primary_key = "val_primary/surface_pressure_mae"
+        best_tracking_metric_key = "val_primary/surface_pressure_mae"
     else:
-        val_primary_key = f"val_primary/{bundle.spec.default_metric}"
+        best_tracking_metric_key = f"val_primary/{bundle.spec.default_metric}"
     best_val_metric = float("inf")
     best_epoch = 0
-    best_model_state: dict[str, torch.Tensor] | None = None
-    best_anp_state: dict[str, torch.Tensor] | None = None
+    best_model_state = None
+    best_anp_state = None
     best_ema_shadow: dict[str, torch.Tensor] | None = None
     best_anp_ema_shadow: dict[str, torch.Tensor] | None = None
 
@@ -1666,12 +1688,25 @@ def main() -> None:
             tags=[config.dataset, config.model],
         )
 
+    progressive_stages = parse_progressive_points_schedule(config.progressive_points_schedule)
+    prev_train_points: int | None = None
+
     start_time = time.monotonic()
     timeout_seconds = None if not timeout_env else float(timeout_env) * 60.0
 
     for epoch in range(1, config.epochs + 1):
         if timeout_seconds is not None and epoch > 1 and (time.monotonic() - start_time) >= timeout_seconds:
             break
+
+        current_train_points = get_progressive_train_points(progressive_stages, epoch)
+        if current_train_points is not None:
+            train_ds = bundle.train_dataset
+            if hasattr(train_ds, "max_surface_points"):
+                if current_train_points != prev_train_points:
+                    print(f"[progressive] epoch {epoch}: train surface points {prev_train_points} -> {current_train_points}", flush=True)
+                    prev_train_points = current_train_points
+                train_ds.max_surface_points = current_train_points
+
         train_metrics = train_one_epoch(
             model=forward_model,
             anp_head=anp_head,
@@ -1719,7 +1754,7 @@ def main() -> None:
             best_model_state = snapshot_module_state(model)
             best_anp_state = snapshot_module_state(anp_head)
 
-        current_val = eval_metrics.get(val_primary_key)
+        current_val = eval_metrics.get(best_tracking_metric_key)
         if current_val is not None and current_val < best_val_metric:
             best_val_metric = current_val
             best_epoch = epoch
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT mitsuha:

## Results

### Summary

**Both trials failed to beat the baseline.** Progressive resolution training converges significantly slower than full-resolution training from the start, and 25k points for extended training causes catastrophic instability.

| Metric | Baseline | Trial 1 (25k→50k@200) | Trial 2 (30k→50k@100) |
|--------|----------|----------------------|----------------------|
| `val_primary/surface_rel_l2_pct` | **3.833%** | 8.147% | 4.711% |
| Best epoch | 511 | 106 | 257 |
| Total epochs | 517 | 263 (killed) | 270 (killed) |
| Status | — | DIVERGED | Converging but slow |
| W&B run | `ncl1dh88` | `u94n7qjo` | `972xv0ty` |

### Trial 1 — 25k→50k@200 (FAILED — catastrophic divergence)

- Schedule: 25k points epochs 0–200, 50k points epochs 201–999
- **Catastrophic training instability at epoch ~112**: grad norms escalated 0.89 → 4.18 → 10.4 → 33.8 → ∞ over 5 epochs
- Best EMA val: 8.147% at epoch 106, frozen there permanently
- After divergence, online val degraded to ~68.15% (near-random), never recovered
- Killed at epoch 263 after running uselessly for 150+ epochs with Infinity grad norms

**Root cause**: Extended training at 25k points (only 50% of full resolution) causes the model to overfit to the coarse representation. By epoch ~110, the model has seen 110 × 394 = 43,340 batches of 25k-point meshes — it specializes so deeply in coarse patterns that gradient signals become unstable. The EMA buffer protected the best checkpoint but the online model was irrecoverable.

### Trial 2 — 30k→50k@100 (NEGATIVE — doesn't beat baseline)

- Schedule: 30k points epochs 0–100, 50k points epochs 101–999
- Smooth transition from 30k→50k at epoch 100 with NO regression (continued improving)
- Best EMA val: **4.7105%** at epoch 257
- Convergence rate analysis (diminishing returns):
  - ep100→150: 7.53% → 5.84% (−0.034%/ep)
  - ep150→200: 5.84% → 5.11% (−0.015%/ep)
  - ep200→250: 5.11% → 4.74% (−0.008%/ep)
  - ep250→270: 4.74% → 4.71% (−0.001%/ep)
- Late-training instability emerging: grad clip events escalated 3→11→16→35→44→92 over epochs 260–269
- Killed at epoch 270 — projected to reach ~4.3–4.5% by timeout, nowhere near 3.833% baseline

**Key finding**: The 30k→50k transition works mechanically (no divergence), but the "coarse-to-fine" phase loses ~150 epochs of effective full-resolution training. The baseline, training at 50k from epoch 1, reaches 3.833% at epoch 511. Trial 2, which only got 170 epochs of full-resolution training (epochs 101–270), was still at 4.71%. The convergence rate at 50k points after progressive warmup appears similar to the baseline rate, but the model starts from a worse position because the 30k-point training didn't fully transfer.

### Bug fix included

Fixed `primary_metric_key` shadowing bug: renamed local variable to `best_tracking_metric_key` to avoid `UnboundLocalError` when the training loop variable shadowed the outer function.

### Commands used

**Trial 1:**
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 \
  --progressive-points-schedule "25000:200,50000:999" \
  --wandb-group dm-progressive-resolution \
  --agent mitsuha --wandb-name "mitsuha/dm-progressive-25k-200ep"
```

**Trial 2:**
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 \
  --progressive-points-schedule "30000:100,50000:999" \
  --wandb-group dm-progressive-resolution \
  --agent mitsuha --wandb-name "mitsuha/dm-progressive-30k-100ep"
```

### What happened

The progressive resolution hypothesis did not work for DrivAerML. The core assumption — that early training learns coarse patterns that transfer well to full resolution — does not hold strongly enough for this architecture/dataset combination. Two key findings:

1. **25k is too aggressive**: Extended training at 50% resolution causes the model to specialize in coarse representations, leading to catastrophic gradient instability. The model essentially overfits to the lower-resolution signal.

2. **30k→50k works mechanically but loses training budget**: The transition from 30k→50k is smooth, but the epochs spent at reduced resolution don't contribute as much as full-resolution epochs. The model reaches 7.53% after 100 epochs at 30k, vs the baseline which is typically around 6-7% at epoch 100 at full resolution. The ~0.5-1% deficit from coarse training persists and compounds — the convergence curve is offset, not steeper.

The speed gain (~20-30% faster epochs at reduced resolution) does not compensate for the convergence penalty. DrivAerML's surface fields contain fine-scale features (pressure, wall shear stress) that require full-resolution context from the beginning of training.

### Suggested follow-ups

- **Multi-stage LR warmup at transitions**: If progressive training is revisited, reset or warm up the learning rate when transitioning to higher resolution, since the model encounters a distribution shift
- **Gentler progression**: 40k→50k@50 (minor reduction, short duration) might avoid the instability while still gaining some speed
- **This is likely a dead end for DrivAerML**: The dataset's surface fields are high-frequency enough that coarse pretraining doesn't provide useful inductive bias. Training budget is better spent at full resolution throughout.

tcapelle:
ADVISOR: Closing PR #3252 because ADVISOR: Closing — progressive resolution is a dead end for DM. Trial 1 (25k→50k@200): crashed at ep112 with catastrophic gradient explosion (grad_norm peaked 2089), best 8.147% (+113%). Trial 2 (30k→50k@100): mechanically stable but converged to 4.711% (+23%) — the 30k warmup imposed a persistent ~0.6pp deficit vs baseline at the same epoch count, and the ~20% speed gain from smaller point count doesn't compensate.

Root cause: DM surface fields require full 50k-point context from epoch 1. The reduced-resolution pretraining creates a representation deficit that late-stage fine-tuning cannot recover within the training budget. This is consistent with 50k being the only viable surface point count (16k/32k/64k all dead).

Good execution and thorough analysis. The primary_metric_key bug fix is noted — this fix has been independently discovered by multiple students and is tracked..

PR #3256, opened 2026-04-23
DrivAerML Coordinate Noise Augmentation (Input Perturbation Regularization)

tcapelle:
## Hypothesis

DrivAerML trains on 50k surface points per mesh, where each point has xyz coordinates and surface normals. The model uses Fourier positional encoding on the coordinates. Currently, the model sees the EXACT same point coordinates every time it processes a given mesh.

By adding small Gaussian noise (σ=0.001–0.005 in normalized coordinate space) to the xyz coordinates during training, we force the model to learn surface field predictions that are locally smooth and robust to small position perturbations. This is:

1. **A form of data augmentation** — each epoch sees slightly different point positions, increasing effective data diversity
2. **Implicit regularization** — forces smooth learned representations
3. **Physics-motivated** — aerodynamic surface fields should be spatially continuous; if moving a point by 0.1mm changes the prediction dramatically, the model has overfit to exact positions
4. **Affects Fourier features** — noise on coordinates creates noisy Fourier features, preventing over-reliance on specific frequency components

At test time, use clean (unperturbed) coordinates.

## Baseline

- test: 4.685%
- W&B run: ncl1dh88

**Target to beat: 3.833%**. External target: <3.71% (AB-UPT). Gap: ~0.123 pp.

## Known Bug Fix

There is a `primary_metric_key` shadowing bug in `train.py` where a local variable (line ~1646–1648) shadows the function (line ~1428), causing an `UnboundLocalError`. Fix by renaming the local variable to `best_tracking_metric_key` if you encounter it.

## Implementation

1. Add `--coord-noise-std` flag (float, default 0.0 — no noise)
2. During training only (not eval), add Gaussian noise to xyz coordinates:
   ```python
   xyz = xyz + torch.randn_like(xyz) * noise_std
   ```
3. Apply noise **BEFORE** Fourier encoding (so the Fourier features are computed on noisy positions)
4. Do **NOT** add noise to normals or target values — only to position coordinates
5. Log `coord_noise_std` to W&B config

## Trial 1 — σ=0.001 (conservative: ~0.1% of coordinate range)

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --coord-noise-std 0.001 \
  --wandb_group dm-coord-noise
```

## Trial 2 — σ=0.005 (moderate: ~0.5% of coordinate range)

Same as Trial 1 but with `--coord-noise-std 0.005`:

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --coord-noise-std 0.005 \
  --wandb_group dm-coord-noise
```

## Reporting

Report `val_primary/surface_rel_l2_pct` at best epoch for both trials. Target: beat **3.833%**.

Use `--wandb_group dm-coord-noise` so both trials appear together in W&B.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -165,6 +165,7 @@ class TrainConfig:
     surface_anchor_points: int = 8_000
     volume_anchor_points: int = 8_000
     grad_clip: float = 0.0
+    coord_noise_std: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
@@ -1254,6 +1255,27 @@ def evaluate_abupt(
     return {"surface_mae": total_surface / max(count_surface, 1), "volume_mae": total_volume / max(count_volume, 1)}
 
 
+def _apply_coord_noise(
+    surface_x: torch.Tensor,
+    noise_std: float,
+    space_dim: int,
+    enable_fourier: bool,
+    fourier_freqs: tuple[float, ...] = (0.5, 2.0, 8.0, 32.0),
+) -> torch.Tensor:
+    surface_x = surface_x.clone()
+    surface_x[..., :space_dim] = surface_x[..., :space_dim] + torch.randn_like(surface_x[..., :space_dim]) * noise_std
+    if enable_fourier:
+        fourier_dim = space_dim * len(fourier_freqs) * 2
+        base_dim = surface_x.shape[-1] - fourier_dim
+        noisy_coords = surface_x[..., :space_dim]
+        fourier_pieces = []
+        for freq in fourier_freqs:
+            fourier_pieces.append(torch.sin(noisy_coords * freq))
+            fourier_pieces.append(torch.cos(noisy_coords * freq))
+        surface_x = torch.cat([surface_x[..., :base_dim], torch.cat(fourier_pieces, dim=-1)], dim=-1)
+    return surface_x
+
+
 def train_one_epoch(
     model: torch.nn.Module,
     anp_head: ANPSurfaceDecoder | None,
@@ -1271,6 +1293,9 @@ def train_one_epoch(
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
     volume_loss_weight: float = 1.0,
+    coord_noise_std: float = 0.0,
+    space_dim: int = 3,
+    enable_fourier: bool = False,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
@@ -1319,9 +1344,14 @@ def train_one_epoch(
                         outputs = maybe_apply_anp(outputs, prepared, anp_head)
                         loss, _ = loss_grouped_tandem(prepared, outputs, volume_loss_weight=volume_loss_weight)
                 else:
+                    noisy_surface_x = (
+                        _apply_coord_noise(batch.surface_x, coord_noise_std, space_dim, enable_fourier)
+                        if coord_noise_std > 0
+                        else batch.surface_x
+                    )
                     with autocast_context(device, amp_mode):
                         outputs = model(
-                            surface_x=batch.surface_x,
+                            surface_x=noisy_surface_x,
                             surface_mask=batch.surface_mask,
                             volume_x=batch.volume_x,
                             volume_mask=batch.volume_mask,
@@ -1688,6 +1718,9 @@ def main() -> None:
             grad_clip=config.grad_clip,
             grad_accum_steps=config.grad_accum_steps,
             volume_loss_weight=config.volume_loss_weight,
+            coord_noise_std=config.coord_noise_std,
+            space_dim=bundle.spec.space_dim,
+            enable_fourier=config.enable_fourier,
         )
 
         if ema is not None:


tcapelle:
STUDENT alphonse:

## Results

### Trial 1: σ=0.001 (coord_noise_std=0.001)

| Metric | Baseline | Trial 1 (best) | Delta |
|--------|----------|-----------------|-------|
| val_primary/surface_rel_l2_pct | 3.833% (ep511) | **6.450%** (ep171) | +2.617 pp (1.68x worse) |

- **W&B run:** `hapy2g9k`
- **Status:** Diverged at epoch 172. Grad norms exploded exponentially: 0.5 → 9.2 → 32.8 → 1,354 → 733K → 1.9M over 5 epochs. Killed at epoch 176.
- **Notable:** Had an earlier instability spike at epochs 124-137 (peaked at 10.68%) from which it fully recovered. The second spike at epoch 172 proved fatal.
- **Peak memory:** ~2.8 GB RSS per worker (GPU VRAM not explicitly logged)

### Trial 2: σ=0.005 (coord_noise_std=0.005)

| Metric | Baseline | Trial 2 (best) | Delta |
|--------|----------|-----------------|-------|
| val_primary/surface_rel_l2_pct | 3.833% (ep511) | **21.44%** (ep23) | +17.61 pp (5.59x worse) |

- **W&B run:** `qfu1j89b`
- **Status:** Diverged by epoch ~25. Grad norms reached Infinity by epoch ~52 and never recovered. Steady-state val ~62.1% (random outputs). Killed at epoch 170.

### Commands used

```bash
# Trial 1
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --coord-noise-std 0.001 \
  --agent alphonse --wandb-name "alphonse/dm-coord-noise-001" --wandb-group dm-coord-noise

# Trial 2
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --coord-noise-std 0.005 \
  --agent alphonse --wandb-name "alphonse/dm-coord-noise-005" --wandb-group dm-coord-noise
```

### What happened

**This hypothesis is dead.** Both noise levels caused catastrophic training instability and eventual divergence. The coordinate noise interacts badly with the Fourier positional encoding — perturbing coordinates by even σ=0.001 changes the high-frequency Fourier features (freq=32.0) by up to 32× the noise magnitude, creating large downstream perturbations that destabilize training.

Key observations:
1. **Higher noise diverges faster:** σ=0.005 diverged by epoch 25; σ=0.001 survived to epoch 171 before diverging.
2. **Instability is intermittent then catastrophic:** Trial 1 had a recoverable spike (ep124-137) before the fatal one (ep172+). The grad norm explosion is exponential once triggered.
3. **Fourier amplification is the root cause:** The `append_fourier_features` function multiplies raw coordinates by frequencies [0.5, 2.0, 8.0, 32.0]. Noise at σ=0.001 becomes σ=0.032 at the highest frequency — a 32× amplification. At σ=0.005, this becomes σ=0.16, which is enormous relative to the normalized coordinate range.
4. **EMA provided only checkpoint protection, not stability:** EMA preserved the best checkpoint but couldn't prevent the underlying model from diverging.
5. **Grad clipping was overwhelmed:** Despite grad_clip=0.5, norms reached 733K+ and Infinity, indicating the gradient explosion was too fast for per-step clipping to contain.

### Suggested follow-ups

1. **Frequency-aware noise scaling:** If coordinate noise is revisited, scale noise inversely with the max Fourier frequency: σ_effective = σ / max_freq. This would keep Fourier feature perturbations bounded.
2. **Feature-space noise instead:** Apply noise to the hidden representations after the Fourier encoding, where the model has learned representations that may be more robust to small perturbations.
3. **Noise scheduling:** Start with zero noise and slowly increase, or use a warmup period before enabling noise, to let the model build stable representations first.
4. **Lower noise magnitude:** σ=0.0001 or σ=0.0002 might survive but the benefit at such tiny scales is questionable.

tcapelle:
ADVISOR: Closing PR #3256 because ADVISOR: Closing — coordinate noise augmentation is fundamentally incompatible with Fourier positional encoding. σ=0.001 → 6.45% best (+68%, crashed ep172), σ=0.005 → 21.44% best (+459%, 117/168 grad norms were Infinity). Root cause: Fourier frequencies up to 32.0 amplify coordinate noise by 32x — σ=0.001 in coord space becomes σ=0.032 in feature space at the highest frequency. This is structural, not tunable.

Your analysis of the Fourier amplification mechanism was spot-on. The feature-space noise idea you suggested (perturbing hidden representations AFTER Fourier encoding) is the correct reformulation — it avoids the amplification problem entirely..

PR #3217, opened 2026-04-23
DrivAerML: Fourier-encode surface normals (angular freq features)

tcapelle:
## Hypothesis

Surface normals (columns 3–5 of the DrivAerML input tensor, `SURFACE_X_DIM=7`) are currently passed to the model as raw floats, while xyz coordinates (columns 0–2) already receive Fourier positional encoding. Brook discovered this asymmetry in PR #3213.

Motivated by NeRF (Mildenhall et al., 2020): MLPs cannot learn high-frequency functions from raw inputs, but Fourier features lift them into a high-frequency-accessible space. Surface normals encode angular information about the car body — sharp creases, flat panels, curved surfaces — that is inherently high-frequency in angular space. We expect Fourier-encoding the normals to give the model a richer, more expressive representation of local surface geometry.

We run two trials to find the right frequency budget for unit-vector inputs:

- **Trial 1**: same number of frequency bands as xyz (normals are bounded in [-1, 1], so the same bands apply directly)
- **Trial 2**: half the frequency bands (normals are unit vectors; lower frequencies may suffice and reduce over-parameterisation)

Reference: Mildenhall et al. (2020), *NeRF: Representing Scenes as Neural Radiance Fields for View Synthesis* — https://arxiv.org/abs/2003.08934

## Instructions

**Step 0 — smoke test first.** Before running full training, run 2 epochs on DrivAerML to confirm no shape mismatches after extending the Fourier encoding:
```bash
cd target/icml2026
SENPAI_MAX_EPOCHS=9999 python train.py --dataset drivaerml \
  --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 \
  --enable-fourier --fourier-encode-normals \
  --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --epochs 2 --batch-size 1 \
  --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 \
  --max-train-batches 10 --max-eval-batches 5
```
If a new flag is needed, add it to the Fourier encoding section in `train.py` (or wherever Fourier features are assembled — check `prepare_drivaerml.py` or the model forward pass). The key change is: locate where Fourier positional encoding is applied to xyz (columns 0–2) and extend the same logic to surface normals (columns 3–5). The input MLP should handle the expanded `input_dim` automatically if it reads the dimension from data at build time.

**Trial 1 — same frequency bands as xyz:**
```bash
cd target/icml2026
[...]

PR #3258, opened 2026-04-23
DM: auxiliary surface normal prediction (multi-task regularization)

tcapelle:
## Hypothesis

The Transolver model receives surface normals (nx, ny, nz) as input features but is never trained to **predict** them. Adding an auxiliary decoder head that predicts surface normals from the model's latent representation forces the model to retain geometric information throughout its layers. This is multi-task regularization — the auxiliary task acts as an information bottleneck that preserves curvature-aware features in the latent space.

This is **fundamentally different** from Fourier-encoding normals (PR #3217, dead — encoded normals as input features and failed). Here, normals are a **supervision target**, not an input encoding. The model already has the normals in its input — the auxiliary task ensures it doesn't discard this geometric information in deeper layers.

**Physics motivation:** Pressure and skin friction are strongly correlated with local surface geometry. A latent space that explicitly represents surface normals should produce better aerodynamic field predictions, especially at high-curvature regions (leading edges, mirrors, wheel arches) where the model currently makes the most errors.

## Baseline

- test: 4.685%
- W&B run: `ncl1dh88`

## Known Bug Fix

There is a `primary_metric_key` shadowing bug in `train.py` where a local variable (line ~1646-1648) shadows the function (line ~1428), causing an `UnboundLocalError`. **Fix by renaming the local variable to `best_tracking_metric_key`** if you encounter it.

## Implementation Guidance

1. Add `--aux-normal-weight` flag (float, default 0.0 meaning disabled)
2. Add a small auxiliary MLP head (e.g., 2 layers, 256→3) that takes the model's output latent representation and predicts (nx, ny, nz) for each surface point
3. Compute `aux_loss = MSE(predicted_normals, true_normals) * aux_normal_weight`
4. Total loss = primary_loss + aux_loss
5. Log `aux_normal_loss` to W&B each step
6. The auxiliary head is **ONLY used during training** — discard it at eval time
7. Normals should be normalized to unit vectors in both prediction and target

### Trial 1 — aux_normal_weight=0.1

```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --aux-normal-weight 0.1 \
  --wandb_group dm-aux-normals
```

### Trial 2 — aux_normal_weight=0.01 (lighter regularization)

Same as Trial 1 but with `--aux-normal-weight 0.01`

## Success Criterion

Report `val_primary/surface_rel_l2_pct` at best epoch. Target: beat **3.833%**.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -86,6 +86,23 @@ class EMAWithWarmup:
         self.backup = None
 
 
+class AuxNormalHead(torch.nn.Module):
+    def __init__(self, hidden_dim: int):
+        super().__init__()
+        self.net = torch.nn.Sequential(
+            torch.nn.LayerNorm(hidden_dim),
+            torch.nn.Linear(hidden_dim, 256),
+            torch.nn.GELU(),
+            torch.nn.Linear(256, 3),
+        )
+        torch.nn.init.zeros_(self.net[-1].weight)
+        torch.nn.init.zeros_(self.net[-1].bias)
+
+    def forward(self, hidden: torch.Tensor) -> torch.Tensor:
+        out = self.net(hidden)
+        return F.normalize(out, dim=-1)
+
+
 LEGACY_VAL_ALIAS = {
     "val_in_dist": "p_in",
     "val_ood_cond": "p_oodc",
@@ -169,6 +186,7 @@ class TrainConfig:
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
+    aux_normal_weight: float = 0.0
 
 
 @dataclass
@@ -1271,10 +1289,15 @@ def train_one_epoch(
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
     volume_loss_weight: float = 1.0,
+    aux_normal_head: AuxNormalHead | None = None,
+    aux_normal_weight: float = 0.0,
+    aux_normal_ema: EMAWithWarmup | None = None,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
         anp_head.train()
+    if aux_normal_head is not None:
+        aux_normal_head.train()
     running = {"loss": 0.0}
     steps = 0
     micro_batches_total = 0
@@ -1327,6 +1350,14 @@ def train_one_epoch(
                             volume_mask=batch.volume_mask,
                         )
                         loss, _ = loss_grouped(batch, outputs, transform, volume_loss_weight=volume_loss_weight)
+                        if aux_normal_head is not None and aux_normal_weight > 0 and outputs.get("surface_hidden") is not None:
+                            pred_normals = aux_normal_head(outputs["surface_hidden"])
+                            target_normals = F.normalize(batch.surface_x[:, :, 3:6], dim=-1)
+                            cos_sim = F.cosine_similarity(pred_normals, target_normals, dim=-1)
+                            aux_loss = ((1.0 - cos_sim) * batch.surface_mask).sum() / batch.surface_mask.sum().clamp(min=1)
+                            loss = loss + aux_loss * aux_normal_weight
+                            running.setdefault("aux_normal_loss", 0.0)
+                            running["aux_normal_loss"] += float(aux_loss.detach().cpu().item())
 
             accum_loss += float(loss.detach().cpu().item())
             (loss / grad_accum_steps).backward()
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT emma:

## Results

### Trial 1 — aux_normal_weight=0.1

| Metric | Value |
|--------|-------|
| **val_primary/surface_rel_l2_pct** | **4.724%** (epoch 198) |
| test surface_rel_l2_pct (best ckpt) | 5.639% |
| test surface_rel_l2_pct (final) | 6.164% |
| aux_normal_loss (final) | 0.00018 |
| Total epochs | 198 (hit 180-min timeout) |
| W&B run | `fkiz67ki` |

**Command:**
```bash
CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=180 python train.py \
  --ema-decay 0.9995 --aux-normal-weight 0.1 \
  --agent emma --wandb-name "emma/dm-aux-normals-w0.1" --wandb-group dm-aux-normals
```

### Trial 2 — aux_normal_weight=0.01

| Metric | Value |
|--------|-------|
| **val_primary/surface_rel_l2_pct** | **4.660%** (epoch 211) |
| test surface_rel_l2_pct (best ckpt) | 5.541% |
| test surface_rel_l2_pct (final) | 5.638% |
| aux_normal_loss (final) | 0.00067 |
| Total epochs | 211 (hit 180-min timeout) |
| W&B run | `g9zs1wsh` |

**Command:**
Same as Trial 1 but with `--aux-normal-weight 0.01`.

### Baseline comparison

| | Baseline (ncl1dh88) | Trial 1 (w=0.1) | Trial 2 (w=0.01) |
|--|--|--|--|
| val surface_rel_l2_pct | **3.833%** (ep 511) | 4.724% (ep 198) | 4.660% (ep 211) |
| test surface_rel_l2_pct | 4.685% | 5.639% | 5.541% |

## What happened

**The auxiliary normal prediction did not beat the baseline.** Both trials ran ~200 epochs before hitting the 180-minute timeout (vs the baseline's 511 epochs), so a direct epoch-matched comparison isn't possible. However, there are reasons to believe more epochs wouldn't close the gap:

1. **The auxiliary task solved itself too quickly.** The aux_normal_loss dropped from ~0.9 to <0.001 within the first ~30 epochs. Once the head learned to predict normals trivially, it stopped providing useful gradient signal to the backbone. This is the "input leakage" pitfall — the model already receives normals (nx, ny, nz) as input features (surface_x channels 3-5), so predicting them from the latent representation is a shortcut the auxiliary head can learn without forcing deeper geometric retention.

2. **Trial 2 (lighter weight) slightly outperformed Trial 1**, suggesting that the heavier aux weight may actually be hurting by adding noise to the primary gradients without proportional benefit.

3. **The convergence trajectory** at epoch ~200 shows both trials tracking above where the baseline was at comparable epoch counts. The baseline's advantage compounds over cosine cycles.

## Research notes

The researcher-agent flagged the input leakage concern before implementation — normals are in the model's input, so predicting them is near-trivial and doesn't shape the representation meaningfully. This is fundamentally different from predicting a held-out geometric property (e.g., curvature, which must be computed from the neighborhood). The cosine similarity loss was used instead of MSE for better gradient behavior on unit vectors.

## Suggested follow-ups

1. **Curvature prediction instead of normals** — predict local mean/Gaussian curvature as the auxiliary target. This can't be shortcut from input features and would force the model to learn neighborhood-aware geometric representations.
2. **Gradient detachment ablation** — detach the backbone from the aux head to confirm whether the aux gradients are helpful or harmful.
3. **More epochs** — if the advisor wants a definitive comparison, the runs need the full 500+ epoch budget that the baseline had. The 180-min timeout was the limiting factor.

tcapelle:
ADVISOR: Closing PR #3258 because ADVISOR: Closing — auxiliary surface normal prediction is conceptually flawed for DM. w=0.1: 4.724% (+23%), w=0.01: 4.661% (+22%). Both runs also suffered epoch starvation (200 vs 511 epochs due to throughput loss from the aux head), but the fundamental problem is input leakage: surface normals are already present in the input features, so the auxiliary head learns to reconstruct them trivially (aux_loss collapses to <0.001 within 30-50 epochs). The task provides no meaningful information bottleneck.

Your analysis of the input leakage was spot-on. The curvature prediction idea you suggested is the right reformulation — predicting mean/Gaussian curvature requires learning about local neighborhood structure that ISN'T directly in the input features. But that's a different hypothesis..

PR #3263, opened 2026-04-24
DrivAerML: Sparse Mixture-of-Experts FFN (K=4/8 experts, top-2 routing)

tcapelle:
## Hypothesis

The remaining DM error at 3.833% is systematic spatial bias — the model consistently under/over-predicts in specific aerodynamic regions (stagnation zone, A-pillar vortex, underbody separation, rear wake). The current architecture uses a **single shared FFN** in each transformer layer, forcing the same nonlinear transformation on tokens from fundamentally different flow regimes. We hypothesize that allowing **regime-specific computation** via sparse Mixture-of-Experts (MoE) FFN layers will reduce spatial bias by letting different experts specialize in different flow physics.

**MoE replaces each FFN with K expert FFNs and a learned router.** Each surface point token is routed to the top-2 experts based on its representation — the router learns which "flow regime" each token belongs to. This allows stagnation-zone tokens to use different nonlinear mappings than wake tokens, directly addressing the structural heterogeneity in the prediction error.

MoE is well-established in LLMs (Switch Transformer, Mixtral) and recently applied to scientific ML (FourCastNet-MoE for weather, arXiv 2409.16295). With 50k tokens per sample and only K=4-8 experts, each expert sees 6-25k tokens — plenty of signal for routing to specialize. Total compute stays similar to baseline since each token uses only top-2 experts rather than all K.

### Implementation

1. **Add CLI flags:**
   ```python
   parser.add_argument('--moe-num-experts', type=int, default=0,
                       help='Number of MoE experts per FFN layer (0=disabled, standard FFN)')
   parser.add_argument('--moe-top-k', type=int, default=2,
                       help='Number of experts each token is routed to')
   parser.add_argument('--moe-balance-coeff', type=float, default=0.01,
                       help='Load-balancing auxiliary loss coefficient')
   ```

2. **Implement the MoE FFN module** (new class):
   ```python
   class MoEFFN(nn.Module):
       def __init__(self, hidden_dim, ffn_dim, num_experts, top_k=2):
           super().__init__()
           self.num_experts = num_experts
           self.top_k = top_k
           # Router: linear projection to num_experts logits
           self.router = nn.Linear(hidden_dim, num_experts, bias=False)
           # K independent FFN experts (same architecture as baseline FFN)
           self.experts = nn.ModuleList([
               nn.Sequential(
                   nn.Linear(hidden_dim, ffn_dim),
                   nn.GELU(),
                   nn.Linear(ffn_dim, hidden_dim)
               ) for _ in range(num_experts)
           ])
       
       def forward(self, x):
           # x: [B, N, D]
           B, N, D = x.shape
           # Router logits and top-k selection
           router_logits = self.router(x)  # [B, N, K]
           top_k_logits, top_k_indices = router_logits.topk(self.top_k, dim=-1)  # [B, N, top_k]
           top_k_weights = F.softmax(top_k_logits, dim=-1)  # [B, N, top_k]
           
           # Compute expert outputs (loop over top-k selected experts)
           output = torch.zeros_like(x)
           for i in range(self.top_k):
               expert_idx = top_k_indices[:, :, i]  # [B, N]
               weight = top_k_weights[:, :, i].unsqueeze(-1)  # [B, N, 1]
               for e in range(self.num_experts):
                   mask = (expert_idx == e)  # [B, N]
                   if mask.any():
                       # Gather tokens for this expert
                       expert_input = x[mask]  # [num_tokens, D]
                       expert_output = self.experts[e](expert_input)  # [num_tokens, D]
                       output[mask] += weight[mask] * expert_output
           
           # Load-balancing loss (stored as attribute, added to total loss)
           # Fraction of tokens routed to each expert
           routing_probs = F.softmax(router_logits, dim=-1)  # [B, N, K]
           density = routing_probs.mean(dim=1)  # [B, K] — avg prob per expert
           # Ideal uniform: 1/K per expert
           self.balance_loss = (density * self.num_experts).var(dim=-1).mean()
           
           return output
   ```

3. **Replace FFN in each transformer layer** when `--moe-num-experts > 0`:
   In the model init, where the standard FFN is created for each transformer layer, check:
   ```python
   if args.moe_num_experts > 0:
       ffn = MoEFFN(hidden_dim, ffn_dim, args.moe_num_experts, args.moe_top_k)
   else:
       ffn = standard_ffn  # existing code
   ```

4. **Add load-balancing loss to total loss:**
   After computing the main MSE loss, accumulate the balance losses from all MoE layers:
   ```python
   if args.moe_num_experts > 0:
       balance_loss = sum(layer.ffn.balance_loss for layer in self.transformer_layers 
                         if hasattr(layer.ffn, 'balance_loss'))
       total_loss = main_loss + args.moe_balance_coeff * balance_loss
   ```

5. **Log expert utilization** (optional but helpful for analysis):
   Log the routing distribution per layer to W&B every 50 steps — this tells us whether experts are actually specializing.

### Trials

**Smoke test first (3 epochs)** — verify shapes, loss is finite, experts get non-degenerate routing.

**Trial 1 — K=4 experts, top-2 (regime-scale):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --moe-num-experts 4 --moe-top-k 2 --moe-balance-coeff 0.01 \
  --wandb_group dm-moe-ffn
```

**Trial 2 — K=8 experts, top-2 (finer-grained):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --moe-num-experts 8 --moe-top-k 2 --moe-balance-coeff 0.01 \
  --wandb_group dm-moe-ffn
```

**Note on throughput:** MoE with expert loops may be slower than a single FFN. If throughput drops >30%, consider using `torch.compile` on the expert modules or implementing a fused scatter/gather. But start with the simple loop — correctness first.

Report `val_primary/surface_rel_l2_pct` for both trials. Also log the expert routing distribution (which experts get which fraction of tokens) — this tells us whether the MoE is actually learning regime-specific specialization or just converging to a uniform split.

## Baseline

| Metric | Value | Epoch |
|--------|-------|-------|

- W&B run: `ncl1dh88`

Reproduce:
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -169,6 +169,9 @@ class TrainConfig:
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
+    moe_num_experts: int = 0
+    moe_top_k: int = 2
+    moe_balance_coeff: float = 0.01
 
 
 @dataclass
@@ -497,6 +500,107 @@ def cp_panel_prior_index(config: TrainConfig, bundle: DatasetBundle) -> int | No
         base_dim -= 1
         return base_dim
     return None
+
+
+class MoEFFN(torch.nn.Module):
+    """Sparse Mixture-of-Experts FFN: each token routed to top-k of K experts."""
+
+    def __init__(self, hidden_dim: int, ffn_dim: int, num_experts: int, top_k: int = 2):
+        super().__init__()
+        self.num_experts = num_experts
+        self.top_k = top_k
+        self.router = torch.nn.Linear(hidden_dim, num_experts, bias=False)
+        torch.nn.init.normal_(self.router.weight, std=0.01)
+        self.experts = torch.nn.ModuleList([
+            torch.nn.Sequential(
+                torch.nn.Linear(hidden_dim, ffn_dim),
+                torch.nn.GELU(),
+                torch.nn.Linear(ffn_dim, hidden_dim),
+            )
+            for _ in range(num_experts)
+        ])
+        self.balance_loss = torch.tensor(0.0)
+        self.register_buffer("_routing_counts", torch.zeros(num_experts), persistent=False)
+        self.jitter_noise = 1.0
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        B, N, D = x.shape
+        x_flat = x.reshape(B * N, D)
+        router_logits = self.router(x_flat)  # [B*N, K]
+
+        if self.training and self.jitter_noise > 0:
+            noise = torch.randn_like(router_logits) * self.jitter_noise
+            router_logits = router_logits + noise
+
+        top_k_logits, top_k_indices = router_logits.topk(self.top_k, dim=-1)
+        top_k_weights = F.softmax(top_k_logits, dim=-1)  # [B*N, top_k]
+
+        output = torch.zeros_like(x_flat)
+        for k_idx in range(self.top_k):
+            indices = top_k_indices[:, k_idx]  # [B*N]
+            weights = top_k_weights[:, k_idx].unsqueeze(-1)  # [B*N, 1]
+            for e in range(self.num_experts):
+                mask = indices == e
+                if mask.any():
+                    expert_out = self.experts[e](x_flat[mask])
+                    output[mask] = output[mask] + weights[mask] * expert_out
+
+        routing_probs = F.softmax(router_logits, dim=-1)  # [B*N, K]
+        tokens_per_expert = F.one_hot(top_k_indices[:, 0], self.num_experts).float().mean(dim=0)
+        avg_prob = routing_probs.mean(dim=0)
+        # Switch Transformer load balance + entropy regularization for anti-collapse
+        switch_loss = (tokens_per_expert * avg_prob * self.num_experts).sum()
+        entropy = -(avg_prob * (avg_prob + 1e-8).log()).sum()
+        max_entropy = math.log(self.num_experts)
+        entropy_penalty = max_entropy - entropy
+        self.balance_loss = switch_loss + entropy_penalty
+
+        self._routing_counts = tokens_per_expert.detach()
+
+        return output.reshape(B, N, D)
+
+
+def _replace_ffn_with_moe(model: torch.nn.Module, config: TrainConfig) -> None:
+    """Replace UpActDownMlp in each TransformerBlock with MoEFFN."""
+    if config.moe_num_experts <= 0:
+        return
+    backbone = model.backbone
+    hidden_dim = model.n_hidden
+    ffn_dim = int(math.ceil(hidden_dim * config.model_mlp_ratio))
+    for block in backbone.blocks:
+        block.mlp = MoEFFN(hidden_dim, ffn_dim, config.moe_num_experts, config.moe_top_k)
+
+
+def _collect_moe_balance_loss(model: torch.nn.Module) -> torch.Tensor:
+    """Sum balance losses from all MoE layers."""
+    total = torch.tensor(0.0, device=next(model.parameters()).device)
+    raw = model._orig_mod if hasattr(model, '_orig_mod') else model
+    backbone = getattr(raw, 'backbone', None)
+    if backbone is None:
+        return total
+    for block in backbone.blocks:
+        if isinstance(block.mlp, MoEFFN):
+            total = total + block.mlp.balance_loss
+    return total
+
+
+def _log_moe_routing_stats(model: torch.nn.Module, run, epoch: int) -> dict[str, float]:
+    """Log per-layer expert routing distribution and router norms to W&B."""
+    raw = model._orig_mod if hasattr(model, '_orig_mod') else model
+    backbone = getattr(raw, 'backbone', None)
+    if backbone is None:
+        return {}
+    metrics: dict[str, float] = {}
+    for i, block in enumerate(backbone.blocks):
+        if not isinstance(block.mlp, MoEFFN):
+            continue
+        counts = block.mlp._routing_counts.detach().cpu()
+        for e in range(block.mlp.num_experts):
+            metrics[f"moe/layer{i}_expert{e}_frac"] = float(counts[e])
+        metrics[f"moe/layer{i}_routing_entropy"] = float(
+            -(counts * (counts + 1e-8).log()).sum()
+        )
+    return metrics
 
 
 def build_model(config: TrainConfig, bundle: DatasetBundle) -> torch.nn.Module:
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT gojo:

## Results

### Trial 1: K=4 experts, top-2 routing

| Metric | Baseline (#3072) | K=4 MoE | Delta |
| val_primary/surface_rel_l2_pct | **3.833%** (ep511) | 5.092% (ep173) | +1.259% ❌ |
| test_primary/surface_rel_l2_pct | **4.685%** | 5.905% | +1.220% ❌ |
| Epochs completed | 511 | 173 | -338 |
| Throughput | ~0.60 min/ep (compiled) | 1.04 min/ep | ~1.7x slower |

- W&B run: `rei05zxk`
- Command: `CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=180 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --moe-num-experts 4 --moe-top-k 2 --moe-balance-coeff 0.1 --wandb-name "gojo/dm-moe-k4-top2-bal01" --wandb-group dm-moe-ffn --agent gojo`

### Trial 2: K=8 experts, top-2 routing

| Metric | Baseline (#3072) | K=8 MoE | Delta |
| val_primary/surface_rel_l2_pct | **3.833%** (ep511) | 5.348% (ep136) | +1.515% ❌ |
| test_primary/surface_rel_l2_pct | **4.685%** | 6.194% | +1.509% ❌ |
| Epochs completed | 511 | 136 | -375 |
| Throughput | ~0.60 min/ep (compiled) | 1.32 min/ep | ~2.2x slower |

- W&B run: `hjkzjnn8`
- Command: same as K=4 but with `CUDA_VISIBLE_DEVICES=1 --moe-num-experts 8 --wandb-name "gojo/dm-moe-k8-top2-bal01"`

### Convergence Curves

| Epoch | K=4 val% | K=8 val% |
|-------|----------|----------|
| 1 | 60.96 | 61.86 |
| 10 | 26.51 | 27.89 |
| 30 | 15.11 | 15.47 |
| 60 | 9.36 | 9.68 |
| 90 | 7.06 | 6.83 |
| 120 | 5.96 | 5.74 |
| 136 | — | 5.35 (final) |
| 150 | 5.36 | — |
| 173 | 5.09 (final) | — |

### Routing Analysis

Routing was healthy — no expert collapse. The `balance_coeff=0.1` (10x higher than PR-suggested 0.01) plus jitter noise (std=1.0) and entropy penalty were necessary to prevent collapse. With 0.01, all tokens routed to a single expert per layer.

- K=4 routing entropy: ~1.37–1.39 (max=ln(4)=1.386) — near-uniform
- K=8 routing entropy: ~1.97–2.08 (max=ln(8)=2.079) — near-uniform
- Balance loss stayed near 4.0 throughout (theoretical minimum for top-2)

### What happened

**The hypothesis did not work.** Neither K=4 nor K=8 MoE beats the baseline.

The fundamental problem is **throughput**: MoE's loop-over-experts routing (required because each token goes to different experts) prevents `torch.compile` and adds ~1.7–2.2x wall-clock overhead per epoch. Within the 180-minute budget, K=4 gets 173 epochs and K=8 gets 136, vs the baseline's 511 epochs.

The per-epoch convergence rate is **not meaningfully faster** than baseline. Both K=4 and K=8 follow a similar loss curve as the dense model — MoE does not give a per-epoch learning advantage that could offset its throughput cost. The routing is well-balanced (entropy near maximum), but expert "specialization" by flow regime didn't emerge in a way that improved sample efficiency.

Even extrapolating the convergence curves, it would take ~400+ epochs for K=4 to reach 3.833%, at which point the dense model (which continues to improve past epoch 500) would still be ahead.

### Key observations

1. **balance_coeff=0.01 causes total expert collapse** with batch_size=1 and 50k correlated tokens. The balance loss gradient is too weak vs task loss (~5–9). Needed 0.1 to maintain routing diversity.
2. **K=4 slightly outperforms K=8** — fewer experts means less overhead and more epochs within budget, and the extra capacity from 8 experts doesn't help.
3. **No compile = no MoE in this setting.** The dynamic routing prevents torch.compile, which is critical for competitive throughput on this task.

### Suggested follow-ups

1. **Compile-friendly MoE**: Use fixed token-to-expert assignment (e.g., hash-based or position-based routing) that avoids dynamic masks and allows torch.compile. This would eliminate the throughput penalty while retaining some specialization.
2. **Expert-choice routing** (Zhou et al., 2022): Let experts choose tokens instead of tokens choosing experts — this can be implemented with fixed-size gather operations that compile well.
3. **Fewer MoE layers**: Only replace the FFN in the last 1-2 layers instead of all 4, reducing overhead while targeting where specialization might matter most.
4. **Grouped MoE**: Process experts in parallel using batched matrix multiplications instead of a loop, which could be compile-friendly.

tcapelle:
ADVISOR: Closing PR #3263 because ADVISOR: Closing — MoE FFN is a clear dead end for DM. K=4: 5.09% (+33%), K=8: 5.35% (+40%).

Two structural failures: (1) **Throughput penalty** — the loop-over-experts implementation can't be compiled, costing 1.7-2.2x per-epoch → only 136-173 epochs vs baseline's 511. This is the dominant factor. (2) **Routing collapse** — balance_coeff=0.1 (needed to prevent collapse at 0.01) forces near-uniform routing entropy. Layers 0-2 hit 99-100% max entropy. The experts never specialize by flow regime; the MoE is just an expensive ensemble average.

The per-epoch learning curve is parallel to the dense model, meaning MoE adds no sample-efficiency benefit — only overhead. The FFN is not the bottleneck for the remaining 3.833% error. Your observation that balance_coeff=0.01 causes full collapse at bs=1 is a useful finding. Expert-choice routing (Zhou et al. 2022) would address both issues but the zero per-epoch benefit suggests the MoE direction is unlikely to help..

PR #3264, opened 2026-04-24
DrivAerML: Weight Standardization on linear layers (bs=1 optimization)

tcapelle:
## Hypothesis

DrivAerML training uses **batch-size=1**, the most extreme small-batch scenario possible. With bs=1, batch statistics are maximally noisy and the loss landscape is rough, forcing the optimizer to navigate through high-variance gradient estimates. We hypothesize that **Weight Standardization (WS)** (Qiao et al., 2019, arXiv:1903.10520) will smooth the loss landscape and improve convergence by standardizing the weight matrices of all linear layers.

WS reparametrizes each weight matrix W as: `W_hat = (W - mean(W)) / std(W)`. This normalizes the weights to have zero mean and unit variance across the fan-in dimension, which:
1. **Smooths the loss landscape** — the Lipschitz constant of the loss is reduced, making gradient descent more stable
2. **Reduces sensitivity to learning rate** — the standardized weights create more uniform gradient magnitudes across layers
3. **Synergizes with LayerNorm** — WS on weights + LN on activations gives comprehensive normalization at both levels

In Big Transfer (BiT) models, WS was essential for achieving SOTA with GroupNorm at small batch sizes. The DM Transolver already uses LayerNorm but lacks weight-level normalization. At bs=1, WS may provide the stability improvement that allows the model to navigate deeper into the loss basin.

**Key difference from spectral norm (dead, #3224):** Spectral norm constrains the max singular value (σ_max=1), which over-restricts attention capacity. WS standardizes the full weight distribution without capping capacity — it reshapes the loss landscape without limiting what the model can express.

### Implementation

1. **Create a WeightStandardizedLinear module:**
   ```python
   class WSLinear(nn.Linear):
       def forward(self, x):
           # Standardize weights along fan_in dimension
           weight = self.weight
           mean = weight.mean(dim=1, keepdim=True)
           std = weight.std(dim=1, keepdim=True).clamp(min=1e-5)
           weight = (weight - mean) / std
           return F.linear(x, weight, self.bias)
   ```

2. **Add CLI flag:**
   ```python
   parser.add_argument('--weight-standardization', action='store_true', default=False,
                       help='Apply weight standardization to all linear layers in transformer')
   ```

3. **Replace nn.Linear with WSLinear in transformer layers** when `--weight-standardization` is enabled. Apply to:
   - Q, K, V projection matrices in self-attention
   - Output projection in self-attention  
   - Both linear layers in each FFN
   - Do NOT apply to the final output regression head (keep that standard)

   The simplest approach: after model construction, iterate through modules and swap:
   ```python
   if args.weight_standardization:
           if isinstance(module, nn.Linear) and 'output_head' not in name:
               # Replace with WSLinear (copy weights and bias)
               ws = WSLinear(module.in_features, module.out_features, bias=module.bias is not None)
               ws.weight = module.weight
               if module.bias is not None:
                   ws.bias = module.bias
               # Set the WSLinear in place of the original
               parent = dict(model.named_modules())[name.rsplit('.', 1)[0] if '.' in name else '']
               setattr(parent, name.split('.')[-1], ws)
   ```
   
   **Alternative (simpler):** Just add the standardization directly in the model's forward pass for QKV and FFN layers, wrapped in an `if args.weight_standardization:` check. Either approach works — pick whichever is cleaner for the codebase.

### Trials

**Smoke test first (3 epochs)** — verify loss is finite, no shape errors, gradients flow.

**Trial 1 — WS on all transformer linear layers (champion config):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --weight-standardization \
  --wandb_group dm-weight-standardization
```

**Trial 2 — WS + slightly higher LR (6e-4):**
WS smooths the loss landscape, which often allows higher learning rates. Since the LR neighborhood above 5e-4 was previously dead WITHOUT WS, this tests whether WS unlocks that region:
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 6e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --weight-standardization \
  --wandb_group dm-weight-standardization
```

## Baseline

| Metric | Value | Epoch |
|--------|-------|-------|

- W&B run: `ncl1dh88`

Reproduce:
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -86,6 +86,45 @@ class EMAWithWarmup:
         self.backup = None
 
 
+class WSLinear(torch.nn.Linear):
+    """Weight Standardization applied to nn.Linear (Qiao et al., 2019)."""
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        weight = self.weight
+        mean = weight.mean(dim=1, keepdim=True)
+        std = weight.std(dim=1, keepdim=True).clamp(min=1e-5)
+        weight = (weight - mean) / std
+        return F.linear(x, weight, self.bias)
+
+
+def apply_weight_standardization(model: torch.nn.Module) -> int:
+    """Replace nn.Linear with WSLinear in transformer layers, skipping output heads."""
+    exclude = {"out.project", "surface_head"}
+    count = 0
+    for name, module in list(model.named_modules()):
+        if any(ex in name for ex in exclude):
+            continue
+        if not isinstance(module, torch.nn.Linear) or isinstance(module, WSLinear):
+            continue
+        ws = WSLinear(module.in_features, module.out_features, bias=module.bias is not None)
+        ws.weight = module.weight
+        if module.bias is not None:
+            ws.bias = module.bias
+        parts = name.rsplit(".", 1)
+        if len(parts) == 2:
+            parent_name, attr_name = parts
+            parent = dict(model.named_modules())[parent_name]
+        else:
+            parent = model
+            attr_name = parts[0]
+        if isinstance(parent, torch.nn.Sequential):
+            parent[int(attr_name)] = ws
+        else:
+            setattr(parent, attr_name, ws)
+        count += 1
+    return count
+
+
 LEGACY_VAL_ALIAS = {
     "val_in_dist": "p_in",
     "val_ood_cond": "p_oodc",
@@ -169,6 +208,7 @@ class TrainConfig:
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
+    weight_standardization: bool = False
 
 
 @dataclass
@@ -1615,6 +1655,9 @@ def main() -> None:
 
     train_loader, val_loaders, test_loaders = build_loaders(config, bundle, num_workers=resolved_num_workers)
     model = build_model(config, bundle).to(device)
+    if config.weight_standardization:
+        ws_count = apply_weight_standardization(model)
+        print(f"[WS] Replaced {ws_count} Linear layers with WSLinear")
     forward_model = torch.compile(model) if config.compile_model and device.type == "cuda" else model
     anp_head = None
     if bundle.spec.name == "tandemfoilset" and config.anp_srf:


tcapelle:
STUDENT vegeta:

## Results

### Summary

**Negative result.** Weight Standardization (WS) introduces catastrophic gradient instability when combined with per-batch cosine-restart scheduling at batch-size=1. Neither trial converges meaningfully — both are ~18x worse than baseline.

### Metrics

| Metric | Baseline | Trial 1 (lr=5e-4) | Trial 2 (lr=6e-4) |
|---|---|---|---|
| **val_primary/surface_rel_l2_pct** | **3.833%** (ep 511) | 67.925% (ep 10) | 71.410% (ep 8) |
| **test_primary/surface_rel_l2_pct** (best ckpt) | — | 70.891% | 71.586% |
| test_primary/surface_rel_l2_pct (final) | — | 74.612% | 92.168% |
| Total epochs | 511+ | 167 (timeout) | 167 (timeout) |
| Best epoch | 511 | 10 | 8 |
| Grad norm (final epoch) | — | **Infinity** | **9.4e16** |
| Grad clip events per epoch | — | 394/394 (100%) | 394/394 (100%) |

- **Baseline**: 3.833% val, epoch 511, run `ncl1dh88`
- **Literature reference**: DrivAerML test target ≈ 2.7% (Bonnet et al.)

### W&B Runs

- Trial 1: [`x0biqh3s`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/x0biqh3s) — `vegeta/dm-ws-lr5e4`
- Trial 2: [`i7nvv6gj`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/i7nvv6gj) — `vegeta/dm-ws-lr6e4`

### Commands

```bash
# Trial 1
CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=120 python train.py \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --weight-standardization \
  --wandb-group dm-weight-standardization --agent vegeta --wandb-name "vegeta/dm-ws-lr5e4"

# Trial 2 (same but --lr 6e-4, --wandb-name "vegeta/dm-ws-lr6e4")
```

### What happened

WS reparameterizes each linear layer's forward pass as `W_hat = (W - mean) / std` along the fan-in dimension. While the implementation is correct (33 layers replaced, smoke test passed, losses were finite early on), WS interacts catastrophically with our training setup:

1. **Gradient explosion**: Grad norms grew from ~800K (epoch 1) to billions within 5 epochs, eventually reaching Infinity (Trial 1) and 9.4e16 (Trial 2) by epoch 167. Every single batch triggered grad clipping (394/394 per epoch).

2. **No convergence**: Best val metrics froze at epoch 10 (Trial 1) and epoch 8 (Trial 2) — never improved again over 157+ subsequent epochs. The model is essentially stuck at ~70% relative error, oscillating wildly between 67% and 248%.

3. **Root cause — WS + per-batch cosine restarts at bs=1**: WS normalizes weights to zero-mean unit-variance each forward pass, creating an implicit constraint. Combined with per-batch cosine restarts (T_max=30 steps = ~0.08 epochs), the optimizer's learning rate swings from 0 to lr_max every 30 batches. Each restart jolts the unnormalized weights, and WS amplifies the disruption by re-centering — creating a feedback loop of ever-growing gradients. At bs=1 there's no mini-batch averaging to dampen this effect.

4. **WS was designed for GroupNorm/no-BN regimes with standard LR schedules**, not aggressive per-batch cosine restarts. The original paper (Qiao et al., 2019) used standard epoch-level scheduling.

### Suggested follow-ups

1. **WS + epoch-level cosine schedule**: If WS is revisited, use `--cosine-t-max` equal to total training steps (or a large fraction), not 30. This removes the restart instability.
2. **WS + lower LR (1e-4 or 5e-5)**: Much lower LR might keep gradients in check, but the per-batch restarts would still be problematic.
3. **WS on attention projections only** (skip FFN): Limiting WS scope might reduce the instability surface, but doesn't address the core LR schedule interaction.
4. **Centered Weight Normalization** instead of WS: CWN (Huang et al., 2017) only divides by the norm without centering — may avoid the mean-shift instability.

tcapelle:
ADVISOR: Closing PR #3264 because ADVISOR: Closing — Weight Standardization is catastrophically incompatible with the DM training setup. lr=5e-4: 67.9% best (18x worse), lr=6e-4: 71.4% best (19x worse). Both had 100% gradient clipping and grad_norm → Infinity.

Root cause: WS reparameterizes weights every forward pass. Combined with T_max=30 cosine restarts (LR peaks every 30 batches) at bs=1, this creates a feedback loop — each LR spike jolts the raw weights, WS recenters them, but AdamW's momentum was built on pre-WS effective weights, creating persistent direction mismatch. The original BiT paper used epoch-level LR decay, not per-batch restarts.

Your analysis of the three-way interaction (WS × restarts × bs=1) was exactly right..

PR #3265, opened 2026-04-24
DrivAerML: hidden-space noise regularization (post-Fourier perturbation)

tcapelle:
## Hypothesis

Your coordinate noise experiment (#3256) revealed that Fourier positional encoding amplifies input-space perturbations by up to 32x, making coordinate-level augmentation structurally incompatible with the DM architecture. However, the **underlying regularization motivation is sound** — injecting controlled noise during training forces the model to learn smoother, more robust representations. The fix is to apply noise AFTER the Fourier encoding and input projection, directly in the hidden representation space where the scale is well-controlled.

**Hidden-space noise** (also called "representation perturbation" or "R-Drop without KL") adds Gaussian noise with small standard deviation to the hidden representations between transformer layers during training only. This:
1. **Avoids Fourier amplification entirely** — the noise is in the 512-dim hidden space, not the coordinate space
2. **Acts as implicit regularization** — similar to dropout but continuous and differentiable, preserving smooth gradient flow
3. **Encourages flat loss basins** — the model must produce correct predictions despite small perturbations, selecting for solutions that are robust to representation noise

This is related to "Noisy Nodes" (Godwin et al., ICLR 2022) which showed that adding noise to node features in GNNs for molecular property prediction improved generalization by 10-15%. The key is that hidden-space noise is scale-appropriate by construction — the perturbation magnitude is relative to the activation norms, not amplified by encoding frequencies.

### Implementation

1. **Add CLI flags:**
   ```python
   parser.add_argument('--hidden-noise-std', type=float, default=0.0,
                       help='Std of Gaussian noise added to hidden representations during training (0=disabled)')
   parser.add_argument('--hidden-noise-layers', type=str, default='all',
                       help='Which layers to apply noise: "all", "first", "last", or comma-separated indices like "0,1"')
   ```

2. **Add noise injection in the transformer forward pass.** After each transformer layer's output (post-residual, post-LayerNorm), during training only:
   ```python
   if self.training and args.hidden_noise_std > 0:
       noise = torch.randn_like(x) * args.hidden_noise_std
       x = x + noise
   ```

   Place this **after** each transformer layer's full computation (self-attention + FFN + residual + LayerNorm) but **before** the next layer. Apply only during training (the `self.training` check handles this automatically via `model.train()` / `model.eval()`).

3. **Layer selection logic** (if implementing `--hidden-noise-layers`):
   ```python
   if args.hidden_noise_layers == 'all':
       apply_noise = [True] * num_layers
   elif args.hidden_noise_layers == 'first':
       apply_noise = [True] + [False] * (num_layers - 1)
   elif args.hidden_noise_layers == 'last':
       apply_noise = [False] * (num_layers - 1) + [True]
   else:
       indices = set(int(i) for i in args.hidden_noise_layers.split(','))
       apply_noise = [i in indices for i in range(num_layers)]
   ```

### Trials

**Smoke test first (3 epochs)** — verify loss is finite, no NaN, gradients normal.

**Trial 1 — σ=0.01, all layers (moderate noise):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --hidden-noise-std 0.01 \
  --hidden-noise-layers all \
  --wandb_group dm-hidden-noise-reg
```

**Trial 2 — σ=0.05, all layers (stronger noise):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --hidden-noise-std 0.05 \
  --hidden-noise-layers all \
  --wandb_group dm-hidden-noise-reg
```

**Trial 3 — σ=0.02, last layer only (targeted late-stage regularization):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --hidden-noise-std 0.02 \
  --hidden-noise-layers last \
  --wandb_group dm-hidden-noise-reg
```

Report `val_primary/surface_rel_l2_pct` for all trials. Target: beat **3.833%**.

## Baseline

| Metric | Value | Epoch |
|--------|-------|-------|

- W&B run: `ncl1dh88`

Reproduce:
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -169,6 +169,8 @@ class TrainConfig:
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
+    hidden_noise_std: float = 0.0
+    hidden_noise_layers: str = "all"
 
 
 @dataclass
@@ -540,6 +542,32 @@ def build_model(config: TrainConfig, bundle: DatasetBundle) -> torch.nn.Module:
             volume_output_dim=bundle.spec.volume_output_dim,
         )
     raise ValueError(f"Unknown model: {config.model}")
+
+
+def register_hidden_noise(model: torch.nn.Module, config: TrainConfig):
+    if config.hidden_noise_std <= 0:
+        return
+    backbone = getattr(model, "backbone", None)
+    if backbone is None:
+        return
+    blocks = list(backbone.blocks)
+    num_layers = len(blocks)
+    if config.hidden_noise_layers == "all":
+        apply_at = [True] * num_layers
+    elif config.hidden_noise_layers == "first":
+        apply_at = [True] + [False] * (num_layers - 1)
+    elif config.hidden_noise_layers == "last":
+        apply_at = [False] * (num_layers - 1) + [True]
+    else:
+        indices = {int(i) for i in config.hidden_noise_layers.split(",")}
+        apply_at = [i in indices for i in range(num_layers)]
+    std = config.hidden_noise_std
+    for i, block in enumerate(blocks):
+        if apply_at[i]:
+            def _hook(mod, inp, out, _std=std):
+                if mod.training:
+                    return out + torch.randn_like(out) * _std
+            block.register_forward_hook(_hook)
 
 
 def build_optimizer(params, config: TrainConfig):
@@ -1615,6 +1643,7 @@ def main() -> None:
 
     train_loader, val_loaders, test_loaders = build_loaders(config, bundle, num_workers=resolved_num_workers)
     model = build_model(config, bundle).to(device)
+    register_hidden_noise(model, config)
     forward_model = torch.compile(model) if config.compile_model and device.type == "cuda" else model
     anp_head = None
     if bundle.spec.name == "tandemfoilset" and config.anp_srf:


tcapelle:
STUDENT alphonse:

## Results

### Summary

Hidden-space noise regularization was implemented via forward hooks on transformer blocks. All three trials eventually **diverged** during training, and **none beat the baseline**. The hypothesis that hidden-space noise would improve generalization did not hold — instead, it introduced training instability that eventually caused gradient explosions in all configurations.

### Metrics

| Trial | Config | Best Epoch | val_primary/surface_rel_l2_pct | test_primary/surface_rel_l2_pct | Diverged At |
|-------|--------|-----------|-------------------------------|--------------------------------|-------------|
| **Baseline** | No noise | 511 | **3.833%** | **4.685%** | N/A |
| Trial 1 | σ=0.01, all layers | 94 | 7.491% | 8.393% | ~epoch 95 |
| Trial 2 | σ=0.05, all layers | 381 | 4.236% | 5.064% | ~epoch 400 |
| Trial 3 | σ=0.02, last layer only | 242 | 4.693% | 5.435% | ~epoch 290 |

### Commands

**Trial 1 (σ=0.01, all layers):**
```bash
  --ema-decay 0.9995 --hidden-noise-std 0.01 --hidden-noise-layers all \
  --wandb-group dm-hidden-noise-reg --agent alphonse --wandb-name "alphonse/dm-hidden-noise-std001-all"
```

**Trial 2 (σ=0.05, all layers):**
Same as Trial 1 with `--hidden-noise-std 0.05` on GPU 1.

**Trial 3 (σ=0.02, last layer only):**
Same as Trial 1 with `--hidden-noise-std 0.02 --hidden-noise-layers last` on GPU 2.

### Peak Memory
~6 GB VRAM per GPU (same as baseline — noise adds negligible overhead).

### W&B Run IDs
- Trial 1: `81vada2j`
- Trial 2: `fbek5zpt`
- Trial 3: `uxmswfej`

### What happened

All three trials trained normally through early epochs but eventually suffered gradient explosions that caused irreversible divergence:

1. **Trial 1 (σ=0.01, all)** diverged earliest at epoch ~95. Grad norm spiked from 0.5 → 9.6 → 67 → 355 → Infinity within a few epochs during a cosine LR restart. The model never recovered.

2. **Trial 3 (σ=0.02, last)** diverged at epoch ~290 with a similar pattern — grad norms went to Infinity during a cosine restart cycle.

3. **Trial 2 (σ=0.05, all)** lasted longest (381 epochs), reaching val 4.24% before also diverging. This is the closest to the baseline but still ~0.4% worse.

**Counterintuitive finding:** Stronger noise (σ=0.05) provided MORE stability, not less. The weaker noise trials diverged earlier. This suggests the noise wasn't acting as regularization in the intended way — instead, the stochastic perturbations may have been interacting pathologically with the cosine annealing restarts, and only the strongest noise kept the model in a sufficiently wide basin to survive longer.

**Root cause of divergence:** The noise injection after each transformer block compounds through the network (4 layers = 4 independent noise injections per forward pass). At cosine LR restarts, when the learning rate jumps back to maximum, the combination of high LR + accumulated noise perturbations appears to push the model out of stable basins. The EMA model (0.9995 decay) initially smooths this, but once the raw model diverges, the EMA eventually follows.

### Suggested follow-ups

1. **Noise annealing** — decay the noise std alongside the cosine LR schedule (e.g., σ_t = σ_0 * (1 - lr_t/lr_max)). This would eliminate noise at high LR and add it only during the fine-tuning phases where regularization matters most.

2. **Noise only during low-LR phases** — enable noise only when lr < 0.5 * lr_max, completely avoiding the high-LR restart regions that caused divergence.

3. **Gradient-aware noise scaling** — scale noise by 1/grad_norm to automatically reduce perturbations when gradients are large.

4. **Dropout comparison** — standard dropout at matching effective rates (p ≈ σ²/(σ² + activation_variance)) may achieve the same regularization benefit without the divergence risk, since dropout's binary mask is more well-studied with standard optimizers.

tcapelle:
ADVISOR: Closing PR #3265 because ADVISOR: Closing — hidden-space noise best val=4.236% is 14% worse than 3.700% baseline. All three noise configs (σ=0.01/0.05/0.02) diverged. Noise destabilizes gradient flow through transformer hidden states. Alphonse reassigned..

PR #3267, opened 2026-04-24
DrivAerML: SwiGLU FFN activation (replace GELU with gated linear unit)

tcapelle:
## Hypothesis

The DM Transolver uses standard GELU activation in its FFN sublayers. **SwiGLU** (Shazeer, 2020; arXiv:2002.05202) replaces the activation with a gated linear unit: instead of FFN(x) = W2 · GELU(W1 · x), SwiGLU computes FFN(x) = W2 · (SiLU(W_gate · x) ⊙ (W_up · x)), where ⊙ is element-wise multiplication. This adds a learned gating mechanism that selectively activates features.

SwiGLU has become the standard FFN activation in modern LLMs (LLaMA, Mistral, Gemma) because it consistently outperforms GELU across model scales. The key properties:
1. **Multiplicative gating** — the gate branch learns which features to amplify vs suppress, providing richer nonlinear expressivity per parameter
2. **Smoother gradients** — SiLU (swish) activation produces smoother gradient landscapes than GELU
3. **Proven at scale** — not just an LLM trick; SwiGLU shows improvements in vision (MetaFormer), audio, and structured prediction tasks

For DM where the FFN processes 50k surface point tokens independently, the gating mechanism could help the model learn context-dependent feature transformations — amplifying relevant pressure/velocity features while suppressing noise, which directly addresses the systematic spatial bias.

**Important:** SwiGLU uses 3 weight matrices (gate, up, down) instead of 2 (up, down). To keep parameter count roughly equal, the FFN hidden dim should be reduced by 2/3. With the current 512d model, the standard FFN dim is likely 2048 (4x expansion). For SwiGLU, use hidden_dim = 1365 (≈ 2048 × 2/3) to maintain parameter parity.

### Implementation

1. **Add CLI flag:**
   ```python
   parser.add_argument('--swiglu-ffn', action='store_true', default=False,
                       help='Replace GELU FFN with SwiGLU activation in transformer layers')
   ```

2. **Implement SwiGLU FFN module:**
   ```python
   class SwiGLUFFN(nn.Module):
       def __init__(self, dim, hidden_dim=None, bias=True):
           super().__init__()
           hidden_dim = hidden_dim or int(dim * 8 / 3)  # 2/3 of standard 4x expansion
           # Round to nearest multiple of 8 for hardware efficiency
           hidden_dim = ((hidden_dim + 7) // 8) * 8
           self.w_gate = nn.Linear(dim, hidden_dim, bias=bias)
           self.w_up = nn.Linear(dim, hidden_dim, bias=bias)
           self.w_down = nn.Linear(hidden_dim, dim, bias=bias)
       
       def forward(self, x):
           return self.w_down(F.silu(self.w_gate(x)) * self.w_up(x))
   ```

3. **Replace the standard FFN** in each transformer layer when `--swiglu-ffn` is enabled:
   ```python
   if args.swiglu_ffn:
       ffn = SwiGLUFFN(hidden_dim)
   else:
       ffn = existing_ffn  # keep current GELU FFN
   ```

### Trials

**Smoke test first (3 epochs)** — verify shapes, loss finite, gradients normal.

**Trial 1 — SwiGLU, parameter-matched (default hidden dim):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --swiglu-ffn \
  --wandb_group dm-swiglu-ffn
```

**Trial 2 — SwiGLU + slightly lower LR (4.5e-4):**
SwiGLU's smoother gradients may prefer a slightly different LR. Test the lower end since higher LR already failed on the standard FFN:
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 4.5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --swiglu-ffn \
  --wandb_group dm-swiglu-ffn
```

## Baseline

| Metric | Value | Epoch |
|--------|-------|-------|

- W&B run: `ncl1dh88`

Reproduce:
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/core/architectures/transolver_reference.py
+++ b/target/icml2026/core/architectures/transolver_reference.py
@@ -84,6 +84,20 @@ class UpActDownMlp(nn.Module):
         return self.fc2(self.act(self.fc1(x)))
 
 
+class SwiGLUFFN(nn.Module):
+    def __init__(self, hidden_dim: int, mlp_hidden_dim: int):
+        super().__init__()
+        swiglu_hidden = int(mlp_hidden_dim * 2 / 3)
+        swiglu_hidden = ((swiglu_hidden + 63) // 64) * 64
+        self.w_gate = nn.Linear(hidden_dim, swiglu_hidden, bias=False)
+        self.w_up = nn.Linear(hidden_dim, swiglu_hidden, bias=False)
+        self.w_down = nn.Linear(swiglu_hidden, hidden_dim, bias=False)
+        self.apply(_init_linear)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.w_down(F.silu(self.w_gate(x)) * self.w_up(x))
+
+
 class TransolverAttention(nn.Module):
     def __init__(self, hidden_dim: int, num_heads: int, num_slices: int, dropout: float = 0.0):
         super().__init__()
@@ -139,6 +153,7 @@ class TransformerBlock(nn.Module):
         mlp_expansion_factor: int | float,
         num_slices: int,
         dropout: float = 0.0,
+        swiglu_ffn: bool = False,
     ):
         super().__init__()
         mlp_hidden_dim = int(math.ceil(hidden_dim * mlp_expansion_factor))
@@ -150,7 +165,10 @@ class TransformerBlock(nn.Module):
             dropout=dropout,
         )
         self.norm2 = nn.LayerNorm(hidden_dim, eps=1e-6)
-        self.mlp = UpActDownMlp(hidden_dim=hidden_dim, mlp_hidden_dim=mlp_hidden_dim)
+        if swiglu_ffn:
+            self.mlp = SwiGLUFFN(hidden_dim=hidden_dim, mlp_hidden_dim=mlp_hidden_dim)
+        else:
+            self.mlp = UpActDownMlp(hidden_dim=hidden_dim, mlp_hidden_dim=mlp_hidden_dim)
 
     def forward(self, x: torch.Tensor, attn_mask: torch.Tensor | None = None) -> torch.Tensor:
         x = x + self.attention(self.norm1(x), attn_mask=attn_mask)
@@ -167,6 +185,7 @@ class Transformer(nn.Module):
         mlp_expansion_factor: int | float,
         num_slices: int,
         dropout: float = 0.0,
+        swiglu_ffn: bool = False,
     ):
         super().__init__()
         self.blocks = nn.ModuleList(
@@ -177,6 +196,7 @@ class Transformer(nn.Module):
                     mlp_expansion_factor=mlp_expansion_factor,
                     num_slices=num_slices,
                     dropout=dropout,
+                    swiglu_ffn=swiglu_ffn,
                 )
                 for _ in range(depth)
             ]
@@ -205,6 +225,7 @@ class ReferenceTransolver(nn.Module):
         n_head: int = 3,
         mlp_ratio: int = 4,
         slice_num: int = 96,
+        swiglu_ffn: bool = False,
     ):
         super().__init__()
         self.space_dim = space_dim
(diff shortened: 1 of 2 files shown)


tcapelle:
STUDENT mitsuha:

## Results

### Trial 1 — SwiGLU, lr=5e-4 (parameter-matched)

| Metric | Value | Epoch | Baseline |
|--------|-------|-------|----------|
| val_primary/surface_rel_l2_pct | **4.454%** | 228 | **3.833%** |
| test_primary/surface_rel_l2_pct | **5.377%** | 228 | **4.685%** |

- **W&B run:** `uc5c2hzw`
- **Command:**
  ```
    --epochs 999 --batch-size 1 \
    --ema-decay 0.9995 --swiglu-ffn --wandb-group dm-swiglu-ffn
  ```
- **Peak memory:** ~6.4 GB
- **Total epochs:** ~497 (stopped by timeout)
- **Divergence:** Training diverged catastrophically after epoch ~240 with Infinity gradient norms. EMA preserved best checkpoint at epoch 228.

### Trial 2 — SwiGLU, lr=4.5e-4

| Metric | Value | Epoch | Baseline |
|--------|-------|-------|----------|
| val_primary/surface_rel_l2_pct | **4.720%** | 196 | **3.833%** |
| test_primary/surface_rel_l2_pct | **5.633%** | 196 | **4.685%** |

- **W&B run:** `7lgw2wji`
- **Command:** Same as Trial 1 but with `--lr 4.5e-4`
- **Peak memory:** ~6.3 GB
- **Total epochs:** ~508 (stopped by timeout)
- **Divergence:** Also diverged catastrophically (Infinity gradients) around epoch ~280. EMA preserved best checkpoint at epoch 196.

### What happened

**SwiGLU did not beat the baseline.** Both trials showed healthy early convergence but diverged catastrophically in the mid-training region, well before reaching the baseline's 511 epochs.

Key observations:
1. **Early convergence was promising** — both trials tracked a smooth improvement curve through ~200 epochs, reaching 4.45-4.72% val.
2. **Catastrophic divergence** — both runs hit Infinity gradient norms and never recovered. Trial 1 (lr=5e-4) diverged earlier (~epoch 240) than Trial 2 (lr=4.5e-4, ~epoch 280), suggesting LR sensitivity.
3. **Parameter count** — SwiGLU with 2/3 hidden dim scaling (1408 hidden, rounded to 64 for GPU efficiency) and no bias (following LLaMA/Mistral convention) maintains parameter parity with the standard GELU FFN.
4. **The gating mechanism may interact badly with the cosine LR cycling** — the divergence consistently happened at high-LR phases of the cosine schedule, suggesting the SwiGLU gate gradients amplify instability when the LR peaks.

### Research notes

- Used `bias=False` for all SwiGLU linear layers (standard practice per LLaMA, Mistral, PaLM 2; supported by Post-Gating Bias literature showing pre-gate biases lose affine freedom after elementwise product).
- Hidden dim rounded to multiple of 64 for GPU efficiency: `int(2048 * 2/3) → 1365 → 1408`.
- No SwiGLU-specific initialization required — standard truncated normal used via `_init_linear`.

### Suggested follow-ups

1. **Lower gradient clipping** (e.g., 0.25 instead of 0.5) could stabilize the SwiGLU training through cosine LR peaks.
2. **Reduce cosine T_max** (e.g., 15-20) so LR peaks are less aggressive when the model is in a more fragile region.
3. **Try SwiGLU + lower LR** (3e-4 or 2e-4) — both trials suggest SwiGLU may need a fundamentally lower LR to avoid divergence in the later training phase.
4. **Weight decay** — the current config uses no weight decay with AdamW. Adding mild WD (1e-5) could help regularize the gating weights.

tcapelle:
ADVISOR: Closing PR #3267 because ADVISOR: Closing — SwiGLU val=4.454% is 20% worse than 3.700% baseline. Both trials diverged with Infinity gradient norms. SwiGLU's multiplicative gating amplifies gradient instability at cosine restarts. Mitsuha reassigned to DM metric-aware sweep..

PR #3269, opened 2026-04-24
DrivAerML: learned error correction head (self-correcting predictions)

tcapelle:
## Hypothesis

The remaining DM error at 3.833% is **systematic spatial bias** — the model consistently over/under-predicts in specific geometric regions (sharp junctions, underbody, A-pillar). This means the prediction errors have learnable spatial structure. We hypothesize that a **secondary correction head** trained to predict and subtract the model's own errors can exploit this structure to reduce the final prediction error.

The approach is inspired by **gradient boosting** and **iterative refinement**: the primary head makes a coarse prediction, then a correction head learns the residual error pattern from the same latent representation and adds a correction term. During training, the correction head's target is the primary head's detached prediction error (target - primary_pred). At inference, the final prediction is primary_pred + correction.

This is fundamentally different from residual prediction (dead for DM — that predicted corrections to freestream). Here, the correction head operates on the model's OWN intermediate representation and corrects the model's OWN errors, not corrections to a physics prior. The primary head provides the scaffold; the correction head refines it.

**Key design choice:** The correction head must be lightweight (1-2 linear layers) to avoid capacity competition with the primary head. And the primary head's prediction must be DETACHED when computing the correction target — otherwise the gradients would fight.

### Implementation

1. **Add CLI flag:**
   ```python
   parser.add_argument('--error-correction-head', action='store_true', default=False,
                       help='Add a secondary head that learns to correct primary prediction errors')
   parser.add_argument('--correction-loss-weight', type=float, default=0.1,
                       help='Weight for the correction head loss')
   ```

2. **Add correction head to the model** (in `__init__`):
   ```python
   if args.error_correction_head:
       # Lightweight MLP: takes latent features, outputs correction per channel
       self.correction_head = nn.Sequential(
           nn.Linear(hidden_dim, hidden_dim // 4),
           nn.GELU(),
           nn.Linear(hidden_dim // 4, output_dim)
       )
   ```

3. **Modify the forward pass** to produce both primary and correction predictions:
   ```python
   primary_pred = self.output_head(latent_features)  # existing prediction
   
   if hasattr(self, 'correction_head'):
       correction = self.correction_head(latent_features)
       final_pred = primary_pred + correction
   else:
       final_pred = primary_pred
   ```

4. **Compute the correction loss** (in the training loss computation):
   ```python
   # Primary loss (on final corrected prediction)
   primary_loss = mse_loss(final_pred, target)
   
   if args.error_correction_head:
       # Correction target: the primary head's error (detached!)
       with torch.no_grad():
           primary_error = target - primary_pred.detach()
       correction_loss = mse_loss(correction, primary_error)
       total_loss = primary_loss + args.correction_loss_weight * correction_loss
   else:
       total_loss = primary_loss
   ```

   **Critical:** `primary_pred.detach()` in the correction target — this prevents the correction loss from back-propagating through the primary head, which would create conflicting gradient signals.

5. **At evaluation/test time:** use `final_pred = primary_pred + correction` as the prediction. No changes needed since the forward pass already produces this.

### Trials

**Smoke test first (3 epochs)** — verify shapes, both losses finite, correction head producing non-zero output.

**Trial 1 — correction weight=0.1 (moderate correction signal):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --error-correction-head \
  --correction-loss-weight 0.1 \
  --wandb_group dm-error-correction-head
```

**Trial 2 — correction weight=0.5 (stronger correction signal):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --error-correction-head \
  --correction-loss-weight 0.5 \
  --wandb_group dm-error-correction-head
```

Log the correction head's loss separately (correction_loss) — if it collapses to near-zero quickly, the primary head is already doing all the work and the correction head adds nothing. If it stays elevated, the correction head is actively learning residual patterns.

## Baseline

| Metric | Value | Epoch |
|--------|-------|-------|

- W&B run: `ncl1dh88`

Reproduce:
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -86,6 +86,21 @@ class EMAWithWarmup:
         self.backup = None
 
 
+class ErrorCorrectionHead(torch.nn.Module):
+    def __init__(self, hidden_dim: int, output_dim: int):
+        super().__init__()
+        self.net = torch.nn.Sequential(
+            torch.nn.Linear(hidden_dim + output_dim, hidden_dim // 4),
+            torch.nn.GELU(),
+            torch.nn.Linear(hidden_dim // 4, output_dim),
+        )
+        torch.nn.init.zeros_(self.net[-1].weight)
+        torch.nn.init.zeros_(self.net[-1].bias)
+
+    def forward(self, hidden: torch.Tensor, primary_pred_detached: torch.Tensor) -> torch.Tensor:
+        return self.net(torch.cat([hidden, primary_pred_detached], dim=-1))
+
+
 LEGACY_VAL_ALIAS = {
     "val_in_dist": "p_in",
     "val_ood_cond": "p_oodc",
@@ -169,6 +184,8 @@ class TrainConfig:
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
+    error_correction_head: bool = False
+    correction_loss_weight: float = 0.1
 
 
 @dataclass
@@ -860,10 +877,13 @@ def evaluate_grouped(
     *,
     amp_mode: str = "none",
     max_batches: int = 0,
+    correction_head: ErrorCorrectionHead | None = None,
 ) -> dict[str, float]:
     model.eval()
     if anp_head is not None:
         anp_head.eval()
+    if correction_head is not None:
+        correction_head.eval()
 
     dataset_name: str | None = None
     total_surface = 0.0
@@ -931,6 +951,10 @@ def evaluate_grouped(
                 volume_x=batch.volume_x,
                 volume_mask=batch.volume_mask,
             )
+            if correction_head is not None and outputs.get("surface_hidden") is not None and outputs["surface_preds"] is not None:
+                primary_pred = outputs["surface_preds"]
+                correction = correction_head(outputs["surface_hidden"], primary_pred)
+                outputs["surface_preds"] = primary_pred + correction * batch.surface_mask.unsqueeze(-1)
         outputs = float_outputs(outputs)
         if dataset_name == "airfrans":
             if not isinstance(transform, TargetTransform):
@@ -1277,10 +1301,15 @@ def train_one_epoch(
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
     volume_loss_weight: float = 1.0,
+    correction_head: ErrorCorrectionHead | None = None,
+    correction_loss_weight: float = 0.0,
+    correction_ema: EMAWithWarmup | None = None,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
         anp_head.train()
+    if correction_head is not None:
+        correction_head.train()
     running = {"loss": 0.0}
     steps = 0
     micro_batches_total = 0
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT emma:

## Results

### Trial 1 — correction_loss_weight=0.1

| Metric | Value |
|--------|-------|
| **val_primary/surface_rel_l2_pct** | **9.169%** (epoch 79) |
| test surface_rel_l2_pct (best ckpt) | 12.864% |
| correction_loss (final) | 0.71 (diverged) |
| Total epochs | 225 (diverged at ~epoch 80, never recovered) |
| W&B run | `vvh651d8` |

**Status: DIVERGED.** Training became unstable around epoch 80 — grad norms exploded to Infinity and the model never recovered. The lighter correction weight (0.1) appeared insufficient to stabilize the correction head's interaction with the primary loss.

**Command:**
```bash
CUDA_VISIBLE_DEVICES=2 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=180 python train.py \
  --ema-decay 0.9995 --error-correction-head --correction-loss-weight 0.1 \
  --agent emma --wandb-name "emma/dm-error-correction-w0.1" --wandb-group dm-error-correction-head
```

### Trial 2 — correction_loss_weight=0.5

| Metric | Value |
|--------|-------|
| **val_primary/surface_rel_l2_pct** | **4.700%** (epoch 227) |
| test surface_rel_l2_pct (best ckpt) | 5.535% |
| test surface_rel_l2_pct (final) | 5.497% |
| correction_loss (final) | 0.006 |
| Total epochs | 228 (hit 180-min timeout) |
| W&B run | `riq65o1p` |

**Command:**
Same as Trial 1 but with `--correction-loss-weight 0.5`.

### Baseline comparison

| | Baseline (ncl1dh88) | Trial 1 (w=0.1) | Trial 2 (w=0.5) |
|--|--|--|--|
| val surface_rel_l2_pct | **3.833%** (ep 511) | 9.169% (ep 79, diverged) | 4.700% (ep 227) |
| test surface_rel_l2_pct | 4.685% | 12.864% | 5.535% |

## What happened

**Neither trial beat the baseline, but Trial 2 shows promise.**

1. **Trial 1 (w=0.1) diverged.** Around epoch 80, grad norms exploded to Infinity and training collapsed. The lighter correction weight didn't provide enough stabilization. The correction head's gradients likely conflicted with the primary head's gradients at a cosine cycle restart, and without sufficient correction loss magnitude, the optimizer couldn't recover.

2. **Trial 2 (w=0.5) converged stably** to 4.700% val in 227 epochs. This is ~0.87pp above the 3.833% baseline, but the baseline ran to epoch 511 and Trial 2 was cut short at 228 epochs by the 180-min timeout. The convergence trajectory at epoch 227 was still improving.

3. **The correction head's loss stayed active** throughout training for Trial 2 (0.006 at end), indicating it was continuously learning residual error patterns rather than collapsing — a positive sign for the hypothesis.

4. **The epoch-matched comparison is unfavorable**: at ~200 epochs, the previous baseline experiments (from PR #3258 aux normals) reached ~4.66-4.72%, and Trial 2 here is at 4.70% — essentially identical. The correction head isn't adding measurable improvement at this epoch count.

## Implementation notes

- The correction head takes `[latent_features, primary_pred.detach()]` as input (concat), following research recommendations that conditioning on the primary prediction helps the correction head learn where errors occur.
- Zero-initialized final layer ensures corrections start from zero and grow gradually.
- The correction is applied at both train and eval time: `final_pred = primary_pred + correction`.
- Correction loss target: `(target - primary_pred.detach())`, trained with MSE.

## Suggested follow-ups

1. **Extended training** — Trial 2 needs the full 500+ epoch budget to determine if the correction head provides a real advantage over the baseline at convergence.
2. **Lower correction weight with warmup** — A correction weight schedule that starts at 0 and ramps up after the primary head has converged could prevent the instability seen in Trial 1 while still providing meaningful correction signal later.
3. **Tanh-bounded correction** — Add `alpha * tanh(correction)` with learned alpha initialized to 0, preventing the correction from growing too large and destabilizing training.
4. **Stronger grad clip** — The divergence in Trial 1 might be prevented with a tighter grad clip (0.3 instead of 0.5).

tcapelle:
ADVISOR: Closing PR #3269 because ADVISOR: Closing — error correction head does not improve DM convergence rate. Trial 2 (w=0.5) at 4.700% val after 227 epochs is epoch-matched identical to baseline (~4.66-4.72% at ep200). The correction head adds architectural complexity and computational overhead without measurable benefit. Trial 1 (w=0.1) diverged at ep80 — the correction-primary gradient interaction is fragile at cosine restart peaks. The correction loss staying active (0.006) shows the head learns something, but it's not translating to primary metric improvement. GPU time better spent on champion config variations. Emma reassigned..

PR #3275, opened 2026-04-24
DrivAerML: local/windowed attention (KNN-restricted attention pattern)

tcapelle:
## Hypothesis

The DM Transolver uses attention over physics-state slices, but the attention computation within each slice can still involve many tokens that are spatially distant. Surface pressure at a given point is primarily determined by **local geometry and flow** — the boundary layer, local curvature, and nearby separation/reattachment. Points on the opposite side of the car contribute noise to the attention pattern.

**Local/windowed attention** restricts each token to attend only to its K nearest spatial neighbors. This:
1. **Reduces attention noise** — distant tokens that contribute little useful signal are masked out
2. **May improve throughput** — sparse attention is cheaper than full attention (if implemented efficiently)
3. **Provides an inductive bias** that matches the physics — surface pressure is a local property

The Transolver already uses a slicing mechanism that provides some spatial grouping, but within each slice, attention is unrestricted. Local attention adds a finer-grained spatial constraint.

**Implementation approach:** Pre-compute the K nearest neighbors for each point using the 3D coordinates. During attention, mask out all non-neighbor pairs by setting their attention scores to -inf before softmax. This is exact (not approximate) and preserves the transformer framework.

### Implementation

1. **Add CLI flags:**
   ```python
   parser.add_argument('--local-attention-k', type=int, default=0,
                       help='Number of nearest neighbors for local attention (0=full attention)')
   ```

2. **Pre-compute KNN indices** at the start of each forward pass (or cache per sample):
   ```python
   if args.local_attention_k > 0:
       # coords: [B, N, 3]
       # Compute pairwise distances
       dists = torch.cdist(coords, coords)  # [B, N, N]
       # Get top-K nearest neighbors (including self)
       _, knn_indices = dists.topk(args.local_attention_k, dim=-1, largest=False)  # [B, N, K]
       # Create attention mask: True for allowed, False for masked
       attn_mask = torch.zeros(B, N, N, dtype=torch.bool, device=coords.device)
       attn_mask.scatter_(2, knn_indices, True)
   ```

3. **Apply the mask in self-attention:**
   ```python
   # In the attention computation, before softmax:
   if attn_mask is not None:
       attn_scores = attn_scores.masked_fill(~attn_mask.unsqueeze(1), float('-inf'))
   ```

   Pass `attn_mask` through the transformer layers. It's the same mask for all layers (geometry doesn't change).

4. **Performance note:** `torch.cdist` on 50k×50k is O(N²) and creates a large tensor. For better performance, consider using the `torch_cluster` package's `knn` function which is more memory-efficient, or compute KNN on CPU and cache it. The KNN computation only needs to happen once per sample (not per layer).

### Trials

**Smoke test first (3 epochs)** — verify shapes, check that attention mask is correct (each token attends to exactly K neighbors), loss finite. Also measure per-epoch time vs baseline.

**Trial 1 — K=512 neighbors (1% of 50k points):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --local-attention-k 512 \
  --wandb_group dm-local-attention
```

**Trial 2 — K=2048 neighbors (4% of 50k points, broader context):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --local-attention-k 2048 \
  --wandb_group dm-local-attention
```

**Important:** Report per-epoch time. If KNN computation adds significant overhead, note it. The mask creation should be cached per sample, NOT recomputed per layer.

## Baseline

Current best DrivAerML metrics (PR #3072):

| Metric | Value | Epoch |
|--------|-------|-------|

- W&B run: `ncl1dh88`

Reproduce:
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -169,6 +169,7 @@ class TrainConfig:
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
+    local_attention_k: int = 0
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT canute:

## Results

### Summary

Both trials ran with KNN-restricted local attention added as a parallel pathway after each Transolver block. The implementation adds a single-head local attention module (proj_dim=64, zero-initialized output) that gathers each token's K nearest spatial neighbors and performs attention over them. Gradient checkpointing is used to control memory.

**Neither trial beats the baseline.** The experiments were severely compute-limited: the KNN attention overhead + disabled torch.compile meant each epoch was ~3-4x slower than the baseline, and within the 6-hour timeout, K=512 reached only 89 epochs and K=2048 reached only 29 epochs. The baseline's best was at epoch 511.

### Trial 1 — K=512 (1% of 50k points)

| Metric | Value | Baseline |
|--------|-------|----------|
| val_primary/surface_rel_l2_pct | **5.813%** (epoch 89) | **3.833%** (epoch 511) |
| test_primary/surface_rel_l2_pct | **6.621%** (epoch 89) | **4.685%** (epoch 511) |
| Epochs completed | 89 | 511 |
| Peak GPU memory | ~19 GB | — |

- W&B run: `9a3yodiw`
- Command:
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=999 SENPAI_TIMEOUT_MINUTES=360 python train.py \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --local-attention-k 512 \
  --wandb-group dm-local-attention --agent canute \
  --wandb-name "canute/dm-local-attn-k512"
```

### Trial 2 — K=2048 (4% of 50k points)

| Metric | Value | Baseline |
|--------|-------|----------|
| val_primary/surface_rel_l2_pct | **13.591%** (epoch 29) | **3.833%** (epoch 511) |
| test_primary/surface_rel_l2_pct | **14.434%** (epoch 29) | **4.685%** (epoch 511) |
| Epochs completed | 29 | 511 |
| Peak GPU memory | ~60 GB | — |

- W&B run: `7shsfomx`
- Command:
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=999 SENPAI_TIMEOUT_MINUTES=360 python train.py \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --local-attention-k 2048 \
  --wandb-group dm-local-attention --agent canute \
  --wandb-name "canute/dm-local-attn-k2048"
```

### What happened

**The local attention approach is computationally prohibitive for this architecture and scale.**

Key findings:

1. **Architecture mismatch:** The Transolver does NOT do point-to-point attention — it uses physics-aware "slicing" where 50k tokens are soft-assigned to 96 virtual slice tokens, and attention happens between these 96 slices (O(S²) complexity). Adding explicit KNN-restricted point-to-point attention creates a parallel O(N·K) pathway that dominates compute cost.

2. **Throughput collapse:** Each epoch took ~4 minutes for K=512 vs ~1 minute for the baseline (estimated). K=2048 was ~3-4x slower still (~12 min/epoch). This is because:
   - KNN computation: chunked GPU cdist on 50k points per batch (~0.5s/batch)
   - Gather operations: materializing [B, 50k, K, 64] tensors for K/V neighbors
   - torch.compile disabled (incompatible with monkey-patched backbone)

3. **Per-epoch learning rate:** At matched epochs, K=512 and K=2048 showed very similar val metrics (within ~1-3%), suggesting the local attention isn't providing meaningful additional signal beyond what the slice mechanism already captures. Comparing at epoch 29:
   - K=512: 12.82%
   - K=2048: 13.59%
   - (K=2048 slightly worse, possibly because of higher memory pressure and less effective gradient flow through the larger gather)

4. **Memory:** K=512 used ~19 GB (modest), K=2048 used ~60 GB (substantial but feasible on 97 GB GPUs with gradient checkpointing).

### Why it didn't work

The hypothesis that local spatial attention would help assumed point-to-point attention, but the Transolver's slice mechanism already provides an implicit spatial grouping — tokens are soft-clustered into physics-coherent slices. The additional KNN attention adds significant compute cost without clear benefit, suggesting the slice mechanism is already capturing the relevant local spatial structure.

### Suggested follow-ups

1. **Spatially-biased slice assignment** — Instead of adding a full local attention pathway, modify the slice assignment logits to encourage spatial coherence (e.g., add a spatial distance penalty to the slice softmax). This would be O(1) extra cost and directly tests whether forcing spatial locality in slicing helps.

2. **PTv3-style serialized attention** — Sort tokens by a space-filling curve (Morton/Hilbert code) and apply sliding-window attention on the sorted sequence. This avoids KNN computation entirely and is compatible with FlashAttention.

3. **Visualize slice assignments** — Before adding more spatial bias, check if the learned slices are already spatially coherent. If they are, explicit locality constraints are redundant.

tcapelle:
ADVISOR: Closing PR #3275 because ADVISOR: Closing as dead end. Both trials confirm architectural mismatch:

- K=512: 5.81% at ep89 (only 89 epochs due to 3-4x throughput collapse). Baseline: 3.83% at ep511.
- K=2048: 13.59% at ep29 (only 29 epochs, 60GB VRAM). 

The Transolver doesn't do point-to-point attention — it routes 50k tokens to 96 physics-aware slices and attends between slices. Adding a parallel KNN point-attention pathway is redundant with the slice mechanism (matched-epoch metrics are nearly identical with and without local attention) and costs 3-4x per-epoch throughput, which is fatal under our epoch-starvation constraint.

Student's insight about spatially-biased slice assignment (modifying slice logits to encourage spatial coherence) is worth noting for future work, but adding explicit KNN attention is not the way to achieve locality in this architecture. GPU reassigned..

PR #3276, opened 2026-04-24
DrivAerML: cosine similarity auxiliary loss (spatial pattern fidelity)

tcapelle:
## Hypothesis

MSE alone optimizes per-point magnitude accuracy — it treats each surface point independently. But surface pressure has **spatial structure** (smooth variations, sharp gradients at separation points, pressure recovery patterns) that MSE doesn't explicitly reward. A model could have low MSE but systematically distort the spatial PATTERN of pressure distribution.

**Cosine similarity** between the predicted and target pressure vectors captures the **shape** of the spatial pattern independently of absolute scale:
\`cos_sim = (pred · target) / (||pred|| × ||target||)\`

Adding \`1 - cos_sim\` as an auxiliary loss encourages the model to match the overall spatial pressure distribution pattern, not just minimize per-point squared differences. This:
1. **Scale-invariant** — a prediction that has the right pattern but wrong magnitude still gets cosine credit
2. **Complements MSE** — MSE handles magnitude, cosine handles pattern fidelity
3. **Computationally free** — dot product and norms over 50k points, negligible vs transformer forward pass
4. **Different gradient signal** — MSE gradients scale with error magnitude (small errors → small gradients), while cosine gradients emphasize direction alignment regardless of magnitude. This provides informative gradients even for well-predicted points.

The key insight: if the remaining 3.833% error is systematic spatial bias (same regions consistently wrong), cosine similarity loss directly penalizes pattern distortion, potentially pulling the model toward the correct spatial structure.

### Implementation

1. **Add CLI flag:**
   ```python
   parser.add_argument('--cosine-sim-weight', type=float, default=0.0,
                       help='Weight for cosine similarity auxiliary loss (0=disabled)')
   ```

2. **Compute cosine similarity loss** alongside MSE:
   ```python
   if args.cosine_sim_weight > 0:
       # pred, target: [B, N, C] — B=1, N=50k points, C=output channels
       # Flatten spatial dimensions for per-channel cosine sim
       pred_flat = pred.reshape(pred.shape[0], -1)    # [B, N*C]
       target_flat = target.reshape(target.shape[0], -1)  # [B, N*C]
       
       cos_sim = F.cosine_similarity(pred_flat, target_flat, dim=-1)  # [B]
       cosine_loss = (1 - cos_sim).mean()
       
       total_loss = mse_loss + args.cosine_sim_weight * cosine_loss
   ```

   **Alternative (per-channel):** Compute cosine similarity per output channel separately, then average:
   ```python
   cos_losses = []
   for c in range(pred.shape[-1]):
       cos_sim_c = F.cosine_similarity(pred[..., c], target[..., c], dim=-1)
       cos_losses.append((1 - cos_sim_c).mean())
   cosine_loss = torch.stack(cos_losses).mean()
   ```
   The per-channel version may be better since different output channels (Ux, Uy, p) have different distributions. Try the simpler flat version first, switch to per-channel if results are promising.

3. **Log the cosine similarity** as a metric (not just the loss):
   ```python
   wandb.log({'train/cosine_similarity': cos_sim.mean().item()})
   ```

### Trials

**Smoke test first (3 epochs)** — verify loss finite, cosine similarity in reasonable range (0.5-0.99).

**Trial 1 — weight=0.1 (moderate):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --cosine-sim-weight 0.1 \
  --wandb_group dm-cosine-sim-loss
```

**Trial 2 — weight=0.5 (strong pattern emphasis):**
```bash
  --dataset drivaerml \
  --optimizer adamw --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --epochs 999 \
  --batch-size 1 \
  --ema-decay 0.9995 \
  --cosine-sim-weight 0.5 \
  --wandb_group dm-cosine-sim-loss
```

Log the cosine similarity metric throughout training — it should increase toward 1.0. If it plateaus early (>0.99 before epoch 100), the pattern is already well-learned and this loss provides diminishing returns. If it's still <0.95 at convergence, the pattern fidelity is a genuine bottleneck.

## Baseline

Current best DrivAerML metrics (PR #3072):

| Metric | Value | Epoch |
|--------|-------|-------|

- W&B run: `ncl1dh88`

Reproduce:
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995
```

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,7 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    cosine_sim_weight: float = 0.0
     save_checkpoint: bool = False
     seed: int = 0
 
@@ -1277,6 +1278,7 @@ def train_one_epoch(
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
     volume_loss_weight: float = 1.0,
+    cosine_sim_weight: float = 0.0,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
@@ -1334,6 +1336,19 @@ def train_one_epoch(
                         )
                         loss, _ = loss_grouped(batch, outputs, transform, volume_loss_weight=volume_loss_weight)
 
+                    if cosine_sim_weight > 0 and batch.surface_y is not None and outputs["surface_preds"] is not None:
+                        pred_masked = outputs["surface_preds"][batch.surface_mask].float()
+                        target_masked = transform.apply(batch.surface_y)[batch.surface_mask].float()
+                        pred_flat = pred_masked.reshape(1, -1)
+                        target_flat = target_masked.reshape(1, -1)
+                        cos_sim = F.cosine_similarity(pred_flat, target_flat, dim=-1)
+                        cosine_loss = (1.0 - cos_sim).mean()
+                        loss = loss + cosine_sim_weight * cosine_loss
+                        running.setdefault("cosine_similarity", 0.0)
+                        running["cosine_similarity"] += float(cos_sim.mean().detach().cpu().item())
+                        running.setdefault("cosine_loss", 0.0)
+                        running["cosine_loss"] += float(cosine_loss.detach().cpu().item())
+
             accum_loss += float(loss.detach().cpu().item())
             (loss / grad_accum_steps).backward()
             micro_count += 1
@@ -1365,6 +1380,10 @@ def train_one_epoch(
     running["loss"] /= max(steps, 1)
     if "grad_norm_mean" in running:
         running["grad_norm_mean"] /= max(steps, 1)
+    if "cosine_similarity" in running:
+        running["cosine_similarity"] /= max(steps, 1)
+    if "cosine_loss" in running:
+        running["cosine_loss"] /= max(steps, 1)
     running["train_steps"] = float(steps)
     running["micro_batches"] = float(micro_batches_total)
     if scheduler is not None:
@@ -1694,6 +1713,7 @@ def main() -> None:
             grad_clip=config.grad_clip,
             grad_accum_steps=config.grad_accum_steps,
             volume_loss_weight=config.volume_loss_weight,
+            cosine_sim_weight=config.cosine_sim_weight,
         )
 
         if ema is not None:


tcapelle:
STUDENT zenitsu:

## Results

### Summary

**Hypothesis: REJECTED.** The cosine similarity auxiliary loss causes training instability at longer horizons. Both weight settings eventually diverge — w=0.5 catastrophically at epoch ~130, w=0.1 more gradually at epoch ~242. Neither trial beat the baseline.

### Metrics

| Trial | Weight | Best val % | Best epoch | Diverge epoch | Baseline val % |
|-------|--------|-----------|------------|---------------|----------------|
| Trial 1 | 0.1 | **4.661%** | 241 | 245 (NaN) | 3.833% |
| Trial 2 | 0.5 | **6.569%** | 109 | ~159 | 3.833% |
| Baseline | 0.0 | **3.833%** | 511 | — | — |

**No test metrics available** — both runs diverged before completing normally (test eval runs at training end).

### Cosine similarity trajectory

| Epoch range | Trial 1 cos_sim | Trial 2 cos_sim |
|------------|----------------|----------------|
| 0-30 | 0.80 → 0.96 | 0.80 → 0.96 |
| 30-100 | 0.96 → 0.991 | 0.96 → 0.993 |
| 100-200 | 0.991 → 0.997 | 0.993 → diverged |
| 200-241 | 0.997 (plateaued) | — |
| 242+ | diverged → NaN | — |

Cosine similarity plateaued above 0.99 by epoch ~100 for both trials, confirming the PR's prediction that pattern fidelity saturates early. After saturation, the cosine loss provides minimal useful gradient but adds noise that eventually triggers instability.

### Divergence analysis

**Trial 1 (w=0.1) — diverged at epoch 242:**
The collapse happened precisely at a cosine annealing LR restart (epoch 240, T_max=30). At the restart, LR jumps from ~0 to 5e-4. The gradient norm trajectory:
- Epoch 241: grad_norm=0.17 (normal)
- Epoch 242: grad_norm=20.6 (120x spike)
- Epoch 243: grad_norm=1437 (catastrophic)
- Epoch 244: grad_norm=inf → NaN from epoch 245

**Trial 2 (w=0.5) — diverged at epoch ~130:**
Gradual degradation starting ~epoch 130. Val jumped from 6.57% best to 9%, then progressively worsened to 86%+ by epoch ~200. The heavy cosine weight amplified gradient conflicts between MSE and cosine objectives, especially at LR restarts.

**Root cause:** When cosine similarity is near 1.0 (saturated), the gradient of `(1 - cos_sim)` becomes poorly conditioned. Combined with the sudden LR jumps from cosine annealing restarts, this creates gradient explosions. The heavier weight (0.5) hits this earlier because it amplifies the problematic gradients more.

### Commands used

**Trial 1 (w=0.1):**
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --cosine-sim-weight 0.1 \
  --agent zenitsu --wandb-name "zenitsu/dm-cosine-sim-w0.1" --wandb-group "dm-cosine-sim-loss"
```

**Trial 2 (w=0.5):**
```bash
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --cosine-sim-weight 0.5 \
  --agent zenitsu --wandb-name "zenitsu/dm-cosine-sim-w0.5" --wandb-group "dm-cosine-sim-loss"
```

### W&B runs
- Trial 1 (w=0.1): `ry1lq2vw`
- Trial 2 (w=0.5): `flnnss55`

### Peak memory
Estimated ~2.5 GB per GPU (NVIDIA RTX PRO 6000 Blackwell, batch_size=1, 50k surface points)

### What happened

The cosine similarity loss initially looked promising — both trials converged faster in early epochs compared to what the baseline trajectory suggests for the same epoch range. However, the loss is fundamentally unstable at longer training horizons due to gradient conditioning issues when cosine similarity approaches 1.0.

The key failure mode: once the model learns the spatial pattern well (cos_sim > 0.99), the cosine loss gradient becomes ill-conditioned. At cosine annealing LR restarts (where LR jumps from ~0 to 5e-4), this ill-conditioned gradient causes catastrophic gradient explosions. Grad clipping at 0.5 was insufficient to prevent the divergence — the explosions happen faster than clipping can stabilize.

Trial 1 (w=0.1) was approaching the baseline — 4.661% vs 3.833% at epoch 241 out of a projected ~500 epochs — but diverged before it could close the gap. It's plausible that without the divergence it could have matched or slightly beaten the baseline, but we can't know.

### Suggested follow-ups

1. **Cosine loss with annealing/decay**: Instead of a fixed weight, decay the cosine weight to 0 as cos_sim approaches 0.99. This keeps the useful early-training signal while avoiding the instability at saturation.

2. **Cosine loss with warm-restart protection**: Disable the cosine loss term during the first few epochs after each LR restart (when LR is highest), then re-enable it once gradients stabilize.

3. **Much smaller weight (w=0.01-0.05)**: The w=0.1 trial was on a promising trajectory before divergence. A smaller weight might survive through 500+ epochs.

4. **Per-channel cosine similarity**: As the PR suggested, computing cosine sim per output channel (Ux, Uy, p separately) might provide better gradient conditioning than the flat version.

5. **Gradient clipping specifically for the cosine loss**: Apply separate, tighter gradient clipping to the cosine loss gradients.

tcapelle:
ADVISOR: Closing PR #3276 because ADVISOR: Closing — dead end. w=0.1: 4.661% (+22% vs baseline), w=0.5: 6.569% (+71%), both crashed at cosine restart boundaries. Key finding: cosine similarity saturates above 0.99 by epoch ~100 in both trials — the model already learns spatial pattern structure from MSE alone, making the cosine loss informationally redundant. The divergence mechanism is clear: near-unity cos_sim makes (1-cos_sim) gradients ill-conditioned, which amplifies LR restart shocks beyond gc=0.5 containment. Per issue #3283, GPU time is redirected to breakout lanes. Good experimental work..

PR #3278, opened 2026-04-24
AF: separate volume prediction head with extra capacity

tcapelle:
## Hypothesis

The current model uses a **shared output MLP** for all points — surface and volume predictions come from the same final layer. But vol_mse (0.002039) is ~7x higher than surface_mse (0.000296), suggesting the shared output bottleneck under-allocates capacity to volume.

We replace the single shared output MLP with **two separate heads**:
- **Surface head**: standard capacity (matches current output MLP — 256→256→out_channels)
- **Volume head**: double-width (256→512→out_channels, ~2x parameters in the output layer)

The transformer trunk is unchanged (shared — still benefits from joint surface+volume representation learning). Only the final output projection is split. This gives volume predictions a dedicated pathway with more parameters to model the more complex interior flow field.

Why this makes physical sense: surface aerodynamics is largely a boundary condition problem (the geometry defines most of the solution). Volume aerodynamics requires modeling wake formation, separation bubbles, and 3D vortical structures — a fundamentally richer prediction problem. The same output capacity used for surface should not suffice for volume.

## Instructions

All changes in `target/icml2026/train.py`. Champion config unchanged — only split the output MLP.

**Step 1: Find the current output MLP in the model class**

Locate the output projection layer (likely `self.output_mlp` or `self.head` or similar — it maps from `hidden_dim` to `output_channels`).

**Step 2: Replace with two separate heads**

In `__init__`, replace the single output MLP with:

```python
# Surface head: standard capacity
self.surface_head = nn.Sequential(
    nn.GELU(),
    nn.Linear(hidden_dim, output_channels),
)
# Volume head: double width for extra capacity
self.volume_head = nn.Sequential(
    nn.Linear(hidden_dim, hidden_dim * 2),
    nn.GELU(),
    nn.Linear(hidden_dim * 2, output_channels),
)
```

**Step 3: Route by point type in forward pass**

The data pipeline already separates surface/volume points. In `forward()`, after the transformer trunk:

```python
# transformer_features: [N, hidden_dim] — all points (surface + volume mixed)
# surface_mask: boolean tensor [N] — True for surface points
# volume_mask: boolean tensor [N] — True for volume points

output = torch.zeros(N, output_channels, device=x.device)
output[surface_mask] = self.surface_head(transformer_features[surface_mask])
output[volume_mask] = self.volume_head(transformer_features[volume_mask])
```

Check how the data pipeline provides point type membership — look for how `vol_loss` vs `surface_loss` are computed to find where the split happens.

**Step 4: Run with `--wandb_group af-separate-vol-head`**

```bash
  --grad-clip 1.0 --weight-decay 1e-2 --enable-fourier \
  --wandb_group af-separate-vol-head
```

Report both `val_primary/surface_mse` and `val_primary/vol_mse` at best checkpoint, and the epoch at which each is achieved.

## Baseline

- **W&B run:** (PR #3135, EMA=0.999 + vol-weight=10x champion)

## Success criteria

- **Primary target:** vol_mse < 0.002039 without surface_mse regression above 0.000296
- **Volume closure goal:** vol_mse < 0.0017 (SpiderSolver benchmark target — currently 1.20x gap)
- Surface must not regress more than 5% (hard constraint per human directive)

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -167,6 +167,8 @@ class TrainConfig:
     grad_clip: float = 0.0
     grad_accum_steps: int = 1
     volume_loss_weight: float = 1.0
+    separate_volume_head: bool = False
+    vol_head_width_mult: int = 2
     save_checkpoint: bool = False
     seed: int = 0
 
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT gohan:

## Results

### Separate Volume Head — NEGATIVE RESULT

| Metric | Baseline (#3135) | Separate Head | Delta |
|--------|------------------|---------------|-------|
| val surface_mse | **0.000296** (ep 704) | 0.001580 (ep 646) | +433% ❌ |
| val vol_mse | **0.002039** (ep 743) | 0.002537 (ep 700) | +24% ❌ |
| test surface_mse | — | 0.001288 (at best-val ep) | — |
| test vol_mse | — | 0.004071 (at best-val ep) | — |
| SpiderSolver target | vol < 0.0017 | 0.002537 | 1.49x gap |

**Both metrics regressed.** Surface regressed catastrophically (5.3x worse), volume regressed 24%.

### Experiment details

- **Command:** `cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 python train.py --dataset airfrans --airfrans-task full --optimizer adamw --lr 6e-4 --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 --epochs 999 --ema-decay 0.999 --volume-loss-weight 10.0 --separate-volume-head --wandb-group af-separate-vol-head --wandb-name "gohan/af-separate-vol-head" --agent gohan`
- **W&B run:** dw5q6x75 (gohan/af-separate-vol-head)
- **Total epochs:** 714
- **Peak VRAM:** ~31 GB (similar to baseline)

### What happened

The hypothesis that giving volume predictions a dedicated wider output head would improve vol_mse was **falsified**. Both surface and volume got worse.

**Why it likely failed:**

1. **Shared projection provides useful cross-domain coupling.** The baseline's single `LinearProjection(n_hidden, surface_dim + volume_dim)` forces surface and volume output channels to co-adapt during training. Splitting this breaks implicit feature sharing at the output layer — the backbone hidden states lose the gradient signal that couples surface and volume predictions.

2. **MLP output heads are harder to optimize than single linear projections.** The baseline uses a single Linear layer for output projection. The separate heads use 2-layer MLPs (256→256→out for surface, 256→512→out for volume). The extra nonlinearity adds optimization difficulty without a corresponding gain. The surface path is particularly affected: it now has *two* stacked MLPs (the new surface_out_mlp + the existing ZeroInitSurfaceRefinementHead), likely over-parameterizing it.

3. **No training recipe adaptation.** The same LR, WD, and schedule were used. More parameters typically need either lower LR or more regularization — the baseline recipe was tuned for the simpler architecture.

### Implementation notes

- Created `SeparateHeadTransolver` subclass in `train.py` that overrides `SenpaiTransolver.forward()` to route through separate surface/volume output MLPs instead of the shared `self.out` linear projection.
- Added `--separate-volume-head` and `--vol-head-width-mult` config flags.
- Also added model param count logging to wandb config for future comparison.

### Suggested follow-ups

- **Volume-only head split:** Keep the shared linear projection for surface (which already has the refinement head), but add a dedicated wider MLP *only* for volume. This preserves the well-tuned surface path while still giving volume more capacity.
- **Wider single linear projection:** Instead of separate MLPs, simply increase the volume output dimension in the shared `self.out` and add a learned projection to map back. Cheaper than full MLP heads.
- **Volume refinement head (zero-init residual):** Mirror the existing `ZeroInitSurfaceRefinementHead` pattern for volume — add a zero-initialized residual MLP after the shared projection. This preserves the baseline behavior at init and lets volume capacity grow during training without disrupting the surface path.

tcapelle:
ADVISOR: Closing PR #3278 because ADVISOR: Closing — separate vol head surface=0.001580 (+433%), vol=0.002537 (+24%). Architecture detrimental to both metrics. Gohan reassigned..

PR #3279, opened 2026-04-24
DrivAerML: ReLU² activation in FFN (Primer-style squared ReLU)

tcapelle:
## Hypothesis

The DM transformer uses GELU activation in its FFN layers. ReLU² (squared ReLU: `relu(x)²`) was identified in the Primer architecture search (So et al., 2021, "Primer: Searching for Efficient Transformers for Language Modeling") as consistently outperforming GELU across transformer scales. ReLU² produces **sparser activations** than GELU — zero for all negative inputs AND quadratically suppressed for small positive inputs. Sparser activations create more feature-selective representations, which could help the model develop specialized spatial features for different aerodynamic regions rather than diffuse representations across all regions.

This directly targets the systematic spatial bias problem: if the model's FFN activations are too dense/uniform, it conflates spatially distinct regions with similar average features. Sparser activations force sharper feature discrimination.

**Zero throughput penalty** — ReLU² is computationally simpler than GELU (one relu + one multiply vs. the GELU approximation). If anything, it should be marginally faster.

Reference: So et al. (2021), "Primer: Searching for Efficient Transformers for Language Modeling" — found ReLU² in automated architecture search, validated across multiple scales.

## Instructions

All changes in `target/icml2026/train.py`. Champion config unchanged — only modify the FFN activation function.

**Step 1: Define ReLU² activation**

```python
def relu_squared(x):
    return torch.relu(x).square()
```

or equivalently:

```python
def relu_squared(x):
    return F.relu(x) ** 2
```

**Step 2: Replace GELU with ReLU² in the FFN layers**

Find where the FFN (feed-forward network) activation is defined in the transformer layers. It will be something like `nn.GELU()` or `F.gelu(x)`. Replace with the ReLU² function defined above.

Only change the FFN activation inside the transformer layers. Leave any other activations (e.g., in the output MLP, in the Fourier encoding) unchanged.

**Step 3: Run with `--wandb_group dm-relu-squared`**

```bash
  --wandb_group dm-relu-squared
```

**Verify:** (1) epoch time ≤ baseline (~22s/epoch), (2) training doesn't diverge in first 30 epochs, (3) report best val surface_rel_l2_pct and the epoch at which it occurs.

## Baseline

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -33,6 +33,37 @@ from core.features import (
 from core.optim import Lion, Lookahead
 
 
+class ReLUSquared(torch.nn.Module):
+    """Squared ReLU activation from Primer (So et al., 2021): relu(x)^2.
+
+    Produces sparser activations than GELU: zero for all negative inputs and
+    quadratically suppressed for small positive inputs.
+    """
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return torch.relu(x).square()
+
+
+def apply_ffn_activation(model: torch.nn.Module, activation: str) -> None:
+    """Replace UpActDownMlp.act in all transformer FFN layers with the chosen activation.
+
+    Only patches the transformer backbone FFN blocks (UpActDownMlp); leaves
+    ZeroInitSurfaceRefinementHead and MLP bias modules untouched.
+    """
+    if activation == "gelu":
+        return
+    from core.architectures.transolver_reference import UpActDownMlp
+
+    if activation == "relu_squared":
+        act_module: torch.nn.Module = ReLUSquared()
+    else:
+        raise ValueError(f"Unknown ffn_activation: {activation!r}. Expected 'gelu' or 'relu_squared'.")
+
+    for module in model.modules():
+        if isinstance(module, UpActDownMlp):
+            module.act = act_module
+
+
 class EMAWithWarmup:
     """EMA with timm-style adaptive decay warmup: min(target, (1+step)/(10+step))."""
 
@@ -169,6 +200,7 @@ class TrainConfig:
     volume_loss_weight: float = 1.0
     save_checkpoint: bool = False
     seed: int = 0
+    ffn_activation: str = "gelu"
 
 
 @dataclass
@@ -1621,6 +1653,7 @@ def main() -> None:
 
     train_loader, val_loaders, test_loaders = build_loaders(config, bundle, num_workers=resolved_num_workers)
     model = build_model(config, bundle).to(device)
+    apply_ffn_activation(model, config.ffn_activation)
     forward_model = torch.compile(model) if config.compile_model and device.type == "cuda" else model
     anp_head = None
     if bundle.spec.name == "tandemfoilset" and config.anp_srf:


tcapelle:
STUDENT gojo:

## Results

### DrivAerML: ReLU² activation in FFN (Primer-style squared ReLU)

| Metric | Baseline (GELU) | ReLU² (this run) | Delta |
|--------|-----------------|-------------------|-------|
| **best val surface_rel_l2_pct** | 3.833% (ep511) | 8.832% (ep72) | +130.4% ❌ |
| **best epoch** | 511 | 72 | — |
| **stability** | Stable through 517 epochs | **Diverged at epoch 73** | — |

**W&B run:** `2k52eunu` (gojo/dm-relu-squared)

**Command:**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=999 python train.py \
  --ffn-activation relu_squared \
  --wandb-group dm-relu-squared \
  --agent gojo --wandb-name "gojo/dm-relu-squared"
```

**Peak memory:** Single GPU (H100 97GB), standard DrivAerML footprint.

### What happened

**Negative result — ReLU² is unstable with cosine LR restarts on DrivAerML.**

The training had two distinct phases:

1. **Epochs 1–72 (stable phase):** Beautiful convergence. 72 consecutive new bests, val dropped from 57.67% to 8.83%. Grad norms declined from 9.92 to 0.71, grad clip events from 394→244. The model was well-conditioned and converging smoothly. At epoch 60, ReLU² was tracking very close to the GELU baseline (~9.86% vs ~9.75%).

2. **Epochs 73–81 (divergence):** Catastrophic instability at a cosine LR peak. Grad norm exploded from 0.71→20.86→18.6 (28x spike). Loss jumped from 0.025→0.35 (14x). Val deteriorated from 8.83%→27.06% in 9 epochs, unwinding all progress back to epoch-9 levels. The divergence was irrecoverable despite gc=0.5 clipping — the EMA model was deeply poisoned.

**Root cause analysis:** ReLU² produces sparser activations than GELU — zero for all negative inputs AND quadratically suppressed for small positive inputs. This sparsity appears to amplify gradient instabilities when the cosine LR schedule pushes the learning rate to its peak (5e-4). GELU's smoother gradient landscape (non-zero gradients for negative inputs) acts as an implicit stabilizer during LR peaks. The gc=0.5 clipping that was sufficient to stabilize GELU+EMA is not strong enough for ReLU².

**Key observations:**
- Epoch time was normal (~22s/epoch), confirming the "zero throughput penalty" claim
- Training did NOT diverge in the first 30 epochs (as requested to verify) — it was stable for 72 epochs
- The divergence happened at the 3rd cosine cycle peak (epoch ~73), suggesting cumulative instability that builds over multiple cycles
- The GELU baseline with identical config was stable through 517+ epochs

### Suggested follow-ups

1. **Stronger grad clipping for ReLU²:** gc=0.3 or gc=0.2 might prevent the gradient explosion at LR peaks. The sparser activation needs tighter guardrails.
2. **Lower peak LR:** Reducing lr from 5e-4 to 3e-4 or 2e-4 with ReLU² might keep it in the stable regime.
3. **Longer T_max:** T_max=50 or T_max=100 would reduce the LR peak frequency and give the model more time to stabilize between peaks.
4. **GELU variants:** If the goal is sparser activations, consider SwiGLU or GEGLU which combine gating with GELU — potentially getting some sparsity benefit without full ReLU² instability.

tcapelle:
ADVISOR: Closing PR #3279 because ADVISOR: Closing as dead end. ReLU² diverges catastrophically at ep73 (8.83%→27% in 9 epochs) during 3rd cosine LR peak. The sparse activation amplifies gradient instability at LR peaks — gc=0.5 isn't sufficient as a guard. GELU is inherently more stable due to non-zero gradients for negative inputs. Not worth pursuing given time constraints. GPU reassigned..

PR #3280, opened 2026-04-24
DrivAerML: per-channel output heads (separate MLPs for Ux, Uy, p)

tcapelle:
## Hypothesis

The DM model uses a **shared output MLP** for all three output channels (Ux, Uy, p). These channels have fundamentally different physical characteristics:
- **Ux/Uy (velocity):** smooth, bounded, spatially coherent flow fields
- **p (pressure):** sharp gradients at stagnation points, separation zones, and geometric discontinuities

A shared output MLP forces all channels through the same nonlinear mapping. If the model optimizes toward pressure prediction (which has larger magnitude gradients), velocity prediction may be under-optimized. Separate per-channel output heads give each physical quantity its own output pathway, allowing the model to learn channel-specific nonlinear mappings from the shared transformer representation.

This is a pure architectural change with **minimal throughput impact** — three small MLPs ≈ one medium MLP in total FLOPS. No loss modification, no train/eval mismatch, no sampling changes.

## Instructions

All changes in `target/icml2026/train.py`. Champion config unchanged — only split the output MLP.

**Step 1: Find the current output MLP**

Locate the output projection that maps from the transformer's hidden dimension (512) to the output channels (3 for DrivAerML: Ux, Uy, p). It will be something like:
```python
self.output_mlp = nn.Sequential(
    nn.GELU(),
    nn.Linear(hidden_dim, output_channels),  # output_channels = 3
)
```

**Step 2: Replace with per-channel heads**

```python
# Per-channel output heads
self.head_ux = nn.Sequential(
    nn.Linear(hidden_dim, 256),
    nn.GELU(),
    nn.Linear(256, 1),
)
self.head_uy = nn.Sequential(
    nn.Linear(hidden_dim, 256),
    nn.GELU(),
    nn.Linear(256, 1),
)
self.head_p = nn.Sequential(
    nn.Linear(hidden_dim, 256),
    nn.GELU(),
    nn.Linear(256, 1),
)
```

**Step 3: Apply in forward pass**

Replace the shared output MLP call with:
```python
# features: [N, hidden_dim] from transformer
pred_ux = self.head_ux(features)  # [N, 1]
pred_uy = self.head_uy(features)  # [N, 1]
pred_p = self.head_p(features)    # [N, 1]
output = torch.cat([pred_ux, pred_uy, pred_p], dim=-1)  # [N, 3]
```

**Important:** Check the target channel ordering in the data pipeline to ensure Ux, Uy, p map to the correct output indices. Look at how the loss is computed to verify.

**Step 4: Run with `--wandb_group dm-per-channel-heads`**

```bash
  --wandb_group dm-per-channel-heads
```

**Verify:** epoch time stays within 5% of baseline (~22 sec/epoch). Report best val surface_rel_l2_pct, epoch at best, and if possible per-channel val metrics to show which channel benefits most from the split.

## Baseline


tcapelle:
ADVISOR: Closing PR #3280 because ADVISOR: Closing per issue #3283 directive. DrivAerML surface target is single-channel cp — separate Ux/Uy/p heads are conceptually mismatched. Reassigning to priority DM lane..

PR #3289, opened 2026-04-24
DrivAerML: 64k surface points (high-density context)

tcapelle:
## Hypothesis
Increasing training surface point density from 50k to 64k gives the model richer local geometry context per step, potentially improving surface error without requiring more epochs. Eval stays at 50k for fast steering.

## Bucket
C — Surface Density Sweep

## Command
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 64000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --wandb_group dm-surface-density
```

## Kill Condition
If throughput drops below ~250 epochs in 10h AND val is not clearly better than baseline by epoch 150, kill.

## Baseline

## Expected outcome
64k points ~28% denser context. If val improves, askeladd/dm-surface-96k will confirm the scaling trend.


tcapelle:
STUDENT himmel:

## Results

### Key Metrics

| Metric | Baseline (50k) | 64k Experiment | Delta |
|--------|----------------|----------------|-------|
| **val surface_rel_l2_pct** | 3.833% (ep511) | **3.763%** (ep819) | **-0.070 pp** ✅ |
| **test surface_rel_l2_pct** | 4.685% | **4.800%** | **+0.115 pp** ❌ |
| AB-UPT target | <3.71% | 3.763% | not reached |

### Training Details
- **Epochs:** 820 (600-min timeout, still actively descending at cutoff)
- **Best epoch:** 819 (val 3.763%)
- **Throughput:** ~82 epochs/hour (~43s/epoch)
- **Peak GPU memory:** ~9 GB (single GPU)
- **W&B run:** [`5jfjyr75`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/5jfjyr75)

### Command
```bash
CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 64000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --wandb-group dm-surface-density --agent himmel --wandb-name "himmel/dm-surface-64k"
```

### What happened

**Mixed result — val improvement but test regression.**

The 64k surface point configuration eventually beat the baseline on val (3.763% vs 3.833%), but this came at significant cost:

1. **Convergence was much slower.** The baseline reached 3.833% at epoch 511. The 64k run took until ep667 (~30% more epochs) just to match the baseline, and reached its best at ep819. The model was still actively descending at timeout — four consecutive bests in the final 4 epochs suggest it hadn't converged yet.

2. **Test metric is worse.** Despite the val improvement, test surface_rel_l2_pct regressed from 4.685% to 4.800% (+0.115 pp). This is the critical finding — the denser training surface points appear to help the model fit the val distribution better but **don't generalize** to unseen test geometries.

3. **Convergence trajectory.** The run showed a distinctive pattern of long plateaus (20-40 epochs) followed by sudden descents, with each plateau settling at a lower floor. Major breakthrough phases occurred around ep465-480, ep540-565, ep595-600, ep660-667, and ep760-820.

4. **AB-UPT target not reached.** The <3.71% val target was not reached, though the model was still descending. More compute might close this gap for val, but the test regression makes this moot.

### Interpretation

The 64k training density likely allows the model to overfit more effectively to the specific geometries and surface patterns in the training/val set. Since eval uses 50k points (same as baseline), the val metric benefits from the richer training signal. However, this specialization doesn't transfer — the test set's unseen geometries don't benefit from the denser training representation.

The hypothesis that denser surface points give "richer local geometry context" is partially supported for in-distribution performance but **falsified for generalization**. The extra compute per step (denser sampling) would be better spent on other improvements.

### Suggested follow-ups

- **Skip the 96k experiment** (dm-surface-96k). If 64k shows test regression, 96k will likely be worse. The density scaling hypothesis is not supported.
- **Try 64k at eval time too** — if the model trained on 64k also evaluated on 64k, it might show different test behavior. But this changes the eval protocol.
- **Focus compute budget elsewhere.** The 50k baseline is a strong local optimum. Gains are more likely from architectural/loss changes than density scaling.

tcapelle:
ADVISOR: Closing PR #3289 because ADVISOR: Closing — 64k surface points val=3.763% beats old MSE baseline but doesn't beat current 3.700% metric-aware baseline. Test regressed (+0.115pp). 64k density doesn't generalize. Himmel reassigned..

PR #3290, opened 2026-04-24
DrivAerML: 96k surface points (high-density context)

tcapelle:
## Hypothesis
Pushing surface point density from 50k to 96k (92% increase) to test whether significantly richer per-step geometry context yields a clear win. If both 64k (himmel) and 96k show improvement, we have a scaling trend to exploit further.

## Bucket
C — Surface Density Sweep

## Command
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 96000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --wandb_group dm-surface-density
```

## Kill Condition
If throughput drops below ~200 epochs in 10h AND val not clearly better than baseline by epoch 150, kill. GPU memory may be the binding constraint — report VRAM usage.

## Baseline

## Expected outcome
96k tests the upper end of the density range. If val improves over 64k result, density is a strong lever. Report VRAM usage and throughput (epochs/h) to help calibrate future runs.


tcapelle:
ADVISOR: Closing PR #3290 because ADVISOR: Closing — student agent failed to set up (6 min session, 0 lines changed). 96k surface points at old T_max=30 config superseded by compound fleet. Reassigning..

PR #3294, opened 2026-04-24
DrivAerML: gradient-spike-safe champion gc=0.25

tcapelle:
## Hypothesis
Tighter gradient clipping (gc=0.25 vs champion gc=0.5) to suppress the gradient spikes that appear to cause late-training collapses. The champion run (ncl1dh88) shows occasional gradient spikes — halving the clip threshold should allow the model to train deeper into the schedule without catastrophic updates.

If --skip-update-grad-norm is available from #3284, also add --skip-update-grad-norm 100 to skip the weight update on any step where the raw grad norm exceeds 100.

## Bucket
D — Gradient Safety

## Command
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.25 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --wandb_group dm-gradient-safe
```

## Kill Condition
If val is clearly worse than 3.833% by epoch 200 with no sign of catching up, kill.

## Baseline

## Expected outcome
If gradient spikes are the primary instability source, gc=0.25 should allow the model to reach lower val loss at late epochs. Report gradient norm statistics (mean, max, p99) alongside val metrics.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -1292,6 +1292,7 @@ def train_one_epoch(
     if anp_head is not None:
         anp_head.train()
     running = {"loss": 0.0}
+    grad_norms_list: list[float] = []
     steps = 0
     micro_batches_total = 0
     batches = loader if max_batches <= 0 else itertools.islice(loader, max_batches)
@@ -1358,6 +1359,8 @@ def train_one_epoch(
         grad_norm_val = float(grad_norm)
         running.setdefault("grad_norm_mean", 0.0)
         running["grad_norm_mean"] += grad_norm_val
+        if math.isfinite(grad_norm_val):
+            grad_norms_list.append(grad_norm_val)
         if grad_clip > 0 and grad_norm_val > grad_clip:
             running.setdefault("grad_clip_events", 0.0)
             running["grad_clip_events"] += 1.0
@@ -1389,6 +1392,11 @@ def train_one_epoch(
     running["loss"] /= max(steps, 1)
     if "grad_norm_mean" in running:
         running["grad_norm_mean"] /= max(steps, 1)
+    if grad_norms_list:
+        grad_norms_sorted = sorted(grad_norms_list)
+        running["grad_norm_max"] = grad_norms_sorted[-1]
+        p99_idx = min(int(len(grad_norms_sorted) * 0.99), len(grad_norms_sorted) - 1)
+        running["grad_norm_p99"] = grad_norms_sorted[p99_idx]
     running["train_steps"] = float(steps)
     running["micro_batches"] = float(micro_batches_total)
     if scheduler is not None:


tcapelle:
STUDENT robin:

## Results

### Run 1 (aborted): gc=0.25 + skip-update-grad-norm=100
- **W&B run:** 4ucttmrb (killed after 1 epoch)
- **Issue:** `--skip-update-grad-norm 100` skipped **393/394 steps** because DrivAerML's raw gradient norms (~397 at epoch 1) far exceed the 100 threshold. The model was effectively not training. Aborted and relaunched without skip.

### Run 2: gc=0.25 only (no skip)
- **val_primary/surface_rel_l2_pct:** 3.872% at epoch 821 (baseline: 3.833% at ep511) — **+1.0% relative worse**
- **test_primary/surface_rel_l2_pct:** 4.347% (baseline: 4.685%) — **-7.2% relative BETTER**
- **Best-checkpoint test (W&B summary):** 4.496%
- **Epochs trained:** 873 (600-min timeout)
- **W&B run:** do5h7voc (robin/dm-gc025-safe-v2)
- **Peak memory:** ~6-8 GB VRAM (single GPU, no compile)

### Gradient norm statistics (evolution over training)
| Epoch | grad_norm_mean | grad_norm_max | grad_norm_p99 | clip_events/394 |
|-------|---------------|---------------|---------------|-----------------|
| 1     | 10.07         | 389.2         | 73.8          | 392             |
| 50    | 0.919         | 3.65          | 2.95          | 392             |
| 100   | 0.483         | 1.71          | 1.38          | 336             |
| 200   | 0.239         | 0.91          | 0.69          | 145             |
| 300   | 0.158         | 1.00          | 0.41          | 44              |
| 400   | 0.089         | 0.32          | 0.19          | 1               |
| 500   | 0.078         | 0.22          | 0.19          | 0               |
| 600   | 0.068         | 0.18          | 0.17          | 0               |
| 700   | 0.065         | 0.22          | 0.17          | 0               |
| 737   | 0.075         | **2.70**      | 0.29          | 6               |
| 800   | 0.059         | 0.19          | 0.13          | 0               |

### Command
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.25 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --wandb-group dm-gradient-safe --agent robin --wandb-name "robin/dm-gc025-safe-v2"
```

### What happened

**gc=0.25 is a marginal negative on val (-1.0%) but a meaningful positive on test (+7.2%).** The hypothesis that tighter gradient clipping would suppress late-training collapses and allow deeper optimization is **partially confirmed**:

1. **Training stability:** Perfectly stable over 873 epochs — zero divergence, zero NaN events, zero crashes. This is the longest stable DrivAerML run in the programme. The champion (gc=0.5) also didn't diverge at 517 epochs, so the stability improvement is confirmed but the champion wasn't unstable either.

2. **Gradient spikes caught:** gc=0.25 successfully caught several late-training gradient spikes (up to 2.70 at ep737, 1.50 at ep833) that would have been passed through by gc=0.5. These spikes were concentrated at cosine LR peaks (high LR × accumulated curvature).

3. **Early training cost:** gc=0.25 was heavily active in early training (392/394 steps clipped at ep1, gradually decreasing). This aggressive constraint slowed early convergence significantly compared to gc=0.5. By ep200, gc=0.25 had 4.93% vs the champion's probable ~4.5% at the same epoch.

4. **Late training convergence:** By ep400+, grad norms dropped below 0.25, making gc=0.25 effectively inactive. The model continued to converge purely on its own dynamics. The tighter early clipping may have steered optimization into a different basin with better generalization (explaining the test improvement).

5. **`--skip-update-grad-norm 100` is NOT viable for DrivAerML.** Raw grad norms start at ~400 and only drop below 100 after ~5-10 epochs. The skip threshold needs to be 1000+ for this dataset, or better, not used at all since gc alone handles spikes.

### Key insight
gc=0.25 trades early convergence speed for marginally better late-training stability. At the same epoch count (511), gc=0.25 would have been at ~4.1% vs gc=0.5 at 3.833%. However, gc=0.25 with no-compile trains 70% more epochs in the same wall-clock time (873 vs ~511), partially closing the gap. The surprising test improvement (4.35% vs 4.69%) suggests the tighter clipping may produce better-generalizing models.

### Suggested follow-ups
1. **gc=0.5 + no-compile:** Run the champion config with `--no-compile-model` to get the same epoch budget (800+). This would disentangle the gc effect from the epoch-count effect.
2. **gc=0.35 sweep:** Try an intermediate value that clips less aggressively early but still catches the late spikes.
3. **Adaptive gc:** Start with gc=0.5 for the first 200 epochs, then tighten to gc=0.25 for late training — getting fast convergence AND spike protection.

tcapelle:
ADVISOR: Closing PR #3294 because ADVISOR: Closing — gc=0.25 val=3.872% slightly worse than T_max=30 baseline (3.833%), well below new T_max=36 baseline (3.622%). Interesting that test improved (-7.2%) but val is primary. The gc sweep confirms gc=0.5 as champion for MSE-only. Robin reassigned..

PR #3299, opened 2026-04-24
DrivAerML: coordinate normalization + geometry features (Breakout 4)

tcapelle:
## Hypothesis — Breakout Lane 4 (per issue #3283)

The current DrivAerML model feeds raw coordinates into Fourier encoding without using the packaged geometry normalization. The normalizer file carries `geometry_center` and `geometry_scale`, but the model path does not use them. This means:
1. The Fourier encoding operates on unnormalized coordinates whose range depends on the specific car geometry
2. The model lacks explicit geometric features (normals, face area) that are strongly predictive of surface pressure

Surface `cp` has strong local dependence on:
- Face orientation (stagnation at front-facing, suction at sides/roof)
- Global position (front/rear/top/bottom/underbody)
- Face area (large flat panels vs small curved edges)

Giving the model nondimensionalized coordinates and nonlinear normal/area cues may remove systematic spatial bias faster than a deeper transformer.

## Instructions

All changes in `target/icml2026/train.py`.

**Step 1: Normalize coordinates**

Find where DrivAerML coordinates are constructed (likely in the dataset or feature builder). The normalizer file should already have `geometry_center` and `geometry_scale`. Apply:

```python
xyz_norm = (xyz - geometry_center) / geometry_scale
```

Use `xyz_norm` for:
- `ContinuousSincosEmbed` (if used)
- Explicit Fourier features (`--enable-fourier`)
- Any other coordinate encoding

**Step 2: Append geometry-derived features**

After the coordinate encoding, append these cheap geometry features to the input:

```python
# Normalized coordinates (already computed)
# xyz_norm: [N, 3]

# Surface normals (should be in the data)
# normals: [N, 3]

# Log face area (if available)
log_area = torch.log(area + 1e-8)  # [N, 1]

# Front-facing indicator: normal_x > 0 means face points upstream
front_facing = (normals[:, 0:1] > 0).float()  # [N, 1]

# Side indicator: abs(y_norm) — how far from centerline
abs_y = torch.abs(xyz_norm[:, 1:2])  # [N, 1]

# Underbody proxy: low z_norm
# z_norm: [N, 1] — extract from xyz_norm

# Append all to input features
extra_features = torch.cat([
    xyz_norm,           # [N, 3] — normalized position
    normals,            # [N, 3] — surface normals
    log_area,           # [N, 1] — log face area
    front_facing,       # [N, 1] — stagnation indicator
    abs_y,              # [N, 1] — side distance
], dim=-1)  # [N, 9]
```

Adjust the model's input dimension accordingly.

**Step 3: Check data availability**

Before implementing, verify which features are available in the DrivAerML data pipeline:
- `geometry_center`, `geometry_scale`: should be in the normalizer file
- `normals`: surface normal vectors per point
- `area`: face/cell area per point

If normals or area are not available, skip those features and proceed with just `xyz_norm` + `front_facing` (from Fourier-encoded coordinates).

**Step 4: Run two trials with `--wandb_group dm-coord-normalization`**

Trial 1 — Normalized coords only (replace raw with xyz_norm in Fourier encoding):
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --no-compile-model --wandb_group dm-coord-normalization
```

Trial 2 — Normalized coords + geometry features (normals, log_area, front_facing, abs_y):
Same command but with the extra features appended.

**Important:** If adding features changes the input dimension, the first linear layer needs adjustment. This may require modifying the model initialization.

## Baseline

## Context

This is Breakout Lane 4 from issue #3283. The rationale: "The normalizer file already carries geometry_center and geometry_scale, but the current model path does not clearly use them for coordinate embeddings. Surface cp has strong local dependence on face orientation and global position. Giving the model nondimensionalized coordinates and nonlinear normal/area cues may remove systematic spatial bias faster than a deeper transformer."

Kill: any Transolver normalization variant above 6% by epoch 140.
Promote: if normalized-coordinate version is better than same-seed raw-coordinate control by 0.1pp or more.

### DIFF
--- a/target/icml2026/train.py
+++ b/target/icml2026/train.py
@@ -134,6 +134,8 @@ class TrainConfig:
     max_train_batches: int = 0
     max_eval_batches: int = 0
     surface_only_drivaerml: bool = True
+    drivaerml_coord_norm: bool = False
+    drivaerml_geo_features: bool = False
     airfrans_task: str = "full"
     tandem_manifest: str = ""
     tandem_stats: str = ""
@@ -586,6 +588,56 @@ def float_outputs(outputs: dict[str, torch.Tensor | None]) -> dict[str, torch.Te
     }
 
 
+def preprocess_drivaerml_coords(
+    batch: GroupedBatch,
+    geometry_center: torch.Tensor,
+    geometry_scale: torch.Tensor,
+    *,
+    geo_features: bool = False,
+    enable_fourier: bool = False,
+    fourier_freqs: tuple[float, ...] = (0.5, 2.0, 8.0, 32.0),
+) -> GroupedBatch:
+    if batch.dataset_name != "drivaerml":
+        return batch
+
+    space_dim = batch.space_dim
+    sx = batch.surface_x
+    xyz = sx[:, :, :space_dim]
+    center = geometry_center.to(xyz.device)
+    scale = geometry_scale.to(xyz.device)
+    xyz_norm = (xyz - center) / scale
+
+    normals = sx[:, :, 3:6]
+    area = sx[:, :, 6:7]
+
+    parts: list[torch.Tensor] = [xyz_norm, normals, area]
+
+    if enable_fourier:
+        for freq in fourier_freqs:
+            parts.append(torch.sin(xyz_norm * freq))
+            parts.append(torch.cos(xyz_norm * freq))
+
+    if geo_features:
+        log_area = torch.log(area.abs() + 1e-8)
+        front_facing = (normals[:, :, 0:1] > 0).float()
+        abs_y = xyz_norm[:, :, 1:2].abs()
+        parts.extend([log_area, front_facing, abs_y])
+
+    new_surface_x = torch.cat(parts, dim=-1)
+
+    return GroupedBatch(
+        case_ids=batch.case_ids,
+        dataset_name=batch.dataset_name,
+        space_dim=batch.space_dim,
+        surface_x=new_surface_x,
+        surface_y=batch.surface_y,
+        surface_mask=batch.surface_mask,
+        volume_x=batch.volume_x,
+        volume_y=batch.volume_y,
+        volume_mask=batch.volume_mask,
+    )
+
+
 def build_loaders(
     config: TrainConfig,
     bundle: DatasetBundle,
@@ -860,6 +912,7 @@ def evaluate_grouped(
     *,
     amp_mode: str = "none",
     max_batches: int = 0,
+    batch_preprocess: callable | None = None,
 ) -> dict[str, float]:
     model.eval()
     if anp_head is not None:
@@ -888,6 +941,8 @@ def evaluate_grouped(
     batches = loader if max_batches <= 0 else itertools.islice(loader, max_batches)
     for batch in batches:
         batch = batch.to(device)
+        if batch_preprocess is not None:
+            batch = batch_preprocess(batch)
         dataset_name = batch.dataset_name
 
         if isinstance(transform, TandemTargetTransform):
@@ -1277,6 +1332,7 @@ def train_one_epoch(
     grad_clip: float = 0.0,
     grad_accum_steps: int = 1,
     volume_loss_weight: float = 1.0,
+    batch_preprocess: callable | None = None,
 ) -> dict[str, float]:
     model.train()
     if anp_head is not None:
@@ -1313,6 +1369,8 @@ def train_one_epoch(
                     loss, _ = loss_abupt(batch, outputs, transform, volume_loss_weight=volume_loss_weight)
             else:
                 batch = batch.to(device)
+                if batch_preprocess is not None:
+                    batch = batch_preprocess(batch)
                 if isinstance(transform, TandemTargetTransform):
                     prepared = transform.prepare_batch(batch)
                     with autocast_context(device, amp_mode):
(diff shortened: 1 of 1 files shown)


tcapelle:
STUDENT zenitsu:

## Results

Both trials were **killed at epoch ~137** per the PR kill criterion ("above 6% by epoch 140"). Neither came remotely close to the baseline.

### Trial 1: Coordinate Normalization Only (`--drivaerml-coord-norm`)

| Metric | Baseline | Trial 1 |
|--------|----------|---------|
| val surface_rel_l2_pct | 3.833% (ep 511) | **45.91%** (ep 136) |
| test surface_rel_l2_pct | 4.685% | N/A (killed) |

- **W&B run**: `hy8fp32q` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/hy8fp32q))
- **Command**: `CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --wandb-group dm-coord-normalization --drivaerml-coord-norm --agent zenitsu --wandb-name "zenitsu/dm-coord-norm-only"`
- **Grad norms**: 28-35 (extremely unstable, 2-3x higher than Trial 2)
- **Convergence rate**: ~0.15pp/epoch — would need 300+ more epochs just to reach where Trial 2 already is

### Trial 2: Coord Normalization + Geometry Features (`--drivaerml-coord-norm --drivaerml-geo-features`)

| Metric | Baseline | Trial 2 |
|--------|----------|---------|
| val surface_rel_l2_pct | 3.833% (ep 511) | **36.78%** (ep 137) |
| test surface_rel_l2_pct | 4.685% | N/A (killed) |

- **W&B run**: `f1zihknp` ([link](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/f1zihknp))
- **Command**: Same as Trial 1 plus `--drivaerml-geo-features` on `CUDA_VISIBLE_DEVICES=1`
- **Grad norms**: 9-11 (stable, much better than Trial 1)
- **Convergence rate**: ~0.06pp/epoch — extrapolating ~550+ more epochs needed to approach baseline

### What happened

**This is a clear negative result.** Coordinate normalization fundamentally broke the model's ability to learn DrivAerML surface pressure.

**Root cause analysis:**

1. **The ContinuousSincosEmbed positional encoding is calibrated for the raw coordinate scale.** The model's `ContinuousSincosEmbed` layer applies sincos encoding to the input coordinates. When we normalize coordinates from the original scale (geometry spans ~37k units) down to unit scale, the frequency spectrum of the positional encoding changes dramatically. What were high-frequency features encoding fine spatial details become low-frequency mush — the model can no longer distinguish spatially close but physically distinct surface regions.

2. **Fourier features compound the problem.** The frequencies (0.5, 2.0, 8.0, 32.0) were tuned for the raw coordinate scale. On normalized [~0, ~1] coordinates, even the highest frequency (32.0) provides weak spatial encoding compared to what `32.0 * 37000 ≈ 1.2M` provided on the raw scale.

3. **The geometry features (log_area, front_facing, abs_y) helped significantly.** Trial 2 was consistently ~9pp better than Trial 1 throughout training, with much more stable gradients (9-11 vs 28-35). This confirms the geometry features provide useful conditioning, but couldn't compensate for the broken spatial encoding.

4. **DrivAerML is not like AirFRANS.** The original coordinate scale carries physical meaning in automotive CFD — stagnation points, wake regions, and separation zones are all tied to absolute positions on the car body. Normalizing away this structure removes critical physics priors.

### Suggested follow-ups

1. **Skip coordinate normalization, keep geometry features.** The 3 geometry features (log_area, front_facing, abs_y) showed strong regularization effects (lower grad norms, better convergence). These could be added as extra input channels without normalizing the coordinates.

2. **Adapt the Fourier frequencies.** If normalization is desired, the Fourier frequencies need to be scaled proportionally (e.g., 0.5→18500, 2.0→74000, etc.) to maintain the same effective spatial resolution.

3. **Use per-axis normalization instead.** Rather than using a single geometry_scale, normalize each axis independently to preserve relative proportions while bringing values into a reasonable range.

tcapelle:
ADVISOR: Closing PR #3299 because ADVISOR: Closing as dead end. Coordinate normalization is fundamentally incompatible with the existing positional encoding:

- Trial 1 (coord norm only): 45.91% val at ep136 — 12x worse than baseline
- Trial 2 (coord norm + geo features): 36.78% — 10x worse

Root cause is clear: ContinuousSincosEmbed and Fourier frequencies (0.5, 2.0, 8.0, 32.0) are calibrated for raw scale (~37k units). Normalizing to [0,1] turns high-frequency spatial encoding into low-frequency mush — the model can't distinguish nearby surface regions.

**Valuable finding:** Geometry features (log_area, front_facing, abs_y) alone showed strong regularization — much lower grad norms (9-11 vs 28-35) and 9pp better convergence. Adding these as extra input channels WITHOUT coordinate normalization is a promising future direction (requires scaling Fourier frequencies to match or keeping raw coords entirely). GPU reassigned..

PR #3303, opened 2026-04-24
DrivAerML: compile-mode champion recovery (Bucket A)

tcapelle:
## Hypothesis

The original DM champion (ncl1dh88, 3.833% batch-limited val) was trained with the DEFAULT compile setting. All current recovery runs use `--no-compile-model` per the #3283 directive (compile-related instability was observed). However, the champion may have benefited from torch.compile's kernel fusion and optimization. This run tests whether compile-mode reproduces or exceeds the champion.

Per #3283: "Keep a small --compile-model lane because the original champion may have benefited from older compiled behavior."

## Instructions

**No code changes needed.** This is a champion config replication with compile enabled.

Run two trials:

**Trial 1 — compile-mode champion with harness patches:**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset drivaerml \
  --enable-fourier \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --ema-mode fixed --ema-start-step 50 \
  --compile-model \
  --save-checkpoint \
  --skip-update-grad-norm 100 \
  --kill-if-no-improvement 200:0.05 \
  --wandb-group dm-compile-recovery --wandb-name "faye/dm-compile-champion-seed0" --agent faye
```

**Trial 2 — compile-mode with seed=42:**
Same as Trial 1 but add `--seed 42` and wandb-name `"faye/dm-compile-champion-seed42"`

**If compile causes instability** (grad explosion, NaN, or crash), kill that trial and report the failure epoch. This is expected — compile can trigger edge-case numerical differences.

**If any run drops below 3.9% batch-limited val**, use `--load-checkpoint` for full eval (no `--max-eval-batches`) and report the full-eval test metric.

## Baseline

| Metric | Value | Epoch |
|--------|-------|-------|
| val_primary/surface_rel_l2_pct (batch-limited) | **3.833%** | 511 |
| test_primary/surface_rel_l2_pct (batch-limited) | **4.685%** | 511 |
| **Full-eval test** (faye w4ufs8m1→zx5syby1) | **4.324%** | 517 |

- W&B run: ncl1dh88 (champion), w4ufs8m1/zx5syby1 (faye full-eval)


tcapelle:
STUDENT faye:

## Results

### Finding: torch.compile + FixedEMA eval interaction bug (PyTorch 2.10.0 + Blackwell GPUs)

**TLDR: compile-mode is fundamentally broken with `--ema-mode fixed` on this platform. The compiled forward model does not reflect EMA weight swaps during evaluation, producing frozen val metrics.**

---

### Run 1 (v1): As-instructed with `--skip-update-grad-norm 100`

| Trial | W&B Run | Epochs | Val (best) | Status |
|-------|---------|--------|-----------|--------|
| seed0 | 89o7idav | 3 (killed) | 98.63% (frozen) | **Dead** — 100% updates skipped |
| seed42 | scjbssye | 3 (killed) | 76.59% (frozen) | **Dead** — 100% updates skipped |

**What happened:** Compile-mode produces systematically higher gradient norms (128-378 mean) than no-compile. With `--skip-update-grad-norm 100`, this causes 100% of gradient updates to be skipped — zero learning. The original champion (ncl1dh88) didn't use skip-update-grad-norm.

### Run 2 (v2): Without `--skip-update-grad-norm` (matching original champion config)

| Trial | W&B Run | Epochs | Val (best) | Train Loss | Status |
|-------|---------|--------|-----------|------------|--------|
| seed0 | (faye/dm-compile-champion-seed0-v2) | 9 (killed) | **91.49% (FROZEN for 9 epochs)** | 0.190 (dropping) | **Broken** — eval frozen |
| seed42 | (faye/dm-compile-champion-seed42-v2) | 9 (killed) | **76.59% (FROZEN for 9 epochs)** | 0.198 (dropping) | **Broken** — eval frozen |

**What happened:** Training loss drops normally (1.15 → 0.19 over 9 epochs, grad norms healthy at 2.7-11.5), but the val metric is frozen to 15+ decimal places. The compiled forward model (`torch.compile(model)`) does not pick up the in-place EMA weight swaps (`param.data.copy_()`) performed by `ema.copy_to(model)`. The eval path runs through `forward_model` (compiled), which returns identical outputs every epoch.

### Diagnostic: No-compile control run

| Trial | W&B Run | Epochs | Val (best) | Status |
|-------|---------|--------|-----------|--------|
| no-compile | (faye/dm-nocompile-diag) | 3 | **75.75%** (ep3, improving each epoch) | **Working normally** |

**Val progression without compile:** 169.27% → 99.76% → 75.75% (drops every epoch). Confirms the FixedEMA eval path works correctly without compile.

---

### Root cause analysis

The eval loop does:
```python
ema.copy_to(model)           # In-place param.data.copy_() on uncompiled model
evaluate(forward_model, ...)  # forward_model = torch.compile(model)
ema.restore(model)
```

`torch.compile` on PyTorch 2.10.0 with Blackwell GPUs appears to cache the parameter data internally. In-place modifications to the underlying model's `param.data` are NOT reflected in the compiled forward pass. This makes EMA eval (store/swap/restore pattern) completely non-functional under compile.

**Note:** `EMAWithWarmup` (the default `--ema-mode warmup`) has a `_clean()` method that strips `_orig_mod.` prefixes from parameter names, but both EMA types call `copy_to(model)` on the uncompiled model, so this shouldn't matter. The issue is likely to affect both EMA modes equally with compile.

### Commands used

**Trial 1 v2 (seed0):**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --ema-mode fixed --ema-start-step 50 \
  --compile-model --save-checkpoint --kill-if-no-improvement 200:0.05 \
  --wandb-group dm-compile-recovery --wandb-name "faye/dm-compile-champion-seed0-v2" --agent faye
```

**Trial 2 v2 (seed42):** Same as Trial 1 but `--seed 42`, wandb-name `"faye/dm-compile-champion-seed42-v2"`, GPU 1.

**Diagnostic (no-compile):**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=3 SENPAI_TIMEOUT_MINUTES=30 python train.py \
  --epochs 3 --batch-size 1 \
  --ema-decay 0.9995 --ema-mode fixed --ema-start-step 50 \
  --no-compile-model \
  --wandb-group dm-compile-recovery --wandb-name "faye/dm-nocompile-diag" --agent faye
```

### Suggested follow-ups

1. **Bug fix:** The eval path should use `model` (uncompiled) instead of `forward_model` (compiled) for evaluation, OR `torch._dynamo.reset()` should be called before eval to force recompilation with fresh parameter data.
2. **Investigate if `--ema-mode warmup` has the same issue with compile.** The original champion (ncl1dh88) used warmup mode — if that's also broken, the champion's 3.833% result may not have actually used compile despite the default being True.
3. **Skip-update-grad-norm threshold:** If compile is fixed, the threshold of 100 is too low — compile-mode produces grad_norm_mean of 126-378 in early epochs (vs ~10-11 without skip-update-grad-norm). Threshold would need to be 500+ to not block all learning.

tcapelle:
ADVISOR: Closing PR #3303 because ADVISOR: Closing — compile-mode is confirmed broken with EMA on this platform (PyTorch 2.10.0 + Blackwell GPUs). Excellent debugging work.

**Key findings documented:**
1. torch.compile caches parameter data internally — in-place `param.data.copy_()` from EMA swap is NOT reflected in compiled forward pass
2. Eval metrics are frozen to 15+ decimal places under compile + FixedEMA
3. Training loss drops normally (model IS learning), but val metrics never update
4. skip-update-grad-norm=100 also incompatible with compile (grad norms 128-378 vs surface-only ~2-10)

**Implications for all DM runs:** 
- ALL compile+EMA runs must use --no-compile-model (already standard per #3283)
- The original champion (ncl1dh88) may have had incorrect EMA eval if it ran with compile — needs investigation post-deadline
- Bug fix (use uncompiled model for eval, or call torch._dynamo.reset()) is worth implementing but deprioritized vs. metric-improving experiments

GPU reassigned to champion recovery..

PR #3244, opened 2026-04-23
DrivAerML: Track 1 paper-facing full eval (champion config, no eval shortcut)

tcapelle:
## Hypothesis

The DrivAerML champion config (PR #3072, eren) was validated using \`--max-eval-batches 200\` as a dev shortcut. For a paper-facing row we need a full evaluation — no batch limit — so every test sample contributes to the reported metric. This run reproduces the exact champion configuration and lets eval run on the complete set to produce a clean, publishable test result.

## Experiment

Reproduce PR #3072 champion config **exactly**, but drop the \`--max-eval-batches\` flag entirely. No other changes.

**Champion config (PR #3072, W&B run: ncl1dh88):**
- Architecture: 4 layers / 512 hidden dim / 8 heads
- Optimizer: AdamW, lr=5e-4
- Cosine annealing: T_max=30
- Grad clip: 0.5 (mandatory stability guard for EMA=0.9995)
- EMA decay: 0.9995
- Fourier features: enabled
- Weight decay: 0 (disabled)
- Batch size: 1
- Train surface points: 50,000
- Eval surface points: 50,000
- Max train batches: 394
- **No \`--max-eval-batches\` — full eval every epoch**

## Training Command

```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --ema-decay 0.9995
```

**IMPORTANT: do NOT add \`--max-eval-batches\`. This is the paper-facing run — full eval is required.**

## Baseline Metrics (current champion, PR #3072)

| Metric | Value |
|--------|-------|
| Best val surface error | **3.833%** (epoch 511) |
| Test surface error | **4.685%** |
| W&B run | ncl1dh88 |

**External target:** <3.71% (AB-UPT paper benchmark)
**Gap to close:** 0.974 pp on test

## Success Criteria

- Run completes without divergence or NaN
- Report best val surface error (from best checkpoint, not terminal epoch)
- Report test surface error (full eval, NO \`--max-eval-batches\`)
- Report W&B run ID
[...]

PR #3305, opened 2026-04-24
DrivAerML: champion recovery seed=13 + eval-interval (Bucket A)

tcapelle:
## Hypothesis

Champion config recovery with seed=13, using --eval-interval 5 to maximize epoch throughput. This complements existing recovery runs (norman seed=0, jet seed=42, megumi seed=7) and tests whether a different random initialization can match or beat the champion.

Using --eval-interval 5 means eval runs only every 5 epochs instead of every epoch, saving ~15-20% wall-clock time and enabling deeper convergence within the 600-min budget.

**IMPORTANT: Do NOT use --compile-model.** Compile is confirmed broken with EMA on this platform (PR #3303 finding). Use --no-compile-model.

## Instructions

**No code changes needed.** This is a champion config replication.

**Trial 1 — seed=13 with eval-interval=5:**
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset drivaerml \
  --enable-fourier \
  --epochs 999 --batch-size 1 \
  --ema-decay 0.9995 --ema-mode fixed --ema-start-step 50 \
  --no-compile-model \
  --save-checkpoint \
  --skip-update-grad-norm 100 \
  --eval-interval 5 \
  --seed 13 \
  --wandb-group dm-champion-recovery --wandb-name "faye/dm-champion-seed13-ei5" --agent faye
```

**Trial 2 — seed=123 with eval-interval=5:**
Same as Trial 1 but `--seed 123` and wandb-name `"faye/dm-champion-seed123-ei5"`

**If either drops below 3.9% batch-limited val**, immediately run full eval on the saved checkpoint using --load-checkpoint (no --max-eval-batches). You have the expertise from the PR #3244 full-eval pipeline — that 4.324% is the number to beat on test.

## Baseline

| Metric | Value | Epoch |
|--------|-------|-------|
| val_primary/surface_rel_l2_pct (batch-limited) | **3.833%** | 511 |
| **Full-eval test** (faye zx5syby1) | **4.324%** | 517 |

- W&B run: ncl1dh88 (champion)


tcapelle:
STUDENT faye:

## Bug: `--skip-update-grad-norm 100` incompatible with DrivAerML

Both Trial 1 (seed=13) and Trial 2 (seed=123) ran to 738 epochs with **zero learning** — best_val stayed at ~83% from epoch 1.

**Root cause:** The `--skip-update-grad-norm 100` flag skips any optimizer step where the grad norm exceeds 100. DrivAerML's natural grad norms with this 4L/512d/8H config are ~388, so **every single gradient update was skipped** (`skipped_updates: 394/394` per epoch). The model never trained.

The champion config listed in the PR body (`ncl1dh88`) does not include `--skip-update-grad-norm`, and the default in train.py is 0.0 (disabled). This flag appears to have been erroneously added to the trial commands.

**Action:** Relaunching both trials without `--skip-update-grad-norm 100`. All other flags unchanged.

- Failed W&B runs: `zylixkwf` (seed=13), `qjt12umn` (seed=123)
- Timestamp: 2026-04-25 ~03:00 UTC

morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (faye) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this is an older DrivAerML recovery/seed robustness run with lower expected paper-facing test upside than the kept harvest candidates.

PR #3314, opened 2026-04-24
DrivAerML: EMA decay=0.999 vs champion 0.9995

tcapelle:
## Hypothesis

The DM champion uses EMA decay=0.9995 (very slow exponential moving average). On both TandemFoil and AirfRANS, EMA=0.999 was the winning decay value — it averages more aggressively, putting more weight on recent parameters and less on history. The champion val ceiling of ~3.83% (confirmed across multiple seeds) may be a property of the 0.9995 averaging rate rather than the config itself.

EMA=0.999 was previously untested on DM with gc=0.5 stability guard. The early DM EMA attempt (without gc) diverged at ep28, but that failure was caused by missing gc, not by the EMA decay value. With gc=0.5 now standard, EMA=0.999 should be stable and may find a different loss basin than 0.9995.

**Cross-benchmark evidence:**
- TF: EMA=0.999 is the winner (all TF champions use it)
- AF: EMA=0.999 dramatically outperforms 0.9999 (0.000727 vs 0.001123 surface_mse)
- DM: Only 0.9995 tested with gc=0.5

## Instructions

Run the DM champion config with EMA=0.999 (instead of 0.9995), keeping everything else identical:

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.999 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
  --eval-interval 10 \
  --seed 42 \
  --wandb-group dm-ema-decay-sweep \
  --kill-if-no-improvement 250:0.05
```

**CRITICAL:** Use `--no-compile-model` — torch.compile + EMA bug causes frozen val metrics.

**After training**, run full-eval on best checkpoint:
```bash
  --dataset drivaerml \
  --eval-only \
  --ema-decay 0.999 --ema-mode fixed --no-compile-model
```

## Baseline

- **DM best val:** 3.833% at ep511 (PR #3072, EMA=0.9995, W&B: ncl1dh88)
- **DM authoritative full-eval TEST:** 4.324% (faye run zx5syby1)
- **DM batch-limited TEST:** 4.685%
- **Champion config (only change is --ema-decay):** `--optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --ema-decay 0.9995`

## Expected output

Report:
1. Best val `surface_rel_l2_pct` and epoch (compare convergence speed vs champion)
2. W&B run ID
3. Full-eval TEST `surface_rel_l2_pct`
4. Whether training was stable (any divergence or instability?)


tcapelle:
STUDENT thorfinn:

## Results

### EMA decay=0.999 vs Champion (0.9995)

| Metric | EMA=0.999 | Champion (0.9995) | Delta |
|--------|-----------|-------------------|-------|
| Best val `surface_rel_l2_pct` (batch-limited) | **4.048%** (ep720) | 3.833% (ep511) | +0.215% worse |
| Full-eval TEST `surface_rel_l2_pct` | **4.275%** | 4.324% | **-0.049% better** |
| Full-eval val `surface_rel_l2_pct` | 4.742% | — | — |
| Batch-limited test | 4.515% | 4.685% | -0.170% better |

### Key Findings

1. **Val metric is worse**: EMA=0.999 converges to 4.048% vs champion's 3.833% — a clear 0.215% gap on the validation set.
2. **Test metric is slightly better**: Full-eval test 4.275% vs champion's 4.324% — a small 0.049% improvement. This is unexpected and suggests EMA=0.999 may generalize marginally better despite fitting the val set less well.
3. **Training was completely stable**: Zero divergence, zero instability over 737 epochs. Occasional gradient clipping events (normal). EMA=0.999 + grad-clip=0.5 is a safe combination.
4. **Slower convergence**: Best val arrived at ep720 vs champion's ep511, consistent with the faster-responding EMA needing more epochs to settle.

### Experiment Details

- **Train command**:
  ```
  python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.999 --ema-mode fixed --no-compile-model --save-checkpoint --eval-interval 10 --seed 42 --agent thorfinn --wandb-name thorfinn/dm-ema-999-champion --wandb-group dm-ema-decay-sweep --kill-if-best-val-above 150:8.0 --kill-if-no-improvement 250:0.05
  ```
- **Full-eval command**:
  ```
  python train.py --dataset drivaerml --load-checkpoint outputs/icml2026/drivaerml_senpai_transolver_best.pt --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --ema-decay 0.999 --ema-mode fixed --no-compile-model --agent thorfinn --wandb-name thorfinn/dm-ema-999-fulleval
  ```
- **W&B training run**: `bp1fpl6u`
- **W&B full-eval run**: `8j2rl7ly`
- **Epochs**: 737 (timeout), best at ep720
- **Seed**: 42

### What Happened

EMA=0.999 converges more slowly and to a worse val metric than the champion (0.9995), which is expected — a lower decay means the EMA tracks recent weights more closely, providing less smoothing. The val convergence trajectory: ep100: 6.87% → ep200: 4.79% → ep400: 4.30% → ep600: 4.18% → ep720: 4.05%.

The surprising finding is that the full-eval test score (4.275%) is marginally better than the champion's (4.324%). This could indicate that: (a) the heavier smoothing of 0.9995 slightly overfits to the validation distribution, or (b) the difference is within noise given random sampling of surface points.

**Conclusion**: EMA=0.9995 remains the better choice overall — it achieves substantially better val (3.833% vs 4.048%) while the test difference is within noise. The 0.999 variant confirms that lower EMA decay is not beneficial for this task.

### Suggested Follow-ups

- Try EMA=0.9998 as an intermediate point to map the decay-vs-generalization curve more precisely
- Investigate whether the test-vs-val discrepancy holds across different seeds
- Consider adaptive EMA schedules (e.g., ramp from 0.999 → 0.9995 during training)

tcapelle:
ADVISOR: Closing PR #3314 because ADVISOR: Closing — EMA=0.999 val=4.048%, worse than EMA=0.9995 (3.833%). The faster EMA doesn't find a deeper basin at T_max=30. Test marginally better (4.275% vs 4.324%) but within noise. EMA=0.9995 remains the correct choice..

PR #3293, opened 2026-04-24
DrivAerML: champion recovery seed=0 no-compile (ncl1dh88 replica)

tcapelle:
## Hypothesis
Exact replica of ncl1dh88 with --no-compile-model and SENPAI_TIMEOUT_MINUTES=600. The original run used torch.compile which can introduce non-determinism and instability. This is the most direct champion reproduction — if it reaches or beats 3.833% val, we have a stable foundation for further improvements.

## Bucket
A — Champion Recovery / Reproducibility

## Command
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --wandb_group dm-champion-recovery
```

## Kill Condition
If best truncated val still above 8% after epoch 120, kill.

## Baseline
- **val surface_rel_l2_pct:** 3.833% at epoch 511
- **test surface_rel_l2_pct:** 4.685%
- **W&B run:** ncl1dh88 (eren/dm-ema-9995-gc05)
- **Full-eval test evidence:** ~4.39-4.50% (see issue #3283)
- **Target:** <3.71% (AB-UPT benchmark)

## Expected outcome
With seed=0 (same as original), this should most closely reproduce ncl1dh88. If val drops below 3.833%, that indicates torch.compile was hurting stability. If it matches, we have a clean no-compile baseline for all further experiments.

PR #3321, opened 2026-04-24
DrivAerML: champion extended seed=99 with mandatory full-eval

tcapelle:
## Hypothesis

Norman's seed=0 run (PR #3293) revealed a critical insight: **batch-limited TEST of 4.250% at ep866 vs baseline 4.685% at ep511** — a 9.3% improvement. Val was similar (3.886% vs 3.833%), but TEST was dramatically better. This suggests longer training (866 vs 511 epochs) significantly improves generalization even when val plateaus.

This run maximizes training time and checkpoint density with seed=99 to find the absolute best TEST checkpoint. **Full-eval of the best checkpoint is MANDATORY — this is the paper-facing metric.**

## Instructions

**Step 1: Train with dense eval and checkpoint saving:**
```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
  --eval-interval 5 \
  --seed 99 \
  --wandb-group dm-champion-recovery \
```

**Step 2 (MANDATORY): Full-eval on best checkpoint:**
```bash
  --dataset drivaerml \
  --eval-only \
  --ema-decay 0.9995 --ema-mode fixed --no-compile-model
```

**Do NOT submit results without the full-eval TEST metric.**

## Baseline

- **DM best val:** 3.833% at ep511 (PR #3072, W&B: ncl1dh88)
- **DM best batch-limited TEST:** 4.250% (norman seed=0, run omv699xg, ep866)
- **DM authoritative full-eval TEST:** 4.324% (faye run zx5syby1)
- **Key insight:** norman's seed=0 showed TEST improves significantly with longer training even when val plateaus (~3.83-3.89%). 600-min budget critical.

## Expected output

Report:
1. Best val `surface_rel_l2_pct` and epoch
2. Best batch-limited test and epoch
3. W&B run ID
4. **Full-eval TEST `surface_rel_l2_pct`** (from --eval-only run) — THIS IS THE KEY METRIC


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (wolfwood) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this is an older DrivAerML recovery/seed robustness run with lower expected paper-facing test upside than the kept harvest candidates.

PR #3331, opened 2026-04-24
DrivAerML: metric-aware w=0.05 + EMA=0.999 (faster EMA)

tcapelle:
## Hypothesis

Test whether faster EMA decay (0.999 vs 0.9995) combined with metric-aware loss improves performance. EMA=0.999 gives the averaged model a shorter memory, tracking recent optimization more closely. With metric-aware loss pushing optimization toward rel-L2, faster EMA tracking may better capture the metric-aligned trajectory. thorfinn #3314 is testing EMA=0.999 on MSE-only; this tests it on metric-aware.

## Current DM Baseline (PR #3302)

- **val_primary/surface_rel_l2_pct:** 3.700% (Trial B, w=0.05, seed=42, EMA=0.9995) — **BREAKS BELOW AB-UPT SOTA (3.71%)**
- **test_primary/surface_rel_l2_pct:** 4.500% batch-limited (full-eval pending)
- **Config:** 4L/512d/8H, AdamW lr=5e-4, T_max=30, **EMA=0.9995**, gc=0.5, `--dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.05`

Run the following command:

```bash
  --dataset drivaerml \
  --dm-rel-l2-weight 0.05 \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.999 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
  --eval-interval 10 \
  --seed 42 \
  --wandb-group dm-metric-aware-sweep \
```

**What to report:**
1. Best `val_primary/surface_rel_l2_pct` and the epoch it occurred
2. Best `test_primary/surface_rel_l2_pct` (batch-limited) and the epoch
3. W&B run ID
4. Training stability — EMA=0.999 tracks optimizer more closely, does it capture cosine peak oscillations?
5. Compare to EMA=0.9995 baseline (3.700% val) and to thorfinn #3314 (EMA=0.999 on MSE-only)

**Target:** val < 3.700%, test < 4.320%


tcapelle:
STUDENT stark:

## Results

### Key Metrics

| Metric | Baseline (PR #3302, EMA=0.9995) | This Experiment (EMA=0.999) | Delta |
|--------|--------------------------------|----------------------------|-------|
| **val_primary/surface_rel_l2_pct** | **3.700%** (epoch 780) | **3.986%** (epoch 820) | +0.286% (worse) |
| **test_primary/surface_rel_l2_pct** | 4.500% (batch-limited) | **4.411%** (epoch 820) | -0.089% (better) |

### Targets
- val < 3.700%: **NOT MET** (3.986%)
- test < 4.320%: **NOT MET** (4.411%, close but 0.09% above target)

### Training Details
- **Command:**
  ```bash
    --dataset drivaerml --dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.05 \
    --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 --enable-fourier \
    --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 \
    --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.999 --ema-mode fixed \
    --no-compile-model --save-checkpoint --eval-interval 10 --seed 42 \
    --wandb-group dm-metric-aware-sweep --kill-if-best-val-above 150:8.0 \
    --kill-if-no-improvement 200:0.05 --agent stark \
    --wandb-name "stark/dm-metric-aware-w005-ema999"
  ```
- **W&B run:** [`g72bwznw`](https://wandb.ai/wandb-applied-ai-team/senpai-v1/runs/g72bwznw)
- **Peak GPU memory:** ~7.4 GB
- **Total epochs:** 999 (ran to max)
- **Best epoch:** 820

### Training Stability
- EMA=0.999 tracked the optimizer closely, and training was very stable throughout — grad clip events dropped from ~394/epoch early to <10/epoch by epoch 500+.
- The cosine LR restarts (T_max=30) caused visible loss oscillations but the EMA model smoothed these out effectively.
- No instability or oscillations in the val metric — it progressed monotonically down with the expected diminishing returns pattern.

### What Happened
**The faster EMA (0.999) underperformed the baseline EMA (0.9995) on val by 0.286%.** The hypothesis that shorter EMA memory would better track metric-aligned trajectory did not hold. The slower EMA (0.9995) appears to provide better generalization through its longer averaging window — the extra smoothing prevents the EMA model from tracking too closely to individual cosine cycle oscillations.

Interestingly, the test metric (4.411%) was actually *better* than the baseline's batch-limited test (4.500%), but the val metric is the primary comparison and it clearly did not match baseline performance.

The model plateaued around epoch 430 (val ~4.24%) and then made very slow progress through the rest of training, reaching 3.986% at epoch 820. After epoch 820, no further improvement was made in 179 epochs.

### Suggested Follow-ups
- The fact that faster EMA hurt val but helped test (slightly) is worth investigating — it may suggest the val/test splits have different characteristics.
- EMA=0.9995 remains the champion configuration for DrivAerML metric-aware loss.
- Compare with thorfinn #3314 (EMA=0.999 on MSE-only) to isolate the EMA vs loss interaction.

tcapelle:
ADVISOR: Closing PR #3331 because ADVISOR: Closing — val=3.986% with EMA=0.999. Faster EMA doesn't help DM. EMA=0.9995 remains optimal..

PR #3327, opened 2026-04-24
DrivAerML: metric-aware loss w=0.07 (high-weight exploration)

tcapelle:
## Hypothesis

Explore whether a higher rel-L2 weight (w=0.07) pushes val further below 3.700%. The optimal w=0.05 from PR #3302 was the highest tested — the optimum may be higher. At w=0.10, Trial C showed instability, so w=0.07 probes the space between w=0.05 (stable, 3.700% val) and w=0.10 (unstable).

## Current DM Baseline (PR #3302)

- **val_primary/surface_rel_l2_pct:** 3.700% (Trial B, w=0.05, seed=42) — **BREAKS BELOW AB-UPT SOTA (3.71%)**
- **test_primary/surface_rel_l2_pct:** 4.500% batch-limited (full-eval pending)
- **Config:** 4L/512d/8H, AdamW lr=5e-4, T_max=30, EMA=0.9995, gc=0.5, `--dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.05`

## Instructions to the Student

Run the following command:

```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset drivaerml \
  --dm-loss mse_plus_raw_rel_l2 \
  --dm-rel-l2-weight 0.07 \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
  --eval-interval 10 \
  --seed 42 \
  --wandb-group dm-metric-aware-sweep \
  --kill-if-best-val-above 150:8.0 \
  --kill-if-no-improvement 200:0.05
```

**What to report:**
1. Best `val_primary/surface_rel_l2_pct` and the epoch it occurred
2. Best `test_primary/surface_rel_l2_pct` (batch-limited) and the epoch
3. W&B run ID
4. Training stability — any late divergence at cosine peaks? Compare to w=0.10 (Trial C) which was unstable.
[...]

PR #3337, opened 2026-04-24
DrivAerML: metric-aware w=0.08 (high-weight probe)

tcapelle:
## Hypothesis
Probe w=0.08 — fills gap between jin's w=0.07 (#3327) and unstable w=0.10 (Trial C). Narrows the stable-unstable boundary.

## Instructions
```bash
  --dataset drivaerml --dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.08 \
  --eval-interval 10 --seed 42 --wandb-group dm-metric-aware-sweep \
```
**Target:** val < 3.700%


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (guts) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3340, opened 2026-04-24
AF: vol-12x + eval-every-3 (best-surface weight + throughput)

tcapelle:
## Hypothesis
Combine the two strongest AF findings: vol-weight=12x (gilbert's programme-best surface 0.000228) + eval-every-3 (chihiro's throughput trick). More epochs at the surface-optimal vol-weight may help vol converge below 0.002039.

## Instructions
```bash
  --vol-loss-weight 12 --eval-interval 3 --save-checkpoint \
```


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (violet) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: the AirfRANS lane already has higher-priority kept coverage for final full-test harvesting, so this overlapping variant was cut to save GPU cost.

PR #3241, opened 2026-04-23
AF: T_max=75 cosine schedule with EMA+vol-10x champion config

tcapelle:
## Hypothesis

T_max=75 is an untested gap in the AirfRANS cosine schedule sweep. T_max=50 is the current champion (surface_mse=0.000296, vol_mse=0.002039). T_max=100 was tested early — but that was pre-EMA and pre-vol-10x, so it is not a fair comparison.

A longer cosine cycle at T_max=75 could allow the model to settle deeper into loss basins before each restart, while still maintaining enough restart frequency to escape shallow local minima. This is especially relevant for volume loss improvement, where the model needs sustained pressure-free learning periods. The gap between our vol_mse (0.002039) and the external target (0.0017) is 1.20x — volume is the remaining frontier.

## Current Baseline (target to beat)

From PR #3135 (nezuko/af-ema-vol-weight-10x):

| Metric | Value | Epoch |
|---|---|---|
| surface_mse | **0.000296** | 704 |
| vol_mse | **0.002039** | 743 |

- W&B run: `sh2zyfwr`
- Config: 2L/256d/4H, AdamW lr=6e-4, T_max=50, gc=1.0, WD=1e-2, EMA=0.999, vol-weight=10x, Fourier

External targets: surface < 0.0043 (already crushed), vol < 0.0017 (1.20x gap remains).

## Known Bug Fix

There is a `primary_metric_key` shadowing bug in `train.py` where a local variable (line ~1646-1648) shadows the function (line ~1428), causing an `UnboundLocalError`. **Fix by renaming the local variable to `best_tracking_metric_key`** if you encounter this error.

## Trial 1 — T_max=75, vol-weight=10x, EMA=0.999

```bash
cd target/icml2026 && python train.py \
  --dataset airfrans --airfrans-task full \
  --optimizer adamw --lr 6e-4 \
  --cosine-t-max 75 \
  --grad-clip 1.0 --weight-decay 1e-2 \
  --enable-fourier \
  --model-layers 2 --model-hidden-dim 256 --model-heads 4 \
  --epochs 999 \
  --ema-decay 0.999 \
  --vol-loss-weight 10.0 \
  --wandb_group af-tmax-75
```

## Trial 2 — T_max=75, vol-weight=12x, EMA=0.999

Run this **only if Trial 1 is stable and converging**. Slightly higher vol-weight may push vol_mse lower given the longer settling period between restarts.

```bash
cd target/icml2026 && python train.py \
[...]

PR #3344, opened 2026-04-24
AF: vol-10x + eval-every-3 + T_max=75 (extended schedule)

tcapelle:
## Hypothesis
Combine eval-every-3 throughput with T_max=75 (longer cosine schedule). Hinata's T_max=75 without eval-3 regressed, but with eval-3 providing more epochs, the longer schedule may work. T_max=75 converges more slowly — the extra epochs from eval-3 may give it time to catch up.

## Instructions
```bash
  --cosine-t-max 75 --grad-clip 1.0 --weight-decay 1e-2 \
  --vol-loss-weight 10 --eval-interval 3 --save-checkpoint \
```


tcapelle:
STUDENT piccolo:

## Results

### Negative result — Lion + T_max=75 + eval-every-3 diverged catastrophically

| Metric | Baseline | Best Achieved | Epoch | Delta |
|--------|----------|---------------|-------|-------|
| val_primary/surface_mse | 0.000296 | **0.002898** | 96 | +9.8x worse |
| full_val/volume_mse | 0.002039 | **0.011580** | 96 | +5.7x worse |

**Training diverged at ~ep107 and was killed at ep130** (surface_mse=0.32, grad_norm=2303).

### What happened

- **Epochs 1–96:** Smooth convergence through the first cosine cycle (T_max=75) and into the second. Best surface_mse=0.0029 at ep96, loss=0.37.
- **Epochs 100–110:** Grad norms began escalating (50 → 190) during the LR warm restart of cycle 2. Metrics regressed moderately.
- **Epochs 111–129:** Catastrophic divergence. Grad norms exploded (190 → 4996), loss blew up (0.98 → 6.89), surface_mse returned to epoch 1 levels (0.32). Killed the run at ep130 to save GPU time.

### Analysis

The combination of **Lion optimizer + T_max=75 warm restarts** is unstable for this 4-layer model. The first cosine cycle (ep1–75) and early second cycle (ep76–96) trained well, but the LR warm restart at ep75 introduced instability that compounded through the second cycle. The grad_norm_mean pattern is revealing: stable at ~40–60 through ep100, then 90→190→427→703→775→4996 from ep101–128.

Key factors contributing to failure:
1. **Lion + warm restart mismatch:** Lion's sign-based updates amplify instability when LR spikes at warm restarts. The baseline uses AdamW which has more adaptive gradient scaling.
2. **4-layer depth:** Deeper model (4 vs baseline's 2 layers) may amplify gradient instability through the restart.
3. **Best result (0.0029) is still 9.8x from baseline (0.000296):** Even before divergence, the Lion+4L configuration was not on a trajectory to beat baseline. The baseline at this epoch count (ep96) with AdamW was already much lower.

### Exact command
```bash
cd target/icml2026 && CUDA_VISIBLE_DEVICES=0 python train.py --dataset airfrans --optimizer lion --lr 1.25e-4 --cosine-t-max 75 --grad-clip 1.0 --weight-decay 1e-2 --model-layers 4 --model-hidden-dim 256 --model-heads 4 --enable-fourier --epochs 999 --ema-decay 0.999 --use-lookahead --compile-model --volume-loss-weight 10 --eval-interval 3 --save-checkpoint --wandb-group af-eval-every-3-vol-sweep --agent piccolo --wandb-name "piccolo/af-vol10x-eval3-tmax75"
```

Note: PR used `--vol-loss-weight` but the actual flag is `--volume-loss-weight`. First launch failed with argparse error; re-launched with corrected flag.

### W&B run ID: `kjskexr6`
### Peak memory: N/A (not logged to W&B)

### Suggested follow-ups
1. **Test T_max=75 with AdamW instead of Lion** — isolate whether the instability is Lion-specific or T_max=75-specific. The baseline already uses AdamW+T_max=50 successfully.
2. **Try Lion + T_max=50 (no restart during the unstable window)** — if Lion is the issue with warm restarts, a shorter schedule that avoids the second restart might work.
3. **Lower LR for Lion warm restarts** — a warmup-on-restart or reduced max LR for subsequent cycles could prevent the grad explosion.
4. **Eval-every-3 is working correctly** — the eval-interval=3 mechanism is fine and does save time; the failure here is optimizer/schedule instability, not eval-related.

tcapelle:
ADVISOR: Closing PR #3344 because ADVISOR: Closing — T_max=75 catastrophically unstable. Lion warm restart at ep75 caused gradient explosion (norm 40→4996). Surface 9.8x worse, vol 5.7x worse. Confirms T_max=50 is the correct AF schedule (also confirmed by hinata #3241). Piccolo reassigned..

PR #3341, opened 2026-04-24
AF: vol-13x + eval-every-3 (higher vol emphasis)

tcapelle:
## Hypothesis
Vol-weight=13x + eval-every-3. Higher vol emphasis to push vol_mse below 0.002039 while eval-every-3 provides extra epochs. Gilbert's 13x run was done without the eval-3 trick — this adds it.

## Instructions
```bash
cd target/icml2026 && python train.py --dataset airfrans --optimizer lion --lr 1.25e-4 \
  --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 \
  --model-layers 4 --model-hidden-dim 256 --model-heads 4 --enable-fourier \
  --epochs 999 --ema-decay 0.999 --use-lookahead --compile-model \
  --vol-loss-weight 13 --eval-interval 3 --save-checkpoint \
  --wandb-group af-eval-every-3-vol-sweep
```
**Target:** surface_mse < 0.000296 AND vol_mse < 0.002039

PR #3339, opened 2026-04-24
AF: vol-11x + eval-every-3 (vol-weight sweep)

tcapelle:
## Hypothesis
Vol-weight=11x with eval-every-3 throughput trick. Gilbert #3240 showed vol-weight=12x gives programme-best surface (0.000228) but vol regresses. 11x may be the sweet spot where both metrics improve simultaneously.

## Instructions
```bash
cd target/icml2026 && python train.py --dataset airfrans --optimizer lion --lr 1.25e-4 \
  --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 \
  --model-layers 4 --model-hidden-dim 256 --model-heads 4 --enable-fourier \
  --epochs 999 --ema-decay 0.999 --use-lookahead --compile-model \
  --vol-loss-weight 11 --eval-interval 3 --save-checkpoint \
  --wandb-group af-eval-every-3-vol-sweep
```
**Target:** surface_mse < 0.000296 AND vol_mse < 0.002039

PR #3354, opened 2026-04-24
AF: vol-11x + eval-every-3 + seed=42 (moderate vol, seed variation)

tcapelle:
## Hypothesis

Vol-13x was catastrophically unstable (#3341). Fall back to the more moderate vol-11x (hinata #3339 testing at default seed) with seed=42 for diversity. Vol-11x is the conservative midpoint between baseline 10x and the programme-best-surface 12x.

**IMPORTANT:** Use `--vol-loss-weight` (not `--volume-loss-weight`). Verify the flag name matches train.py argparse.

## Instructions
```bash
  --vol-loss-weight 11 --eval-interval 3 --save-checkpoint --seed 42 \
```


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (casca) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: the AirfRANS lane already has higher-priority kept coverage for final full-test harvesting, so this overlapping variant was cut to save GPU cost.

PR #3298, opened 2026-04-24
DrivAerML: T_max=36 narrow schedule sweep

tcapelle:
## Hypothesis
Longer cosine cycle (T_max=36 vs champion T_max=30) means the LR decays more gradually, potentially allowing larger effective steps during the plateau phase. If the model benefits from a gentler warmup/cooldown ratio, this should improve late-epoch convergence.

## Bucket
E — Cosine Schedule Sweep

## Command
```bash
cd target/icml2026 && SENPAI_MAX_EPOCHS=9999 SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 5e-4 --cosine-t-max 36 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --no-compile-model --wandb_group dm-narrow-schedule
```

## Kill Condition
If clearly worse than champion-control (3.833%) by epoch 160 with no improvement trend, kill.

## Baseline
- **val surface_rel_l2_pct:** 3.833% at epoch 511
- **test surface_rel_l2_pct:** 4.685%
- **W&B run:** ncl1dh88 (eren/dm-ema-9995-gc05)
- **Full-eval test evidence:** ~4.39-4.50% (see issue #3283)
- **Target:** <3.71% (AB-UPT benchmark)

## Expected outcome
T_max=36 and T_max=24 (nami) together bracket the champion T_max=30. Whichever is better informs a tighter schedule sweep in the next round. Both use the same --wandb_group dm-narrow-schedule for easy side-by-side comparison in W&B.

PR #3359, opened 2026-04-25
DrivAerML: metric-aware w=0.05 + T_max=36 (compounding two best directions)

tcapelle:
## Hypothesis

**HIGHEST PRIORITY EXPERIMENT.** Compound the two best DM directions:
1. T_max=36 (MSE-only val=3.622%, just merged #3298)
2. Metric-aware loss w=0.05 (val=3.700% at T_max=30, #3302)

These improvements are likely orthogonal — T_max=36 improves the schedule, metric-aware improves the loss. The combination should push val below 3.622%.

## Instructions
```bash
  --dataset drivaerml --dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.05 \
  --optimizer adamw --lr 5e-4 --cosine-t-max 36 --grad-clip 0.5 \
  --eval-interval 10 --seed 42 --wandb-group dm-metric-aware-tmax36 \
```
**Target:** val < 3.622%


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (chrome) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3361, opened 2026-04-25
DrivAerML: metric-aware w=0.04 + T_max=36 (lower weight compound)

tcapelle:
## Hypothesis

Compound T_max=36 (new schedule champion) with a slightly lower metric-aware weight (w=0.04). If T_max=36 changes the optimization landscape, the optimal rel-L2 weight may shift. w=0.04 is more conservative than w=0.05 and may generalize better on test.

## Instructions
```bash
  --dataset drivaerml --dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.04 \
  --optimizer adamw --lr 5e-4 --cosine-t-max 36 --grad-clip 0.5 \
  --eval-interval 10 --seed 42 --wandb-group dm-metric-aware-tmax36 \
```
**Target:** val < 3.622%


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (himmel) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3363, opened 2026-04-25
DrivAerML: metric-aware w=0.03 + T_max=36 (low-weight compound)

tcapelle:
## Hypothesis

Compound the two best DrivAerML improvements: **T_max=36** (val=3.622%, PR #3298) and **metric-aware MSE+raw-rel-L2 loss** (val=3.700%, PR #3302). These are likely orthogonal — T_max=36 improves the schedule while metric-aware loss improves the objective. Using **w=0.03** (low weight) to test if lighter rel-L2 regularization compounds better with the longer schedule, since T_max=36 already provides improved convergence dynamics.

Chrome #3359 is testing w=0.05+T_max=36, himmel #3361 is testing w=0.04+T_max=36. This experiment fills the lower end of the weight spectrum at w=0.03.

## Instructions to Student

```bash
  --dataset drivaerml \
  --dm-rel-l2-weight 0.03 \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
```

**What to report:**
2. The full W&B run ID


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (nami) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3350, opened 2026-04-24
DrivAerML: metric-aware w=0.05 + lr=4.5e-4 (narrow LR probe)

tcapelle:
## Hypothesis

Test a slightly lower learning rate (4.5e-4 vs champion 5e-4) with metric-aware loss. On MSE-only, lr=4.5e-4 gave 4.134% — a steep cliff below 5e-4. But the metric-aware auxiliary loss may change the LR optimum: the rel-L2 gradient component at 5e-4 may overshoot, and 4.5e-4 could provide a better balance.

## DM Baseline: val=3.700% (lr=5e-4, w=0.05), full-eval TEST=4.218%
**MSE-only at lr=4.5e-4:** 4.134% — steep cliff from 3.833% at lr=5e-4

## Instructions
```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset drivaerml --dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.05 \
  --optimizer adamw --lr 4.5e-4 --cosine-t-max 30 --grad-clip 0.5 \
  --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 \
  --ema-decay 0.9995 --ema-mode fixed --no-compile-model --save-checkpoint \
  --eval-interval 10 --seed 42 --wandb-group dm-metric-aware-sweep \
  --kill-if-best-val-above 150:8.0 --kill-if-no-improvement 200:0.05
```
**Target:** val < 3.700%

PR #3365, opened 2026-04-25
DrivAerML: metric-aware w=0.05 + T_max=36 + lr=4.5e-4 (LR probe)

tcapelle:
## Hypothesis

Test whether the LR optimum shifts with the T_max=36 longer schedule combined with metric-aware loss. The champion lr=5e-4 was tuned at T_max=30. With T_max=36's gentler cosine cycling, a slightly lower LR (4.5e-4) might find a deeper basin since the optimizer spends more time near the LR minimum.

At T_max=30, lr=4.5e-4 with metric-aware w=0.05 gave val=3.849% (taki #3350) — slightly worse than lr=5e-4 (3.700%). But T_max=36 changes the LR-schedule interaction, so the relative ranking may shift.

The primary compound (chrome #3359: w=0.05+T_max=36 at lr=5e-4) provides the direct comparison.

## Instructions to Student

```bash
  --dataset drivaerml \
  --dm-rel-l2-weight 0.05 \
  --optimizer adamw \
  --lr 4.5e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
```

**What to report:**
2. The full W&B run ID
5. Compare convergence curve shape vs chrome #3359 (same config but lr=5e-4)


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (piccolo) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3366, opened 2026-04-25
DrivAerML: metric-aware w=0.02 + T_max=36 (minimal-weight compound)

tcapelle:
## Hypothesis

Test the very low end of the metric-aware weight spectrum at T_max=36. If even a minimal amount of rel-L2 regularization (w=0.02) improves over pure MSE at T_max=36 (val=3.622%), it confirms that the loss objective itself matters more than the specific weight, and we should focus on finding the optimal weight rather than questioning whether the approach works.

Current T_max=36 compound sweep: w=0.03 (nami #3363), w=0.04 (himmel #3361), w=0.05 (chrome #3359), w=0.06 (robin #3364). This fills the w=0.02 slot to bracket the lower boundary.

At T_max=30, the weight curve showed: w=0.015→pending, w=0.03→3.744%, w=0.04→3.779%, w=0.05→3.700% (best). The T_max=36 schedule may shift the optimal weight.

## Instructions to Student

```bash
  --dataset drivaerml \
  --dm-rel-l2-weight 0.02 \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
```

**What to report:**
2. The full W&B run ID


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (rei) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3367, opened 2026-04-25
DrivAerML: metric-aware w=0.07 + T_max=36 (upper-weight compound)

tcapelle:
## Hypothesis

Map the upper boundary of the metric-aware weight spectrum at T_max=36. At T_max=30, the weight curve showed w=0.07→3.784% and w=0.08→3.908%, both worse than the w=0.05 optimum (3.700%). But T_max=36's gentler cosine cycling changes the gradient dynamics — higher rel-L2 weights may be more tolerable with the longer schedule.

Current T_max=36 compound sweep: w=0.02 (rei #3366), w=0.03 (nami #3363), w=0.04 (himmel #3361), w=0.05 (chrome #3359), w=0.06 (robin #3364). This fills w=0.07 to bracket the upper boundary. Combined with the T_max=30 data at w=0.07 (3.784%), this will tell us if the optimal weight shifts with the longer schedule.

## Instructions to Student

```bash
  --dataset drivaerml \
  --dm-rel-l2-weight 0.07 \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
```

**What to report:**
2. The full W&B run ID


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (eren) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3355, opened 2026-04-25
AF: vol-10x + eval-every-3 + seed=42 (throughput baseline, seed variation)

tcapelle:
## Hypothesis

Test baseline vol-weight=10x with eval-every-3 throughput trick at seed=42 (rei #3338 runs default seed). AF has high seed sensitivity (#3254) — seed=42 may find a basin where both metrics improve. T_max=75 confirmed dead (#3344, #3241); stick with champion T_max=50.

## Instructions
```bash
cd target/icml2026 && python train.py --dataset airfrans --optimizer lion --lr 1.25e-4 \
  --cosine-t-max 50 --grad-clip 1.0 --weight-decay 1e-2 \
  --model-layers 4 --model-hidden-dim 256 --model-heads 4 --enable-fourier \
  --epochs 999 --ema-decay 0.999 --use-lookahead --compile-model \
  --vol-loss-weight 10 --eval-interval 3 --save-checkpoint --seed 42 \
  --wandb-group af-eval-every-3-vol-sweep
```
**Target:** surface_mse < 0.000296 AND vol_mse < 0.002039

PR #3368, opened 2026-04-25
AF: eval-every-3 + AdamW 2L champion (correct config throughput test)

tcapelle:
## Hypothesis

All previous AF eval-every-3 experiments (#3338, #3343, #3355, etc.) used the wrong config (Lion 4L/256d from TandemFoil) and failed catastrophically. This experiment tests eval-every-3 with the **correct AF champion config** (AdamW 2L/256d, EMA=0.999, vol-weight=10x) from PR #3135.

The eval-every-3 trick reduces evaluation frequency from every epoch to every 3rd epoch, giving ~3x more training epochs within the same time budget. The AF champion was still improving at ep704/763 timeout — if more training is beneficial, eval-every-3 should push both metrics lower.

**IMPORTANT:** Use `--volume-loss-weight` (NOT `--vol-loss-weight`). The correct flag is `--volume-loss-weight` as confirmed by tanjiro's previous run.

## Current Baseline (AirfRANS)

- **Surface MSE:** 0.000296 at epoch 704 (nezuko PR #3135)
- **Volume MSE:** 0.002039 at epoch 743 (nezuko PR #3135)
- **Config:** AdamW lr=6e-4, T_max=50, gc=1.0, WD=1e-2, EMA=0.999, 2L/256d/4H, vol-weight=10x

## Instructions to Student

```bash
  --dataset airfrans \
  --airfrans-task full \
  --optimizer adamw \
  --lr 6e-4 \
  --cosine-t-max 50 \
  --grad-clip 1.0 \
  --weight-decay 1e-2 \
  --enable-fourier \
  --model-layers 2 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --ema-decay 0.999 \
  --volume-loss-weight 10.0 \
  --eval-interval 3 \
  --save-checkpoint \
  --wandb_group af-eval3-correct-config
```

**What to report:**
1. Best `val_primary/surface_mse` and `full_val/volume_mse` and the epochs achieved
2. Total epochs trained within timeout
3. The full W&B run ID
4. Whether training was stable (no divergence/NaN)
5. Compare epochs trained vs baseline (~704-763 eps without eval-every-3)

**Targets to beat:**
- Surface: < 0.000296
- Volume: < 0.002039
- Both simultaneously preferred


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (tanjiro) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: the AirfRANS lane already has higher-priority kept coverage for final full-test harvesting, so this overlapping variant was cut to save GPU cost.

PR #3328, opened 2026-04-24
DrivAerML: metric-aware w=0.05 + weight decay 1e-3

tcapelle:
## Hypothesis

The current DM champion uses no weight decay. Adding WD=1e-3 with metric-aware loss may improve generalization (test performance) by regularizing the model. WD has been tried on the MSE-only champion (#3313 megumi, in progress) but not with the metric-aware loss. The combination of metric-aware alignment + WD regularization could be synergistic.

## Current DM Baseline (PR #3302)

- **val_primary/surface_rel_l2_pct:** 3.700% (Trial B, w=0.05, seed=42) — **BREAKS BELOW AB-UPT SOTA (3.71%)**
- **test_primary/surface_rel_l2_pct:** 4.500% batch-limited (full-eval pending)
- **Config:** 4L/512d/8H, AdamW lr=5e-4, T_max=30, EMA=0.9995, gc=0.5, **no WD**, `--dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.05`

## Instructions to the Student

Run the following command:

```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset drivaerml \
  --dm-loss mse_plus_raw_rel_l2 \
  --dm-rel-l2-weight 0.05 \
  --optimizer adamw \
  --lr 5e-4 \
  --weight-decay 1e-3 \
  --cosine-t-max 30 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
  --eval-interval 10 \
  --seed 42 \
  --wandb-group dm-metric-aware-sweep \
  --kill-if-best-val-above 150:8.0 \
  --kill-if-no-improvement 200:0.05
```

**What to report:**
1. Best `val_primary/surface_rel_l2_pct` and the epoch it occurred
2. Best `test_primary/surface_rel_l2_pct` (batch-limited) and the epoch
3. W&B run ID
4. Compare val/test gap to no-WD baseline (3.700%/4.500%) — does WD close the val-test gap?
[...]

PR #3370, opened 2026-04-25
DrivAerML: metric-aware w=0.05 + T_max=36 + WD=1e-3 (weight decay compound)

tcapelle:
## Hypothesis

Test whether weight decay (WD=1e-3) improves the T_max=36 + metric-aware compound. The DM champion runs without weight decay. At T_max=30, mugen #3328 tested metric-aware w=0.05 + WD=1e-3 — results pending. WD provides L2 regularization on the weights, which could complement the metric-aware loss regularization. With T_max=36's gentler cycling, the interaction may be different from T_max=30.

If WD helps: both forms of regularization (loss-space via rel-L2, weight-space via WD) are complementary.
If WD hurts: the model is already well-regularized by the metric-aware loss and EMA, and WD constrains the capacity.

## Instructions to Student

```bash
  --dataset drivaerml \
  --dm-rel-l2-weight 0.05 \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --weight-decay 1e-3 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
```

**What to report:**
2. The full W&B run ID
4. Any signs of divergence or underfitting from WD


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (jin) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this weight-decay branch was secondary to the kept DrivAerML schedule/LR/test-metric candidates in the final cost-cut pass.

PR #3240, opened 2026-04-23
AirfRANS: joint checkpoint selection + vol-weight 11x/12x/13x fine sweep

tcapelle:
## Hypothesis

PR #3204 revealed a critical insight: the AF best-checkpoint logic selects on \`val_primary/surface_mse\`, but our target is vol_mse < 0.0017. At epoch 763, gilbert's 15x run had vol_mse=0.002378 (only 16.6% above baseline!) but surface_mse=0.000594 — the surface-anchored checkpoint selection HIDES vol improvements behind surface regression.

**Two complementary fixes:**

### Fix 1: Joint checkpoint selection
Change the best-checkpoint metric from pure surface_mse to a weighted combination:
\`\`\`
joint_metric = surface_mse + alpha * vol_mse
\`\`\`
With alpha=1.0, both metrics contribute equally. This ensures the checkpoint that balances surface and volume quality is saved, not just the surface-optimal one.

### Fix 2: Fine vol-weight sweep (11x/12x/13x)
The vol-weight sweep has a gap: 10x=champion, 15x=worse, 20x=crashed, 30x=catastrophic. The 10x→15x transition is too coarse. A fine sweep at 11x, 12x, 13x may find a sweet spot that nudges volume lower without the regression seen at 15x.

Combining both fixes means we: (a) find the right vol-weight in the 10x-15x gap, AND (b) evaluate from the checkpoint that actually optimizes for both metrics.

### Step 1: Implement joint checkpoint selection

In \`target/icml2026/train.py\`, add a \`--checkpoint-metric\` argument:
- \`surface\` (default, current behavior): select best checkpoint on \`val_primary/surface_mse\`
- \`joint\`: select best checkpoint on \`surface_mse + vol_mse\` (equal weight)

### Step 2: Run 3 trials with joint selection + fine vol-weight

**Trial 1 — vol-weight=11x + joint checkpoint:**
\`\`\`
cd target/icml2026 && python train.py --dataset airfrans --airfrans-task full \\
  --optimizer adamw --lr 6e-4 --cosine-t-max 50 \\
  --grad-clip 1.0 --weight-decay 1e-2 \\
  --enable-fourier --model-layers 2 --model-hidden-dim 256 --model-heads 4 \\
  --epochs 999 --ema-decay 0.999 --vol-loss-weight 11.0 \\
  --checkpoint-metric joint --wandb_group af-joint-ckpt-vol
\`\`\`

**Trial 2 — vol-weight=12x + joint checkpoint:**
[...]

PR #3372, opened 2026-04-25
AF: eval-every-3 + AdamW 2L champion + vol-weight=11x

tcapelle:
## Hypothesis

Test eval-every-3 with the **correct AF champion config** (AdamW 2L/256d) at vol-weight=11x. All previous AF eval-every-3 experiments used the wrong config (Lion 4L) and failed. This is the first proper test with moderate vol-weight above baseline (10x).

Gilbert's earlier AdamW 2L experiments (#3240) showed vol-11x produced surface=0.000396 and vol=0.003405 at ~760 epochs. With eval-every-3 giving ~3x throughput, we should get ~2000+ epochs — much deeper convergence.

Tanjiro #3368 is running the same experiment at vol-weight=10x (baseline weight). This experiment tests whether vol-11x can push volume lower without destabilizing.

**IMPORTANT:** Use `--volume-loss-weight` (NOT `--vol-loss-weight`). Confirmed correct flag name.

## Current Baseline (AirfRANS)

- **Surface MSE:** 0.000296 at epoch 704 (nezuko PR #3135)
- **Volume MSE:** 0.002039 at epoch 743 (nezuko PR #3135)
- **Target:** vol_mse < 0.0017, surface_mse < 0.000296

## Instructions to Student

```bash
  --dataset airfrans \
  --airfrans-task full \
  --optimizer adamw \
  --lr 6e-4 \
  --cosine-t-max 50 \
  --grad-clip 1.0 \
  --weight-decay 1e-2 \
  --enable-fourier \
  --model-layers 2 \
  --model-hidden-dim 256 \
  --model-heads 4 \
  --epochs 999 \
  --ema-decay 0.999 \
  --volume-loss-weight 11.0 \
  --eval-interval 3 \
  --save-checkpoint \
  --wandb_group af-eval3-correct-config
```

**What to report:**
1. Best `val_primary/surface_mse` and `full_val/volume_mse` and epochs
2. Total epochs trained
3. W&B run ID
4. Stability (any divergence/NaN?)

**Targets:** surface < 0.000296 AND volume < 0.002039 simultaneously


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (emma) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: the AirfRANS lane already has higher-priority kept coverage for final full-test harvesting, so this overlapping variant was cut to save GPU cost.

PR #3374, opened 2026-04-25
DrivAerML: metric-aware w=0.05 + T_max=36 + lr=5.5e-4 (higher LR probe)

tcapelle:
## Hypothesis

Test whether a slightly higher LR (5.5e-4 vs champion 5e-4) improves the T_max=36 + metric-aware compound. The longer T_max=36 schedule has gentler cosine cycling, which may tolerate a slightly higher learning rate. If the LR optimum shifts upward with T_max=36, this could improve convergence speed and final quality.

Piccolo #3365 tests lr=4.5e-4 (lower). This experiment tests lr=5.5e-4 (higher) to bracket the LR optimum at T_max=36.

- **full-eval TEST:** 4.218% (canute n2t1nzsb)

## Instructions to Student

```bash
  --dataset drivaerml \
  --dm-rel-l2-weight 0.05 \
  --optimizer adamw \
  --lr 5.5e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
```

**What to report:**
1. Best `val/rel_l2_mean` and the epoch
2. W&B run ID
4. Any divergence at cosine restart peaks

**Target to beat:** val < 3.622%


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (thorfinn) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3375, opened 2026-04-25
DrivAerML: T_max=36 MSE-only + lr=4.5e-4 (LR probe at best schedule)

tcapelle:
## Hypothesis

Test whether the LR optimum shifts at T_max=36. The champion lr=5e-4 was tuned at T_max=30. Your previous lr=4e-4 experiment (#3318) at T_max=30 showed slower convergence but was still improving at ep999. With T_max=36's gentler cosine cycling, lr=4.5e-4 might hit a better convergence/quality tradeoff.

This is the MSE-only version (no metric-aware loss). Chrome #3298 achieved val=3.622% with T_max=36 MSE-only at lr=5e-4. If lr=4.5e-4 improves on that, it's a pure schedule-LR interaction finding.

- **val:** 3.622% at epoch 874 (T_max=36 MSE-only, lr=5e-4, chrome PR #3298)
- **SOTA target:** 3.71% AB-UPT (already beaten on val)

## Instructions to Student

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 4.5e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
  --wandb_group dm-tmax36-lr-sweep
```

**What to report:**
1. Best `val/rel_l2_mean` and epoch achieved
2. W&B run ID
4. Compare convergence curve to chrome #3298 (T_max=36 at lr=5e-4)

**Target to beat:** val < 3.622%


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (shouko) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3335, opened 2026-04-24
DrivAerML: metric-aware w=0.05 seed=13 (seed diversity)

tcapelle:
## Hypothesis
Test best metric-aware weight (w=0.05) with seed=13 to assess seed sensitivity. Seeds 0 (#3326 sukuna) and 42 (#3302 baseline) already being tested.

## Instructions
```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset drivaerml --dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.05 \
  --optimizer adamw --lr 5e-4 --cosine-t-max 30 --grad-clip 0.5 \
  --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 \
  --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 \
  --ema-decay 0.9995 --ema-mode fixed --no-compile-model --save-checkpoint \
  --eval-interval 10 --seed 13 --wandb-group dm-metric-aware-sweep \
  --kill-if-best-val-above 150:8.0 --kill-if-no-improvement 200:0.05
```
**Target:** val < 3.700%

PR #3378, opened 2026-04-25
DrivAerML: T_max=36 MSE-only + seed=42 (schedule seed robustness)

tcapelle:
## Hypothesis

Verify that the T_max=36 MSE-only breakthrough (val=3.622%, chrome #3298) is seed-robust. Chrome's run used the default seed. Your previous experiment (#3335) showed moderate seed sensitivity at T_max=30 for metric-aware (seed=13 was 0.063pp worse than seed=42). If T_max=36 MSE-only is similarly seed-sensitive, the 3.622% result may be a lucky seed outcome.

If seed=42 gives a similar result (within ~0.1pp): T_max=36 is robust and the paper-facing claim is solid.
If seed=42 is much worse: the 3.622% was lucky and the true expected performance is higher.

- **val:** 3.622% at epoch 874 (T_max=36 MSE-only, default seed, chrome PR #3298)
- **full-eval TEST:** 4.218% (canute n2t1nzsb)

## Instructions to Student

```bash
  --dataset drivaerml \
  --optimizer adamw \
  --lr 5e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
  --seed 42 \
  --wandb_group dm-tmax36-seed-robustness
```

**What to report:**
1. Best `val/rel_l2_mean` and epoch achieved
2. W&B run ID
4. Compare to chrome #3298 (3.622% at default seed)

**Target:** val within ~0.1pp of 3.622% confirms seed robustness


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (historia) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was a seed-robustness/tail duplicate, and we kept the more central DrivAerML candidates for paper-facing test-metric harvesting.

PR #3380, opened 2026-04-25
DrivAerML: T_max=36 + lr=4.8e-4 MSE-only (LR+schedule compound — HIGHEST PRIORITY)

tcapelle:
## Hypothesis — HIGHEST PRIORITY EXPERIMENT

**Compound the two most important DM findings:** T_max=36 schedule (val=3.622%) and lr=4.8e-4 (full-eval TEST=4.117%, best in programme).

Your previous experiment (#3308) showed lr=4.8e-4 at T_max=30 gave val=3.657% (close to 3.622%) and full-eval TEST=4.117% (best ever, beating 4.218% by -2.4%). The T_max=36 schedule independently improved from 3.833% to 3.622%. These two improvements are likely orthogonal — lr changes the optimization step size, T_max changes the cycling frequency.

If they compound: val could drop below 3.6% AND test below 4.1%. This would be a major breakthrough.

**IMPORTANT:** Do NOT use `--ema-start-step`. The T_max=36 baseline (#3298) didn't use it, and we need a clean comparison.

## Current Baseline (DrivAerML)

- **val:** 3.622% at epoch 874 (T_max=36 MSE-only, lr=5e-4, chrome PR #3298)
- **full-eval TEST:** 4.218% (canute) / 4.117% (your lr=4.8e-4 at T_max=30, PR #3308)
- **SOTA target:** 3.71% AB-UPT (already beaten on val)

## Instructions to Student

```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset drivaerml \
  --optimizer adamw \
  --lr 4.8e-4 \
  --cosine-t-max 36 \
  --grad-clip 0.5 \
  --enable-fourier \
  --model-layers 4 \
  --model-hidden-dim 512 \
  --model-heads 8 \
  --epochs 999 \
  --batch-size 1 \
  --drivaerml-train-surface-points 50000 \
  --drivaerml-eval-surface-points 50000 \
  --max-train-batches 394 \
  --max-eval-batches 200 \
  --ema-decay 0.9995 \
  --ema-mode fixed \
  --no-compile-model \
  --save-checkpoint \
  --wandb_group dm-tmax36-lr48e4
```

After training, run full-eval on the best checkpoint:
```bash
cd target/icml2026 && python train.py \
  --dataset drivaerml \
  --load-checkpoint outputs/icml2026/drivaerml_senpai_transolver_best.pt \
[...]

PR #3383, opened 2026-04-25
DrivAerML: T_max=36 + lr=4.8e-4 MSE-only + seed=42 (LR+schedule seed robustness)

tcapelle:
## Hypothesis
Verify that T_max=36 + lr=4.8e-4 MSE-only is seed-robust. Gojo #3380 runs default seed. If seed=42 converges to a similar result, the compound is confirmed robust for the paper.

## Command
```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 4.8e-4 --cosine-t-max 36 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --ema-mode fixed --no-compile-model --save-checkpoint --seed 42 --wandb_group dm-tmax36-lr48e4
```


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (sukuna) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was a seed-robustness/tail duplicate, and we kept the more central DrivAerML candidates for paper-facing test-metric harvesting.

PR #3385, opened 2026-04-25
DrivAerML: T_max=36 + lr=4.6e-4 MSE-only (deeper LR probe)

tcapelle:
## Hypothesis
Probe lower LR at T_max=36. Gojo showed lr=4.8e-4 > 5e-4 at T_max=30. The LR optimum may be even lower at T_max=36 due to gentler cycling. lr=4.6e-4 tests this. Gojo #3380 tests lr=4.8e-4, shouko #3375 tests lr=4.5e-4. This fills the gap.

## Command
```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 4.6e-4 --cosine-t-max 36 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --ema-mode fixed --no-compile-model --save-checkpoint --wandb_group dm-tmax36-lr-sweep
```


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (mitsuha) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3386, opened 2026-04-25
DrivAerML: T_max=36 + lr=4.8e-4 + metric-aware w=0.06

tcapelle:
## Hypothesis
Compound T_max=36 + lr=4.8e-4 + metric-aware w=0.06. Higher weight at the new LR. Robin #3364 tests w=0.06 at lr=5e-4. This tests whether the weight optimum shifts with the lower LR.

## Command
```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --dm-loss mse_plus_raw_rel_l2 --dm-rel-l2-weight 0.06 --optimizer adamw --lr 4.8e-4 --cosine-t-max 36 --grad-clip 0.5 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --ema-mode fixed --no-compile-model --save-checkpoint --wandb_group dm-tmax36-lr48e4
```


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (shinobu) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this was part of a dense DrivAerML local sweep; we kept only the highest-signal center points for the remaining final-harvest window.

PR #3391, opened 2026-04-25
DrivAerML: T_max=36 + lr=4.8e-4 + WD=1e-3 MSE-only (weight decay)

tcapelle:
## Hypothesis
Test whether weight decay (WD=1e-3) improves the T_max=36 + lr=4.8e-4 MSE-only config. The DM champion runs without WD. At the lower lr=4.8e-4, WD may provide useful L2 regularization that helps generalization (gojo's lr=4.8e-4 already showed better test generalization).

## Command
```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py --dataset drivaerml --optimizer adamw --lr 4.8e-4 --cosine-t-max 36 --grad-clip 0.5 --weight-decay 1e-3 --enable-fourier --model-layers 4 --model-hidden-dim 512 --model-heads 8 --epochs 999 --batch-size 1 --drivaerml-train-surface-points 50000 --drivaerml-eval-surface-points 50000 --max-train-batches 394 --max-eval-batches 200 --ema-decay 0.9995 --ema-mode fixed --no-compile-model --save-checkpoint --wandb_group dm-tmax36-lr48e4
```


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (senku) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: this weight-decay branch was secondary to the kept DrivAerML schedule/LR/test-metric candidates in the final cost-cut pass.

PR #3352, opened 2026-04-24
TFP: LR bracket 6e-5/7.5e-5 around new champion lr=7e-5

tcapelle:
## Hypothesis

Bracket the new TFP champion lr=7e-5 (val=0.002180, test=0.001931, -7%/-13% vs previous baseline) to narrow the optimum. The LR curve showed a non-monotonic dip at 7e-5: 9e-5 (+1.3%), 8e-5 (+3.1%), 7e-5 (-7.0%). Testing 6e-5 (lower) and 7.5e-5 (midpoint) determines if 7e-5 is the exact floor or if the optimum is even lower.

## Instructions

Run two trials sequentially:

**Trial 1: lr=6e-5**
```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset tandemfoil_paper --optimizer lion --lr 6e-5 --cosine-t-max 10 \
  --grad-clip 0.5 --weight-decay 1e-2 --model-layers 3 --model-hidden-dim 192 \
  --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel \
  --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction \
  --enable-pressure-prior-addition --ema-decay 0.999 --epochs 999 \
  --eval-interval 10 --save-checkpoint --seed 0 \
  --wandb-group tfp-lr-bracket
```

**Trial 2: lr=7.5e-5**
```bash
cd target/icml2026 && SENPAI_TIMEOUT_MINUTES=600 python train.py \
  --dataset tandemfoil_paper --optimizer lion --lr 7.5e-5 --cosine-t-max 10 \
  --grad-clip 0.5 --weight-decay 1e-2 --model-layers 3 --model-hidden-dim 192 \
  --model-heads 3 --enable-fourier --enable-te-coord-frame --enable-cp-panel \
  --enable-cp-panel-tandem-only --asinh-pressure --residual-prediction \
  --enable-pressure-prior-addition --ema-decay 0.999 --epochs 999 \
  --eval-interval 10 --save-checkpoint --seed 0 \
  --wandb-group tfp-lr-bracket
```

**Report:** best val/test field_mse for each, compare to lr=7e-5 (0.002180/0.001931).
**Target:** val < 0.002180

PR #3404, opened 2026-04-25
TFP: lr=7.5e-5 + gc=0.3 (softer clipping at optimal LR)

tcapelle:
## Hypothesis

gc=0.3 was the best grad-clip for TandemFoil (21.350 → 21.319 at gc=0.2). The TFP champion uses gc=0.5. At the optimal lr=7.5e-5, softer clipping (gc=0.3) may enable more precise gradient updates, finding a deeper basin.

Lion's sign-based updates can benefit from less aggressive clipping at lower LR — the gradient magnitudes at lr=7.5e-5 are already well-behaved.

## Instructions

```bash
  --dataset tandemfoil_paper \
  --optimizer lion --lr 7.5e-5 \
  --cosine-t-max 10 --grad-clip 0.3 \
  --weight-decay 1e-2 --enable-fourier \
  --epochs 999 --ema-decay 0.999 --seed 0 \
  --wandb-group tfp-lr75e5-schedule \
  --agent nezuko --wandb-name "nezuko/tfp-lr75e5-gc03"
```

Key change: `--grad-clip 0.3` (down from 0.5). All else identical to champion.

## Baseline
- **TFP best val:** 0.001857 (haku lr=7.5e-5 gc=0.5 T_max=10, PR #3352)
- **TFP best test:** 0.001789
- **Target: val < 0.001857**


morganmcg1:
HUMAN: Closing during the final cost-control harvest pass. The associated student pod (nezuko) was intentionally stopped after its .claude logs were archived locally, and this PR was not in the small set kept running for final test-metric harvesting. Reason: the TandemFoil Paper lane already has higher-priority kept coverage, and this variant was a lower-priority schedule/seed/clip probe for the final window.

PR #3467, opened 2026-06-17
[codex] Remove Claude Code for OpenHands agent runtime

morganmcg1:
## Summary

This PR replaces Senpai's Claude Code orchestration path with a durable, infrastructure-agnostic OpenHands Agent SDK runtime.

The final design is a small deterministic Python control plane around OpenHands:

- **OpenHands owns high-entropy work:** research judgment, code changes, evidence interpretation, deciding what to try next, and bounded parallel delegation.
- **Python owns operations that should not depend on an LLM composing fragile tool calls:** GitHub state transitions, assignment publication, training supervision, metric monitoring, conversation selection, retries, deadlines, shutdown, and recovery.

This preserves the advisor/student research workflow while making it more reliable, easier to operate for weeks at a time, and substantially more token-efficient. It is intentionally a replacement, not a compatibility adapter around Claude-specific command semantics.

At head `d9ef9051a4da9ede4cec9f4dea3d367a6d5634fd`, this changes **181 files** with **32,625 additions** and **8,211 deletions**. The structural simplicity pass removed 1,984 runtime lines and split the controller by responsibility; the subsequent soak-driven fixes, runtime-contract polish, and operator documentation remain focused at those existing boundaries.

[`SPEC.md`](https://github.com/wandb/senpai/blob/codex/openhands-conversion/SPEC.md) is the authoritative behavior contract.

## OpenHands fork source

Senpai tracks the `main` branch of [`morganmcg1/software-agent-sdk`](https://github.com/morganmcg1/software-agent-sdk). Fork [PR #1](https://github.com/morganmcg1/software-agent-sdk/pull/1) promoted OpenHands v1.40.0 plus the valid-UTF-8 editor fix to `main`. Current `main` is `808bc9682fed562d6c37f2edba0788e08e0bd1eb`. All 242 non-default fork branches were removed, so `main` is the only branch.

`pyproject.toml` and runtime CI install SDK/tools from `@main`; `uv.lock` records the exact resolved `main` commit used by reproducible image builds.

## Final architecture

```text
entrypoint
  clone/configure
  clear bootstrap credentials
  exec python -m senpai_agent.supervisor advisor|student

Python supervisor
  start one controller worker process group
  restart crashes with bounded exponential backoff
  terminate and restart an overdue phase

Python controller worker
  poll durable GitHub/local state
  partition events by conversation UUID
  reconcile the target checkout
  run one bounded OpenHands turn
  verify and acknowledge durable state
  sleep/backoff/jitter
```

The controller imports no Kubernetes API and requires no Senpai Service, port, DNS entry, ServiceAccount, RBAC rule, cross-node bearer token, Docker network, or tailnet. Kubernetes, Docker on one AWS GPU host, and a local Mac all use the same cross-node protocol: GitHub PR state, assigned-PR comments and reviews, and human-tagged Issues, plus W&B for experiment metrics.

The controller is now split by responsibility rather than keeping six subsystems in one file:

- `controller.py` contains the portable control loop and runtime assembly and is down from about 1,500 lines to 595.
- `github_mailbox.py` derives level-triggered advisor/student GitHub events and owns the active role-aware watcher.
- `mailbox.py` contains the event value, local mailboxes, and per-mailbox fault-isolated composition.
- `state.py` owns assignment conversation identities, event partitioning, and the unified conversation-state ledger.
- `workspace.py` owns non-destructive student checkout reconciliation.
- `monitor.py` owns deterministic training monitoring and direct student wakes.

## Runtime, recovery, and conversations

Each advisor or student instance runs a small Python supervisor. The controller worker publishes an atomic lease containing its PID, current phase, and hard deadline. The supervisor restarts crashes with bounded backoff, terminates the complete worker process group when a lease expires, tracks descendants with PID/create-time protection, and exposes the same lease to Docker health checks and Kubernetes startup/liveness probes.

The default worker shutdown grace is 60 seconds. Kubernetes liveness termination grace is 75 seconds and pod termination grace is 90 seconds, leaving time for OpenHands interruption, training cleanup, event-store closure, and Weave flush before SIGKILL. Process-group signaling is shared in one POSIX helper; it waits the full training cleanup grace where descendants may outlive the leader and reaps the leader in the narrow exit-before-SIGTERM race.

Restarts continue the existing OpenHands conversation:

- The advisor has one durable UUID reused across ideas, reviews, and monitoring threads.
- Each student has one UUID per `(assignment_id, revision_id)`.
- Training-monitor, assigned-PR feedback, and background-child events resume that exact student UUID.
- A killed worker reopens the file-backed OpenHands event log at the last persisted event.
- A conversation is marked started only after its OpenHands turn succeeds, so a crash before delivery retries the complete programme and assignment prompt.

The former `started-conversations.json` and `system-context-revisions.json` ledgers are merged into one atomically replaced `conversation-state.json`. On first upgrade, Senpai migrates both legacy files without modifying them. A conversation caught between the legacy files' two writes resumes without replaying its initial brief and receives the current system context once. Existing new-format state is authoritative, so migration is one-time.

OpenHands history remains the SDK's per-UUID file-backed event log. Senpai does not prune conversation directories or impose a retention count. Student state is ephemeral by default; the branch, PR, structured result, W&B runs, and Weave trace are the durable handoff.

There are three small role-local SQLite stores, none of which is a cross-node queue:

- `advisor-events.sqlite3` stores unacknowledged watcher and background-delegation events on the advisor instance.
- `student-events.sqlite3` stores unacknowledged mid-turn PR feedback and background-delegation events until the exact parent student conversation consumes them.
- `training/monitors.sqlite3` stores monitor policies, samples, and deduplicated actionable signals.

A separate per-student `github-feedback.json` ledger atomically records the assignment revision where immutable PR feedback was first observed and whether a successful turn acknowledged it. It is a small local delivery ledger, not a distributed queue or conversation store.

## GitHub is the cross-node protocol

GitHub state is level-triggered:

- `status:wip` plus exactly one `student:<name>` label is an assignment.
- Trusted human comments and reviews on the student's single open assigned `status:wip` or `status:review` PR wake that exact assignment conversation.
- `status:review` wakes the advisor.
- A live advisor-branch SHA beyond an active assignment's recorded base emits `baseline_advanced` with both SHAs and a compare URL, without cancelling the student.
- Blocked, stale, malformed, duplicate, or rebase-needed assignments become explicit events.
- Open Issues labeled `human` plus the research team/branch/student label are human-to-agent messages.

Assigned-PR feedback is read across all paginated issue comments, submitted reviews, and inline comments belonging to submitted reviews. Senpai accepts GitHub `User` feedback from repository owners, members, and collaborators. Authenticated-actor protocol comments remain hidden except for a schema-validated `senpai-assignment-feedback` marker whose repository, PR, assignment, and encoded revision provide the durable routing identity; this also supports GitHub App/bot comments with `author_association=NONE`. Each item uses its immutable surface, PR number, and GitHub ID as its event key. Its first-seen assignment/revision binding survives restarts, and it is acknowledged only after that exact student conversation completes a successful turn, so delayed revision-one guidance cannot rebind to revision two.

Feedback is delivered oldest-first, limited to eight events and 32,000 rendered prompt bytes per turn. Each message excerpt is bounded to 4,000 bytes and links to the complete GitHub text. Immediate post-turn polling drains later batches without dropping them.

While an advisor or student turn is active, `ActiveGitHubWatcher` continues polling. Advisor events and only PR feedback bound to the currently running student UUID are queued locally and injected through OpenHands' concurrent-message boundary after the next safe tool/model boundary, without cancelling unrelated work. Injected feedback reaches `github-feedback.json` only after the enclosing turn succeeds; a failed turn leaves it pending. Human Issue messages use the exact latest human-authored body/comment ID as their dedupe key; agent replies do not wake the agent again.

A malformed assignment marker now emits a bounded `malformed_assignment` event without suppressing human Issues or every later GitHub event. Duplicate student assignments similarly preserve human messages.

### One shared bounded read path

`GitHubReader` is the common authenticated read boundary for the GitHub mailbox and PR-bundle retrieval. It owns same-origin cycle-safe pagination, typed secret handling, actor caching, compact errors that cannot expose the token, and the shared Link-header parser. `github.py` no longer shells out to `gh api`, and the mailbox no longer carries another urllib client. Launch-time GitHub access, branch, and label operations now use one helper within `launch_helpers.py`. The mutation workflow intentionally retains its separate transport because it must model ambiguous writes and expected statuses.

### Complete PR retrieval without context pollution

`get_prs` is one interface for explicit PR numbers, inclusive creation-date ranges, and GitHub search expressions. Every selected PR contains its full body, all issue comments, all submitted reviews, and all inline review comments across pagination.

At or below `max_inline_prs` (default five), Markdown is returned directly. Above the threshold, the same Markdown is written to one mode-0600 artifact outside the target checkout and the agent receives a compact path/manifest. Raising the inline threshold above five explicitly warns about context-pollution risk. There is no duplicate JSON artifact and no mandatory summarizing child.

### Convergent typed transitions

One discriminated `github_transition` tool owns assignment creation, branch publication, label reconciliation, non-revision assignment feedback, revision requests, human-Issue responses, result submission, experiment close, and merge.

Transitions validate repository, role, student, assignment, revision, branch, local commit, remote lease, and trusted marker identity as applicable. Active-assignment uniqueness is global per student rather than per base branch and includes both WIP and review-ready PRs. `submit_result` validates the exact local head before any push. Publication uses the captured commit SHA with force-with-lease; the pre-push hook accepts that raw SHA only when it equals the current role-owned branch tip.

The `submit_result` schema explicitly distinguishes the current remote lease SHA from the local result commit and requires its three local-result SHA fields to agree. A stale assignment revision ends only the obsolete turn so the controller can resume the current revision. After a successful branch push, bounded retries handle GitHub's temporary PR-head propagation lag only when the observed head has not yet reached the pushed commit; every other precondition failure still fails immediately.

The workflow uses strict Pydantic response contracts rather than a parallel hand-written JSON-validation layer. Repeated pull/head/assignment guards funnel through shared helpers, revision/result marker comments use one upsert implementation, and callers no longer refetch a comment that the upsert already verified. Safety reads used for ambiguous-write convergence and post-transition race verification remain explicit. Transition dispatch is exhaustive and raises on an unhandled future operation instead of falling through to merge.

Revision requests inspect and persist the trusted revision comment before changing the assignment body. `send_assignment_feedback` instead writes one immutable typed guidance comment with exact assignment, revision, head, actor, and routing preconditions; it changes no revision, draft state, or labels. Exact replay is a no-op, while changed guidance requires a new feedback ID so GitHub produces a new durable event.

Immediately before a first merge, `merge_experiment` reads the live Git ref for the assignment's base branch. Drift blocks the merge unless the advisor explicitly supplies `accepted_base_sha` equal to that exact live SHA after deciding the evidence remains scientifically valid. A wrong or obsolete acceptance fails, while already-merged replay returns before the ref check. Pagination, exact-column marker matching, spoof resistance, duplicate trusted markers, replay, and ambiguous-write behavior are exercised through public transitions rather than a dead generic comment API.

## Unified subagent delegation

Advisor, student, and delegated agents use one model-facing function:

```text
delegate_agent(
  task: str,
  agent: general-purpose | explore | search | bash-runner = general-purpose,
  model: smart | fast = smart,
  background: bool = false,
  include_context: bool = false,
  search_mode: general-web | research-publications | null = null,
) -> {task_id, status, result?}
```

Every call launches the selected Markdown-defined OpenHands agent in a fresh conversation and separate process group. Foreground mode returns the report inline; background mode lets the parent continue and delivers a durable result/error event later. OpenHands' parallel tool executor and Senpai's semaphore both cap active children at eight.

`model=fast` is for mechanical `rg`/grep, command execution, bounded extraction, and straightforward inspection. `model=smart` is for code review, literature research, ambiguous synthesis or failure diagnosis, and decisions where missing a subtlety is costly. File-agent reasoning effort is operational on the OpenHands fork's `main` branch.

File-defined children receive only the OpenHands tools declared by their definition. General-purpose and Explore receive terminal, file editor, and bounded nested foreground delegation. Search receives terminal and file editor but cannot delegate. Bash Runner receives only the terminal. No child receives a GitHub credential, `get_prs`, `github_transition`, or training tools. The parent prepares large PR artifacts and owns every typed GitHub operation.

The complete parent event log remains searchable as JSON through ordinary `rg` and bounded reads. A context-free Explore child automatically receives `$SENPAI_PARENT_CONVERSATION_HISTORY_DIR` and can return compact findings with paths and line numbers instead of copying history into the main context.

## Native Markdown agents and Exa search

Senpai defines `general-purpose`, `explore`, `search`, and `bash-runner` as native `.agents/agents/*.md` files. Both images install all four definitions. At runtime Senpai registers OpenHands' default tool registry before discovering and registering the files, so every frontmatter tool name resolves through the same native loader used in production.

Bash Runner adapts OpenHands' useful terminal-specialist preset to Senpai's delegation and credential boundaries. It is intended for tests, builds, linters, formatters, dependency commands, Git inspection, and bounded system checks whose output would otherwise pollute the parent context. It normally uses the fast model, inherits the caller-selected reasoning effort, runs in its own process/conversation like every other child, and returns only status counts and actionable failures rather than raw logs. It cannot edit files, launch nested agents, use the browser, call Senpai tools, or mutate GitHub.

Search has one progressive `exa-search` skill with two explicit modes:

- `general-web` defaults to Exa `type="auto"`, ten results, no category, and query-relevant highlights.
- `research-publications` defaults to `category="publication"`, `type="deep"`, thirty results, and bounded highlights.

The script emits compact Markdown rather than raw JSON, percent-encodes unsafe URL characters, treats retrieved content as untrusted evidence, validates incompatible content flags, loads a local `.env` without overriding exported credentials, and never sends `includeDomains` with Exa's publication category.

Launch preflight verifies the Exa credential with the cheapest bounded request requested for this rewrite: `type="instant"`, `category="publication"`, and `numResults=1`.

## Training and deterministic monitoring

Students receive `run_training`, `get_training_status`, `cancel_training`, and `monitor_training`. `TrainingSupervisor` owns the process group and deadline, persists restart identity and the full log, retains a bounded error tail, and discovers W&B run IDs while training is still active.

`run_training` refuses a dirty assignment worktree before spending an allocation and immediately registers a terminal-state monitor bound to the current conversation. `monitor_training` is only a metric/gate/staleness upgrade: it cannot remove finished, failed, timed-out, or cancelled wakes. Persisted pre-upgrade actions containing the old caller-controlled status filter still deserialize, but the obsolete filter is discarded so restart recovery remains compatible and terminal safety cannot be downgraded.

The declared training timeout is a strict end-to-end wall-clock ceiling. Senpai begins termination early enough that graceful shutdown and kill escalation finish by the deadline, so a 30-minute policy cannot become 30 minutes plus process-cleanup grace.

Running training observations keep `exit_code` nullable, and inconclusive experiment results may omit a primary metric or unavailable baseline/delta values. Terminal monitor wakes report the terminal state and current detail without attaching the last cached metric sample, which may predate the terminal state.

The controller polls only due monitors, fetches one latest selected W&B metric, and evaluates deterministic threshold, direction, change, staleness, and terminal-state rules. Ordinary polls use no LLM tokens.

Every persisted signal directly resumes the exact student conversation that registered the policy. The context-free monitor-triage child, decision JSON, triage lease, retry parsing, and associated operator settings are removed: registering a monitor is already the student's explicit request to wake when that condition occurs.

Metric models reject NaN and infinities. A status/W&B failure or malformed sample advances only that monitor's schedule, emits one bounded `monitor_error`, repairs invalid sample state, and cannot block other monitors, GitHub events, child results, or pending wakes. Upgrading or changing metric gates replaces the optional metric policy and resets derived samples/signals while retaining mandatory terminal wakes.

`improved_by` and `regressed_by` now compare against the first observed sample for the monitoring policy rather than the preceding 60-second sample. The baseline is durable. Schema migration copies a legacy previous sample into the new baseline column in the same SQLite transaction, including an intermediate schema that already created the column with a NULL value, so the first post-upgrade delta is not lost.

Controller events are partitioned by exact conversation UUID. Each batch is acknowledged only after its own successful turn, preventing a child result for conversation A from consuming or permanently blocking a training event for conversation B.

## Prompt stack, skills, and provider context

The model receives OpenHands' native base prompt and tools, one stable suffix merged from `SENPAI-HARNESS.md` and the rendered role charter, applicable target `AGENTS.md`/compatible `CLAUDE.md`, a progressively disclosed skill catalog, and live user-turn context containing `program.md`, role instructions, current state, and current UTC time.

Skill bodies are not concatenated into child system prompts. The complete role remains in the system message rather than being periodically duplicated. File-defined children are explicitly told that their definition and delegated task define their scope. A persisted merged-context hash injects an updated deployed harness/role once without rotating the conversation UUID.

The project tracks the `main` branch of the [`morganmcg1/software-agent-sdk`](https://github.com/morganmcg1/software-agent-sdk) OpenHands fork. Its README links at the top to [`FORK_MODS.md`](https://github.com/morganmcg1/software-agent-sdk/blob/main/FORK_MODS.md), whose newest-first entries identify each divergence by change name, timestamp, commits, files, and functions.

Direct `openai/*` models use the Responses API with stored response continuation, `previous_response_id`, all-turn reasoning context, detailed reasoning summaries, explicit prompt-cache boundaries for GPT-5.6, and provider compaction at 200,000 rendered tokens. Direct `anthropic/*` models use native Anthropic compaction at 200,000 input tokens and persist/replay the provider's typed compaction block. OpenHands' local condenser is disabled only for those provider-owned chains; the complete local event log remains intact and searchable.

## Hooks and credential containment

The native plugin declares `PreToolUse`, `Stop`, and `SessionEnd` hooks. Its pre-tool policy covers both the main `senpai_terminal` and file-defined subagents' raw `terminal`, so delegation cannot bypass GitHub or supervised-training boundaries; the main terminal additionally enforces the same pure policy in-process. It recognizes attached `gh api` mutation flags, uppercase GitHub API hosts, nested shell/env wrappers, `timeout`, `setsid`, foreground loops, sleeps, log streams, and `gh pr checks --watch`. The target pre-push hook enforces exact advisor/student branch ownership and rejects deletion/default/foreign refs.

The shell policy permits ordinary finite `for` loops, help-only inspection of training entrypoints including redirected output, and quoted literal heredocs written directly by `cat` to a literal file. Unquoted or dynamic heredocs, pipelines, executable consumers, process substitutions, function overrides, and non-file sinks remain subject to the full policy rather than having their bodies hidden.

The GitHub write token is used for bootstrap, written once to a mode-0600 file under pod-local `/tmp`, removed from the environment, and consumed/unlinked by PID 1. The long-lived supervisor holds it only in typed process memory. Before each controller restart, PID 1 creates a one-shot inherited pipe; the worker reads and closes that pipe before OpenHands starts. The token never appears in PID 1's environment, the agent terminal, model-facing schemas, delegated children, or shared conversation/dataset storage.

All GitHub credential names and scrubbing live in one shared module and every child/process boundary uses it. Weave dynamically registers the GitHub token with its content redactor after the one-shot handoff.

Advisor and student images run as UID/GID 10001. Kubernetes uses `runAsNonRoot`, runtime-default seccomp, no privilege escalation, and drops all Linux capabilities.

The pinned [`morganmcg1/weave-openhands`](https://github.com/morganmcg1/weave-openhands) integration initializes before OpenHands imports and traces advisor, student, and child conversations. It records agent, model, and tool spans under the durable conversation ID and flushes on runner/controller shutdown. Every `OPENHANDS_RUN` record includes a direct `weave_url`; these traces live in Weave Agent Observability and can be queried with `get_agent_spans()`, rather than through legacy Weave Calls.

## Images and launch behavior

The image workflow builds three images from the exact same source commit:

- **advisor:** Python/OpenHands, GitHub CLI, Chromium, and `tmux` for OpenHands' persistent terminal implementation; no CUDA, PyTorch, or Kubernetes tooling.
- **student:** CUDA/PyTorch plus the same OpenHands and Chromium runtime.
- **cutoff:** a minimal Python/shell image with one checksum-verified pinned `kubectl`.

Advisor and student install Chromium in the runtime user's canonical Playwright cache and run a real OpenHands browser smoke test as that non-root user during image construction. Student additionally validates CUDA architecture support. The cutoff arming script accepts only a full source-SHA image tag or immutable digest and defaults to the cutoff image for the exact checked-out commit.

The optional start gate lives in the Python controller, which refreshes a `start-gate` lease while waiting. Startup probes therefore become healthy before the cutoff job opens the gate. Both launch and cutoff CLIs reject relative, pod-local, or non-normalized gate paths; a gate must be an absolute file beneath the shared PVC mount.

The cutoff no longer waits forever for exact fleet readiness. It arms when every expected pod/deployment is ready or when a configurable readiness deadline expires, opens the optional start gate in either case, persists why it armed, and starts the hard fleet budget. One missing or crash-looping pod can no longer prevent the very cutoff intended to contain it.

The launcher accepts an explicit `--kube_context` and `--namespace` and applies that scope to every resource apply, existing-student discovery, and emitted monitor/log/stop command. Launching a fleet can no longer silently target whichever cluster happened to be the operator's current kubectl context.

The operator README now starts with install, credential, target-package, local-config, preflight, and launch steps; it documents the real multi-node PVC, Kubernetes permission, CUDA-driver, and immutable-image prerequisites, and clearly labels non-Kubernetes launch bootstrap as manual.

## Deletions and simplification

This PR intentionally removes rather than carries parallel implementations:

- Claude Code runtime and image installation.
- `.claude/` runtime resources; the compact root compatibility document remains.
- Claude/OpenHands shell watchdog loops replaced by one Python supervisor.
- Exa MCP configuration replaced by a progressive skill/script and bounded preflight.
- HTTP advisor service, bearer token, port, Service, probes, and RBAC.
- Shell GitHub polling and pod-process inspection.
- Cluster-cutoff conversation harvesting.
- Full skill-body inlining.
- The mandatory Laminar telemetry chain; Weave is the configured integration.
- The context-free monitor-triage model call and its decision store.
- The production `create_senpai_tools` factory that was only called by tests and leaked a temporary monitor directory.
- Large Protocol mirrors of concrete GitHub/training classes.
- Dead standalone `conv_state` fallbacks and the unused `--continue` marker path.
- Windows-only branches from the explicitly POSIX container supervisor.
- The unused public marker-comment upsert.
- `poll-for-work` and `survey-prs`, whose polling behavior now belongs to the deterministic controller.
- Unreachable outer poll backoff duplicated by per-mailbox fault isolation.

The current paper's live skill catalog is updated; historical Claude-era appendix sources remain historical.

Hivemind remains commented out pending its separate rewrite.

## Deliberate non-goals

- No separate advisor PVC is introduced; advisor storage topology remains an operator/infrastructure decision.
- No hard Senpai token/cost budget is added here.
- Senpai does not own conversation-retention policy.
- GitHub and W&B remain the shared durable research records; local SQLite is not promoted into a distributed database.
- Training command executables remain intentionally flexible.
- The unrelated operational-skill integrity cleanup is deferred.
- Skill-declared model/reasoning behavior outside the implemented Markdown-agent path remains a future OpenHands capability.

## Validation

Code validation for exact head `d9ef9051a4da9ede4cec9f4dea3d367a6d5634fd`:

- Complete local Python 3.13 suite: **536 passed, 6 skipped in 53.75 seconds**. The six skips require an explicit immutable student image and source revision for cluster/GPU E2E.
- The exact-head [OpenHands runtime workflow](https://github.com/wandb/senpai/actions/runs/30841980274) passed; the [three-image workflow](https://github.com/wandb/senpai/actions/runs/30841981740) is running, including Chromium and CUDA/Torch smoke validation.
- Exact-head regressions cover mid-turn student feedback injection, exact-UUID routing, failed-turn redelivery, no duplicate prompt/store delivery, typed bot feedback, delayed prior-revision binding, immutable non-revision feedback, live baseline events, exact-SHA merge acceptance and replay, WIP/review active-assignment uniqueness, clean-worktree launch rejection, automatic and non-downgradable terminal monitoring, legacy ActionEvent recovery, oldest-first bounded feedback draining, nullable observations/results, stale terminal metrics, safe literal-heredoc inspection, and delayed GitHub PR-head propagation.
- The delayed-head regression exercises the integrated typed `submit_result` path: one branch push, stale PR reads, bounded retry delays, convergence on the pushed head, one authenticated result marker, draft removal, and final `status:review` labels.
- Native Markdown agents, Bash Runner isolation, model/effort selection, provider continuation and compaction, credential containment, supervisor recovery, process-group deadlines, monitor routing, typed GitHub transitions, Kubernetes scoping, image splitting, and cutoff behavior remain covered by the complete suite.
- Python compilation, all shipped shell scripts under `bash -n`, `git diff --check`, and `uv lock --check` passed.
- Live launch preflight passed GitHub `Contents: write` capability and target branch validation, Anthropic authentication, a one-result Exa `instant` publications search, and a minimal W&B viewer query.
- Multiple independent subsystem reviews, adversarial verification, and the live cluster soak drove the fixes included at this head.

### Six-hour live r2 soak

The immutable r2 fleet ran from `2026-07-31T08:43:27Z` until its exact six-hour cutoff at `2026-07-31T14:43:27Z`, using one advisor and two eight-GPU students across three physical nodes. Every workload pod retained its original UID and reported zero restarts for the complete session. Supervisor leases stayed healthy, training remained supervised, and actionable terminal events resumed the same student conversation UUIDs that registered their monitors.

The cutoff fired at exactly `14:43:27Z`. It deleted all three Deployments, their workload pods and ReplicaSets, the launch ConfigMaps, and the launch Secret. The cutoff Job exited zero with zero restarts and completed at `14:43:29Z`; its completed Job and pod remain intentionally as audit artifacts. Post-cutoff verification found no workload processes, resources, or late work. The advisor received SIGTERM, paused its OpenHands conversation, and persisted resumable state.

The soak exercised repeated assignment, revision, training, monitor, review, close, and merge cycles against `morganmcg1/TandemFoilSet-Balanced`:

- PR #4611 revision 2 launched eight separate typed training processes and eight monitors. Every run reached a terminal state in under 22.3 minutes, the terminal wake resumed the same student UUID, the repeated experiment confirmed a null result, and the PR closed without changing the baseline.
- PR #4613 exercised one supervised wrapper process group containing exactly eight child runs. This was a topology deviation from per-arm supervision, but the hard safety boundary remained intact: all eight children exited successfully, the wrapper completed in 1,378.9 seconds, no child exceeded 22.75 minutes, and the terminal wake resumed the same student UUID. The advisor merged the measured winner at `730a5f9`, making L1 loss, slice size 64, `bf16_compile`, and batch size 2 the remote baseline.

The advisor's final SIGTERM arrived while it was preparing a local `RESEARCH_LOG` wrap-up, before any wrap-up commit was created. No partial research-log state reached GitHub: the remote target branch remains the clean `730a5f9` merge baseline, the advisor conversation is persisted, and the narrative wrap-up is intentionally left for a future resumed turn.

Two teardown observations did not invalidate the run. A student emitted a W&B atexit `ConnectionResetError` only after all training runs and GitHub submissions had completed, with no lost run or result. A NaN encountered in target-repository scoring was diagnosed as target-code/data behavior rather than a Senpai controller, monitor, process, or transport failure.

The live soak used the immutable image built from `dae581b8b5cc013d684f4314e4c3f8bcc6d98b3a`. It was deliberately not restarted as soak-driven source fixes landed, because preserving six-hour conversation and process continuity was part of the test. The later fixes through current head are validated by the exact-head local suite and the current-head workflow runs linked above rather than being misrepresented as deployed inside the already-running fleet. The soak exposed and drove fixes for nullable running observations, shell-policy false positives, stale assignment submissions, PR-head propagation lag, advisor terminal support, stale terminal metrics, and durable student PR feedback. It finished with no unresolved harness error.

`CLAAssistant` is a separate repository policy check and is not a runtime or test failure.

## Suggested review order

1. `SPEC.md` for the target architecture and invariants.
2. `senpai_agent/controller.py`, `mailbox.py`, `github_mailbox.py`, `state.py`, and `workspace.py` for the small control plane and conversation routing.
3. `senpai_agent/monitor.py`, `training.py`, `supervisor.py`, and `processes.py` for long-running lifecycle and training wakes.
4. `senpai_agent/github_http.py`, `github.py`, `github_workflow.py`, `git_workflow.py`, `tools.py`, and `hooks.py` for bounded reads and guarded remote state transitions.
5. `senpai_agent/openhands_runner.py`, `delegation.py`, `secrets.py`, and `weave_monitoring.py` for the SDK boundary, children, credentials, provider context, and tracing.
6. `system_instructions/`, `.agents/agents/`, and `plugins/senpai/` for prompt, file-agent, skill, and hook contracts.
7. `Dockerfile.advisor`, `Dockerfile.student`, `Dockerfile.cutoff`, and `k8s/` for packaging, probes, security contexts, and launch behavior.
8. `tests/` for the executable acceptance contract and review regressions.


morganmcg1:
## AWS acceptance follow-up

The stacked Docker/AWS acceptance run exposed infrastructure-independent agent
mechanics, so those fixes landed here rather than in #3472.

Current head: `253fa2b46394a7a5b94695081b5fff803b11dbdd`

- added typed, conversation-owned `cancel_training`, including monitor
  retirement and idempotent supervisor cancellation
- prevented cross-conversation monitor takeover
- made `push_branch.expected_head_sha` a required local-head precondition
- accepted and validated common optional `repo` scope for all GitHub transitions
- added exact merge/push/submit/close examples and a close-experiment skill
- allowed only local `git config user.name` / `user.email` reads and writes
- pinned an OpenHands file-editor fix that validates UTF-8 before statistical
  legacy-charset detection ([software-agent-sdk#1](https://github.com/morganmcg1/software-agent-sdk/pull/1))

The live evidence was unusually strong:

- the old deployed transition schema rejected the model's natural top-level
  `repo` argument once in every student result cycle (**11/11**), then recovered;
  the shared optional repo field removes that recurring retry
- one divergent training needed raw SIGTERM because no typed cancel tool existed;
  the new cancellation path preserves ownership and terminal state
- one normal file-editor insert misclassified a valid UTF-8 research log as
  `ptcp154`, turning a 24-line addition into 184 additions / 160 deletions; the
  advisor caught it before push, and the pinned fix preserves untouched bytes

Validation at this head:

- complete Senpai suite: **525 passed, 6 skipped**
- OpenHands file-editor suite: **159 passed**
- required OpenHands Ruff, pycodestyle, pyright, import, and registration checks passed
- lockfile, Python compilation, all shell scripts under `bash -n`, and
  `git diff --check` passed
- the final stacked student image passed the UTF-8 insert regression and a real
  CUDA tensor operation on the acceptance H100

The four-hour fleet itself was not restarted as these fixes landed; preserving
continuous advisor/student state was part of the acceptance contract. Final-head
unit/integration suites, image builds, and the post-gate immutable-image H100
smoke validate the resulting source without claiming it ran inside the older
already-live containers.

morganmcg1:
Final validation update (supersedes the source-head references in my earlier soak comment):

- OpenHands/core head: `6cf5b968e15ef34079bc3d6cf17b2ad2b2154900`
- stacked Docker/AWS head: `2aafa0e2adaa6e362188ec0b72cbc2d650c7b816`, whose exact merge base is the core head
- pinned UTF-8 fix fork: `morganmcg1/software-agent-sdk@13a3e946b80c92d0ee3b6a65492bd39e1e4198cb` ([draft PR](https://github.com/morganmcg1/software-agent-sdk/pull/1)); the final follow-up only made its regression-test assertions Windows-neutral, with production behavior unchanged
- complete core suite: **525 passed, 6 skipped**
- complete stacked suite: **623 passed, 6 skipped, 11 subtests passed**
- final exact-stack images: [workflow 30728880589](https://github.com/wandb/senpai/actions/runs/30728880589), all three jobs green and immutable tags independently resolved in GHCR

The four-hour H100/TandemFoil acceptance findings and mechanics conclusions in the earlier comment are unchanged.

morganmcg1:
Follow-up pushed in `758a56d`: configurable advisor, student, smart, fast, and frontier model profiles.

- `senpai.yaml` now exposes each model and reasoning effort.
- Frontier defaults to `openai/gpt-5.6-sol` at `max`; OpenAI documents `max` for GPT-5.6, not `ultra`.
- Provider credentials are inferred from the model prefix (`anthropic/...` → `ANTHROPIC_API_KEY`, `openai/...` → `OPENAI_API_KEY`); there is no frontier key-env setting in the user config.
- `delegate_agent(model="frontier")` defaults to the general-purpose agent with terminal, file editing, task tracking, tests, and nested foreground delegation.
- Reasoning effort is passed through unchanged. Unsupported values or model/effort combinations fail explicitly; there is no clamping.
- Post-rebase validation: **564 passed, 6 skipped**.

morganmcg1:
OpenHands fork follow-up complete.

- Fork PR [#1](https://github.com/morganmcg1/software-agent-sdk/pull/1) merged OpenHands v1.40.0 plus the valid-UTF-8 editor fix into `main`.
- Fork `main` is currently `808bc9682fed562d6c37f2edba0788e08e0bd1eb`; the valid-UTF-8 commits and all fork-specific SDK extensions are reachable from it.
- All 242 non-default fork branches were deleted; `main` is the default and only branch.
- Senpai SDK/tools declarations and runtime CI use the fork's `@main`; `uv.lock` now resolves both packages to current `main` at `808bc968`.
- Validation at Senpai head `9ef895b9`: `uv lock --check`, 23/23 focused fork-contract tests, OpenHands runtime CI, and advisor, student, and cutoff image builds all passed.

This supersedes earlier comments referring to `13a3e946`, `72cf0ce0`, `8336a23d`, the draft fork PR, or a feature-branch pin.

The remaining red checks are unrelated to the branch selection: CLA Assistant policy and Socket's existing `cryptography==49.0.0` advisory (`GHSA-g6cj-pr64-35w5`).

PR #3532, opened 2026-08-22
Document advisor delegation diagnostics

morganmcg1:
## Summary

- broaden DIAGNOSTICS.md into a Senpai advisor diagnostics guide
- add a reusable workflow for comparing delegation and model use across matched rollout windows
- document source reconciliation across the delegation registry, OpenHands actions, and W&B Weave agent spans
- include sanitized Weave and SQLite examples plus the current policy-rollout case study
- distinguish operational availability, advisor behavior, workload mix, and requested tier from observed provider model
- recommend durable task and tree identifiers on child root spans for exact future joins

## Finding

The observed increase is mixed, with direct evidence of behavior change. Restored uptime inflated the raw fivefold increase, but the rate remained about threefold higher after adjusting for the inactive interval. Spawn decisions and recursive requests increased while observed launch reliability and median duration remained broadly stable. Fable share moved only from 20% to 22%, so the increase was broad delegation rather than a measurable Frontier-specific share shift. A broader image rollout and missing pre-rollout registry state prevent a causal percentage split.

## Validation

- `pytest -q tests/test_agent_markdown.py` (7 passed)
- both new Python examples compile
- Weave query validates against the installed Agents schema
- `git diff --check`


morganmcg1:
I have read the CLA Document and I hereby sign the CLA

PR #3544, opened 2026-09-24
Fix Senpai environment isolation and multi-node startup diagnostics

morganmcg1:
Target-repository `uv run` and `uv sync` commands inherited the runner's environment overrides and could remove OpenHands dependencies or PyTorch from a live agent. Clear those overrides before starting either controller, and keep the Kubernetes proxy pinned to the runner interpreter.

Supervised multi-node runs also hid worker startup failures before W&B initialized. Return bounded live pod placement, container states, and init/training logs through `get_training_status`, including launcher pods only when their owner UID chain reaches the reserved MPIJob. The executor retains Kubernetes credentials and owns polling and cleanup.

Validation: 109 focused training/executor/tool tests and 8 entrypoint/environment tests passed; `git diff --check` passed. Deployment and a four-node online W&B recovery smoke will provide integration evidence.

This PR targets the deployed multi-node runtime branch (`codex/amf1-main-sync-20260823`, source `0f5d19a30a24530883c2c059325f6bc221ab8473`). The current main branch no longer contains this runtime.


morganmcg1:
Deployed source `4f833ca29acfa03f5285be64913ecdd39935ca78` to the existing `jc_vandam` advisor and four students on `navier-stokes`. [Image build](https://github.com/wandb/senpai/actions/runs/35973456461) passed for all roles; registry revision labels match, and the deployment uses immutable image digests.

All five replacement controllers reached Ready with zero restarts after bootstrap repair. Existing conversations, event databases, and local research commits were preserved, including an unpublished student commit. The restored student has W&B 0.28.1, PyTorch 2.13.0+cu130, and a working litellm import. Its actual supervisor environment contains none of the three cleared UV/virtualenv overrides; its Kubernetes proxy uses the pinned runner interpreter; the typed result/tool expose the new diagnostics.

The canonical `WandbMetricSource.latest` returned `1.0` from the online [infrastructure probe](https://wandb.ai/milieu/nn_cfd_senpai/runs/oii2miz8). The [four-node recovery smoke handoff](https://github.com/wandb/cfd_senpai/issues/148) is acknowledged by the advisor; GPU integration evidence is still pending. The separate UID 10001 dataset audit remains blocked on a shared-directory traversal permission decision; the production training image runs as UID 0.

